PR #40, opened 2023-01-24
Fix table column names

Arnei:
includes #39, should only be merged after it.

Partially adresses #57. Pagination should not fail anymore, and previously unsortable columns should now be sortable again.

Fixes various table column names for events and series that have changed over the years. This should make sorting for those columns work again.

### DIFF
--- a/modules/admin-ui-picard/app/src/components/shared/Table.js
+++ b/modules/admin-ui-picard/app/src/components/shared/Table.js
@@ -138,13 +138,11 @@ const Table = ({
 	};
 
 	const sortByColumn = (colName) => {
-		requestSort(colName);
+		const direction = requestSort(colName);
 		setSortBy(colName);
-		if (sortConfig) {
-			reverseTable(sortConfig.direction);
-		} else {
-			reverseTable("ASC");
-		}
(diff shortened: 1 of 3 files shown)


naweidner:
The sorting behavior still seems odd/strange.

Example:
- I have 15 videos/events total and the view is limited to 10 entries per page.
- Go to page 2 and sort descending by title (2 clicks)
- Go back to page 1
- Sort ascending by title (1 click)
- Wait a few seconds and see how the sorting changes 

I would expect "I love Gänse 01" to be the first entry (as it 
is for the first few seconds).
But after a few seconds the sorting changes and "I love Gänse 06" is the first entry.

[Video of sorting behavior.webm](https://user-images.githubusercontent.com/93541078/217508908-ad7fd67a-8b69-41b3-91a7-3fe638bda5ce.webm)

Arnei:
Thanks again for the report, the latest commit should fix this behaviour.

You may still experience wrong sorting for a few seconds after initiating a new sorting, which then corrects itself. I agree that that is not optimal, but would suggest making a separate issue/PR for that. It is somewhat out-of-scope for this PR, at least in matters of urgency I think,

naweidner (approved):
Sorting looks good now.

PR #134, opened 2023-02-14
Extract admin ui picard from Opencast #2

Arnei:
*Some the original PR #98 broke, probably during the numerous merge conflict resolves. Since this PR removes and moves a lot of files, this may happen again in the future. Definitely check if this still builds before merging. I redid this PR to start over clean.*

Deletes all Opencast files, leaving only the Create React App Project with the new Admin UI.

Goes together with https://github.com/opencast/opencast-admin-interface/pull/97.


naweidner:
Have built it locally and tested the mock view a bit.
So far everything seems to be working as expected.

naweidner (approved):

Arnei:
> Have built it locally and tested the mock view a bit. So far everything seems to be working as expected.

Could you also test the normal view with local and external Opencast (e.g. develop.opencast.org) while you're at it?

PR #188, opened 2023-05-08
Change anchors to buttons

### LINKED ISSUE #100: Replace misused anchor tags
The new Admin UI uses `<a>` tags in many places where the arguably do not belong, e.g. the language select dropdown or the actions in the events details view. This is problematic from a usability and accessibility point-of-view, as they are not appropriately styled to accommodate for either of the two. For example, they are hardly keyboard navigable and will trip up screen reader users quite a lot. Thus, the `<a>` tags  should be replaced with more suitable tags.

I think using `<button>` tags makes sense in most places where this in an issue, although they need to be styled to maintain the current "look" of the Admin UI. Alternatively, we could build our own buttons from `<div>` tags, but that is probably more trouble than it is worth.

Arnei:
This switches a ton of anchor tags for button tags, thereby removing a ton of ESLint warnings and
hopefully preparing the Admin UI for future
accessibility updates.

Since this is supposed to be just a semantic update, everything should look the same as before, particularly every clickable thing.

Does not fix the warnings in TimeSeriesStatistics.js because I don't currently know how to display the associated UI and thus can't check any changes.

Resolves #100

### DIFF
--- a/app/src/components/Header.js
+++ b/app/src/components/Header.js
@@ -294,14 +294,14 @@ const MenuLang = () => {
 			{/* one list item for each available language */}
 			{languages.map((language, key) => (
 				<li key={key}>
-					<a onClick={() => changeLanguage(language.code)}>
+					<button className="button-like-anchor" onClick={() => changeLanguage(language.code)}>
 						<img
 							className="lang-flag"
 							src={language.flag}
 							alt={language.code}
 						/>
 						{language.long}
-					</a>
+					</button>
 				</li>
 			))}
 		</ul>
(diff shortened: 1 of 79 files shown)


naweidner:
Great. This significantly reduces the "jsx-a11y/anchor-has-content" warnings.

There are still 4 of these warnings left (besides the warnings in TimeSeriesStatistics.js - and I also have no idea how to solve and check this).

The left anchors are the external-link icons in the details of the attachments tab in the event modal and the play icon in the publication tab.   I tested a bit but couldn't get them to work as buttons (they were not clickable anymore)

These are the external link icons, that I mentioned:
grafik
(Go to _event-details > Assets > any of the assets types details > Details again > icon in the last row_)

naweidner (approved):
I think we can merge this and then see if eslint still complains.

Issue #42, opened 2023-01-25
Linting


Arnei:
Building the `admin-ui-picard` module via `mvn install` spits out a slew of linting warnings. Maybe we should do something about that.

Arnei:
A big amount of warnings are pertaining to `jsx-a11y/anchor-is-valid` and `jsx-a11y/anchor-has-content`. This is because the new Admin UI thoroughly misuses `<a>` tags as easy-to-stiyle buttons. Fixing this properly will likely require some systematic planning and a lot of CSS rewriting. So much so that I think it's worth its own issue.

PR #189, opened 2023-05-09
Fix eslint warnings about comments in jsx

Arnei:
Fixes `jsx-no-comments warnings by disabling them, as the double slash in this case is not for comments, but for a string.

Helps with #42.


naweidner:
I'm fine with that.

naweidner (approved):

PR #192, opened 2023-05-09
Remove unused columns from server table

### LINKED ISSUE #61: Remove columns from server table
The two columns mean runtime and mean queuetime do not exist anymore in the old Admin UI, so we should probably remove them from the new admin UI as well.

Arnei:
The mean time columns have long been removed in the old admin ui, so we should remove them here as well.

If you do not see the effect you have to clear your persistant storage by deleting cookie and website data.

Resolves #61.


app/src/i18n/org/opencastproject/adminui/languages/lang-de_DE.json
```diff
@@ -1592,9 +1592,6 @@
 				"CORES": "Kerne",
 				"COMPLETED": "Jobs abgeschlossen",
 				"RUNNING": "Jobs in Bearbeitung",
-				"QUEUED": "Jobs in Warteschlange",
```
naweidner, line 1595: Is that on purpose?
"QUEUED" is included in the other translations.

naweidner (approved):
Looks good.

PR #193, opened 2023-05-10
Update CRA to 5

Arnei:
Also updates some related (and unrelated) dependencies.

This should fix the babel-jest version mismatch between the two package.jsons.

Also converts some SVGs because I can't figure out how to set babel configuration in CRA.

Includes #191 to work and as such should not be merged before it.


gregorydlogan:
Fun times, I can't get a working build for #191 *without* this.  So it's getting merged too :)

PR #199, opened 2023-09-07
Add event upload progress

Arnei:
Use notifications to show the progress of an event upload. In order for this to work, this adds the ability to add notifications with certain ids, as well as the ability to override notifications.

### DIFF
--- a/app/src/actions/notificationActions.js
+++ b/app/src/actions/notificationActions.js
@@ -22,6 +22,11 @@ export const createNotification = (notification) => ({
 	payload: { notification, id: nextNotificationId++ },
 });
 
+export const createNotificationWithId = (notification, id) => ({
+	type: CREATE_NOTIFICATION,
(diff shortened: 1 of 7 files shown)


narickmann:
Looks good. You now receive immediate feedback that the event is being uploaded and something is happening. 

One thing: When I upload an event, the notification comes, but without the percentage. Surely the progress should also be displayed? Or did I miss something?

grafik

Arnei:
Should absolutely display a percentage. Will currently only work when setting your Opencast to en_US, as it relies on changes in the translation files (and I only made those changes in the en_US file).

narickmann:
> Will currently only work when setting your Opencast to en_US, as it relies on changes in the translation files (and I only made those changes in the en_US file).

Whoops, I guess I picked the wrong flag then when testing this :D With en_US this works like expected. Nice

narickmann (approved):

PR #200, opened 2023-09-11
Add typescript capabilities

Arnei:
This PR uses [ts-migrate](https://github.com/airbnb/ts-migrate) to add typescript to the app part of the project.

To clarify, this PR only introduces typescript to the application, it does not actually make an effort to properly type components. Thus, components are implicitly typed or still `any`. However, this PR provides the groundwork for introducing proper typing, which can happen bit by bit in follow-up PRs and new PRs. 

This PR adds A LOT of lines mainly due to comments disabling typescript errors, most of which complain about implicit `any`s. I personally think it's a waste of time to add explicit `any`s everywhere, as hopefully proper typing will come later, so I'm in favor of leaving the comments in for now.


owi92 (approved):
I think this should be safe to merge (...he said before everything blew up)

Arnei:
Thanks for the review, merging.

PR #201, opened 2023-09-11
Replace bourbon vendor files with npm files

Arnei:
Removes the bourbon files in the vendor folder, and instead installs the corresponding npm packages.
This has the upside of centralizing our dependencies and avoid hassling around with ruby gems. Unfortunately, it also has the downside of requiring relative paths to the bourbon node modules.


narickmann (approved):
After reviewing #202 I think it is okay to approve.

PR #205, opened 2023-09-15
Fix exception when opening timepicker

Arnei:
If you open a time picker (e.g. "startDate" in the event details), the ui crashes with an exception:
"Format string contains an unescaped latin alphabet character `n` ".

This fixes that by downgrading the date-fns dependency. Allegedly the issues is with material-ui not providing the correct types for date-fns, but downgrading material-ui would cause more problems.

PR #206, opened 2023-09-15
Remove datepicker.scss

Arnei:
Drops datepicker.scss, since we don't use it anywhere. We are instead using material-ui styles.

Requires #205 to test.


KatrinIhler (approved):
Looks good.

PR #209, opened 2023-09-29
Typing /app/src/utils/utils.ts

Arnei:
First foray into typing, starting with simple stuff.

Also adds interfaces for user information, so we should think about
- Do we want to primarily use `interface` or `type`? Or do we want to leave it up to developers to choose what they feel is right?
- Do we want to put interfaces together with code, or do we want seperate files for them?


Arnei:
> Good summary: https://www.youtube.com/watch?v=zM9UPcIyyhQ
 > "If you just like the word "interface", like, you know, you've come from object oriented programming, then use an interface".
 
It me.

But sure, we can do types per default. Will get around to changing that, hopefully.

PR #212, opened 2023-10-12
Fix proxy not working with typescript

### LINKED ISSUE #210: Typescript broke proxy
Seems the typescript PR https://github.com/opencast/opencast-admin-interface/pull/200 broke our proxy server for developing. No requests are being proxied anymore

Arnei:
"setupProxy" relies on CRA auto-magic, but the magic will only work if it has the ".js" extension. And then I also changed something in the tsconfig that seems unrelated to me and now it works...?

Fixes #210 .

### DIFF
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -1,4 +1,8 @@
 {
+  "exclude": [
+    "node_modules",
+    "dist"
+  ],
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig to read more about this file */
 
@@ -40,7 +44,7 @@
 
     /* JavaScript Support */
     "allowJs": false,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
-    // "checkJs": true,                                  /* Enable error reporting in type-checked JavaScript files. */
+    "checkJs": false,                                  /* Enable error reporting in type-checked JavaScript files. */
     // "maxNodeModuleJsDepth": 1,                        /* Specify the maximum folder depth used for checking JavaScript files from 'node_modules'. Only applicable with 'allowJs'. */
 
     /* Emit */
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "exclude": [
-    "node_modules",
-    "dist"
+    "node_modules"
   ],
   "compilerOptions": {
     /* Visit https://aka.ms/tsconfig to read more about this file */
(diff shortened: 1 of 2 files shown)


app/tsconfig.json
```diff
@@ -1,4 +1,8 @@
 {
+  "exclude": [
+    "node_modules",
+    "dist"
```
owi92, line 4: I think you can remove the `dist` line, if I'm not mistaken. At least I don't see any folder with that name, and it seems to work without.

Actually, things still work when I comment out the whole `exclude` array, can you confirm that? If so, I guess you can remove that as well.

app/tsconfig.json
```diff
@@ -40,7 +44,7 @@
 
     /* JavaScript Support */
     "allowJs": false,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
-    // "checkJs": true,                                  /* Enable error reporting in type-checked JavaScript files. */
+    "checkJs": false,                                  /* Enable error reporting in type-checked JavaScript files. */
```
owi92, line 47: Commenting this out again also doesn't seem to make any difference. I mean I guess that makes sense since that should theoretically be equivalent to setting this to `false`. 
I don't think you made these changes without reason though, so maybe I'm doing something wrong while testing this.

owi92 (approved):
Can confirm that this fixes the issue! Though I am uncertain whether the changes in `tsconfig` are really necessary. Well they also don't seem to break anything so maybe just keep them 🤷

Arnei:
Seems the changes to the tsconfig were indeed unnecessary? Weird, could've sworn the exclude was important. But hey, the simpler our config the better.

app/tsconfig.json
```diff
@@ -44,7 +40,7 @@
 
     /* JavaScript Support */
     "allowJs": false,                                  /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
-    "checkJs": false,                                  /* Enable error reporting in type-checked JavaScript files. */
+    "checkJs": true,                                  /* Enable error reporting in type-checked JavaScript files. */
```
owi92, line 43: Err wait, I believe you just need to comment this line out. Setting it to `true` gives me an error in the `tsconfig` file because `allowJs` also needs to be enabled for this. But also, I'm a little confused by this. JS is not allowed (line 42) but `setupProxy.js` is cleary a JS file. Or does that not count as "part of the program"? 🤨

owi92:
Ok hold on, I've just tested this with the redux-toolkit branch I also "reviewed" the other day, and it seems like the `tsconfig` changes you made are necessary after all, at least for that PR (minus the `dist` thing in line 4). I'm really sorry for this back and forth, maybe you can understand my confusion.

I think I would suggest to either drop your latest revert-commit here, or make the corresponding `tsconfig` changes in #213 instead.

owi92 (approved):

PR #214, opened 2023-10-16
Modernize redux

Arnei:
Resolves #213.

This PR aims at laying the groundwork for modernizing redux. Thankfully, the modernization can happen incrementally, so this PR does not need to and likely will not update everything at once. It will however try to touch all the relevant parts so it can serve as a guideline for future PRs that seek to continue with the modernization.

As such, discussions on code styles and best practices are very welcome.

PR #219, opened 2023-11-02
Refresh workflow operations in event details

### LINKED ISSUE #218: Workflow ops in event details do not live update
If you want to look at the progress of a workflow in real time in the Admin UI under Actions -> Workflows -> Details, the view remains static. I.e., the current state of processing is displayed when the page is called. The display does not automatically switch to the next step when a step has been successfully completed.

_Originally posted by @mwuttke in https://github.com/opencast/opencast-admin-interface/issues/69#issuecomment-1759331806_

Arnei:
Fixes #218.

In the old admin ui, the status of workflow operations in the event details was updated in real-time, allowing you to "watch" your workflow running. This updates our admin ui to do the same by re-fetching workflow operations every 5 seconds.

Unfortunately, this is causing rerenders even if the response from the Opencast backend has not changed between calls. We could probably memoize this somehow, but I can't find a straightforward way, at least not one that is more straightfoward than switching to redux toolkit, like in #214. Therefore, it might be worth to wait until #214 gets merged, then implement redux toolkit here as well.

### DIFF
--- a/app/src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowOperations.tsx
+++ b/app/src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowOperations.tsx
@@ -1,12 +1,12 @@
 import { connect } from "react-redux";
-import React from "react";
+import React, { useEffect } from "react";
 import Notifications from "../../../shared/Notifications";
 import {
 	getWorkflow,
 	getWorkflowOperations,
 	isFetchingWorkflowOperations,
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Not that I use this much, but man that rerender is brutal.  Bringing that refresh rate low enough could trigger seizures if we're not careful!

Arnei:
Agreed. Should look into reducing rerenders in the future (ideally after the associated "Modernize redux" PR is merged)

PR #220, opened 2023-11-03
Show conflicting scheduling events for new events

### LINKED ISSUE #132: Show user with which events they have a scheduling conflict
When scheduling a new event, you can conflict with other events. The old Admin UI would show you with which events you are in conflict:

Image

The new Admin UI does not do this:

Image

But it should.

Follow-up to #68.

Arnei:
Fixes #132

When trying to schedule a new event, there already is a notification if the new event would cause a conflict with an already scheduled event. However, it would not tell you with which event, leaving you guessing on how to resolve the conflict. This display the conflicting event, same as when trying to reschedule an event in the event details.


gregorydlogan:
Testing this I saw:
- False positive warnings where it said I had a conflict when I hadn't scheduled anything at all yet
- The warning did not go away when the desired time changed
  - The false positive might be responsible for this too, manually closing the warning made it go away for a screen or two, but then it'd be back
 - Uncaught Runtime Errors about state mutations.  I see this one a lot, across all reviews.

Arnei:
Thanks for the review, will look into it.

Arnei:
> Uncaught Runtime Errors about state mutations. I see this one a lot, across all reviews.

If it happens across many different PRs, it is likely it's own issue?

app/src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
```diff
@@ -103,6 +105,32 @@ const NewSourcePage = ({
 					<div className="full-col">
 						{/*Show notifications with context events-form*/}
 						<Notifications context="not_corner" />
+
+            {
+              /*list of scheduling conflicts*/
+              conflicts.length > 0 && (
```
gregorydlogan, line 111: The main conflict blob is appearing, but this doesn't ever get added to the DOM.  Searching for `scheduling-conflict` doesn't find any elements.

gregorydlogan (changes requested):

app/src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
Arnei: "Works on my machine":
Bildschirmfoto vom 2024-02-29 14-55-32

Maybe you could share more details on how exactly you are testing this?

Arnei: Well, i dare say your normal is not expected behaviour. :D

Could it be that proxying to docker is somehow the problem? Seems to be main difference between our systems, and that proxy is a fickle thing.

Issue #213, opened 2023-10-13
Modernize React Redux


Arnei:
Our redux code is still "oldschool", meaning it's unwieldy, there's a hell of a lot of boilerplate and it's nearly impossible to type well. Therefore we should try to migrate to "modern" redux, using redux-toolkit.

Also recommended by the developers of React Redux: https://redux.js.org/usage/migrating-to-modern-redux

Arnei:
#214 did not fully resolve this issue, so reopening. This is one of those issues that gets resolved over multiple PRs anyway. #214 does however serve as a reference on how to further resolve this issue.

PR #242, opened 2023-12-19
Modernize redux: userSlice

Arnei:
Switching to redux toolkit for getting users for the user table.

Helps with #213, based off of #214.


owi92 (approved):
As requested, I tested these redux PRs (without looking at the code this time), just to see if anything breaks with these. We talked about any issues I encountered internally, so I won't repeat those here (since they are either unrelated to a specific PR or have been fixed). If anyone is interested, my findings and Arne's comments are documented [here](https://hackmd.io/39AIpQGPQG6rZg3BGcDgIA).
These have been open for quite a while now and needed to get tested by another person before they are merged. Since everything seems to be working I'll approve this and the other ones.
You'll see this a lot on the other redux PRs, this will be my only comment mentioning the above stuff.

PR #243, opened 2023-12-20
Modernize redux: userDetailsSlice

Arnei:
Helps with #213.

Switching to redux toolkit for getting a usersdetails for the user details modal.

Helps with #213. Includes two commits from #242.


owi92 (approved):
User details modal appears to be working fine.

PR #246, opened 2023-12-21
Remove useAppDispatch from tableThunks.ts

### LINKED ISSUE #245: Error when trying to switch pages
An Invalid Hook Error `Hooks can only be called inside the body of a function component` occurs when trying to switch pages in the event view, preventing the user from switch pages.

Arnei:
Fixes #245, which was caused by #214.

Those thunks are likely passed to a `useReducer` somewhere, so we can't have hooks here. We will still want to use useAppDispatch here eventually, but it can wait until we migrated everything else to redux toolkit and can finally get to the tables.


Arnei:
Merging this without review, because:
- The bug this fixes is pretty critical
- I am extremely certain that this works
- It cannot make anything worse than it already is

PR #257, opened 2024-01-05
Modernize redux: recordingSlice

Arnei:
Helps with #213.

Switching to redux toolkit for getting recordings for the recording table (which is translated as "Location" in the ui).


owi92 (approved):
Locations/recordings table is working. (See https://github.com/opencast/opencast-admin-interface/pull/242#pullrequestreview-1882338091 for more info on these approvals)

PR #268, opened 2024-01-11
Dont store scheduling source as date in redux

Arnei:
While storing dates as date objects in redux does not necessarily break anything, it can interfere with some features (persisting, debugging) and lead to bad practices (for an encompassing discussion see https://github.com/reduxjs/redux-toolkit/issues/456).

This changes the scheduling info fetching method to store dates as strings.


app/src/thunks/eventDetailsThunks.ts
```diff
@@ -871,12 +871,12 @@ export const fetchSchedulingInfo = (eventId) => async (dispatch, getState) => {
 		const source = {
 			...schedulingResponse,
 			start: {
-				date: startDate,
+				date: schedulingResponse.start,
```
gregorydlogan, line 874: So, for me these render as (DD/MM/YY HH:MM) when the modal opens, but then clicking on them to change them switches to the long date (Feburary 22nd 11pm).  This is quite confusing IMO - I would not expect the format to change when clicking in.

gregorydlogan (changes requested):

PR #270, opened 2024-01-16
Remove redux persist

### LINKED ISSUE #231: Error that is resolved by clearing browser cache
Some users have been getting a completely broken ui and an error in the console logs like `RangeError: Format string contains an unescaped latin alphabet character `n``.

I'm not sure, but this is likely cause by redux persist. If users persist and old redux state in their browser, and then open a newer version where there have been changes to the redux state structure, this may cause issues. This requires futher testing though.

If it truly is a problem with redux persist, I wouldn't mind removing it, since I'm not exactly sure what we're hoping to achieve by using it anyway.

Arnei:
Aims to resolve #231

Redux persist in its current form is causing to many problems to keep it around. Therefore this simply
removes it.

The most common problem with redux persist is
that users make changes to configuration, which
are then not taken over into the ui as they are blocked by the existing redux state. This can also happen
when new versions of the ui are released. Resolving this requires a hard page reload which is simply un- userfriendly.

Whatever efficiency benefits redux persist may bring, they are not worth requiring clearing your browser cache every other day.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -40,7 +40,6 @@
     "react-select": "^5.7.3",
     "redux": "^4.2.1",
     "redux-devtools-extension": "^2.13.9",
-    "redux-persist": "^6.0.0",
     "redux-thunk": "^2.4.2",
     "reselect": "^4.1.8",
     "styled-components": "^5.3.10",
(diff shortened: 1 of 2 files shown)


Arnei:
Putting this on hold because I finally noticed that redux-persist is responsible for remembering changes to table filters or column de-/activity and ordering.

Arnei:
Closing this because throwing out redux-persist without any sort of replacement for saving info in localstorage is not really a great solution after all.

PR #279, opened 2024-01-25
Fix workflow tab in event details not loading

Arnei:
Cause by passing undefined to a function which could not handle undefined as a parameter. Hopefully we can get typing soon.


Piriya-VM:
I couldn't reproduce this bug. But it seems to be fixed in OC 15.2 (?) 🤔

PR #283, opened 2024-02-06
Add notes column back in

### LINKED ISSUE #50: Notes column is missing
In the old Admin UI, there is an optional "Notes" column for the event details view. This is completely absent from Admin UI Picard. ETH will probably want it back.

Arnei:
Resolves #50

Add a column called  "notes" to  the event table. This was already present in the old admin ui. The column provides a small text area where a user can take short on an event. The notes are saved as a comment (and can thus also be viewed in the event details).

The notes column is disabled by default, and needs to be manually enabled. Only if it is enabled will additional comment information be fetched from the backend,
which puts a bit of additional load on the Opencast admin if you have a lot of comments.

Known issues:
- ~This does not play nice with redux-persist. You will need to clear website data to test. Removing redux-persist like in #270 also poses an issue, as this will reset your column settings everytime you reload the page (meaning you will need to manually enable the notes column every time you load the page).~
  - **WONT-FIX**: You will need to clear website data to test, because this does not play nice with redux-persist and getting redux-persist to play nice is way too much work for a change like this that happens once every blue moon.
- ~When enabling the notes column, the text ares take several seconds to show up.~
  - Fixed.


Piriya-VM (approved):
Works for us 👍🏾 Thank you :)
Bildschirmfoto vom 2024-04-19 10-42-20

KatrinIhler:
Gonna trust Piri's review on this ¯\_(ツ)_/¯

PR #286, opened 2024-02-08
Respect "displayFallback"

Arnei:
The admin ui documentation for "Asset Upload" specifies both "displayOverride" and "displayFallback" as ways to replace translations. We were only respecting "displayOverride" and completely ignoring "displayFallback". This changes that.

To test this, you should have access to an Opencast instance where you can modify
listproviders/event.upload.asset.options.properties. Then add something like the following line to that file

`EVENTS.EVENTS.NEW.SOURCE.UPLOAD.NON_SEGMENTABLE_PART0={"id": "track_presenter_part0","type":"track", "flavorType": "presenter","flavorSubType": "source+part-0", "multiple":false, "displayOrder":20, "accept": ".avi,.flv,.m4v,.mkv,.mov,.mp4,.mpeg,.mpg,.ogv,.webm,.wmv", "displayFallback.SHORT": "Multipart: Presenter Part 1", "displayFallback.DETAIL": "This option should be chosen, if you want to concatenate multiple presenter videos."}`


KatrinIhler:
Does what it says on the tin. :) Gonna merge #301 which includes this one.

PR #287, opened 2024-02-09
Fix startDate table filter

Arnei:
This fixes an issue where users had trouble confirming
a start date filter in the event table (should also fix it in
the series table).

Previously, the date filter would only be confirmed if
the end date was set through the calendar picker while
the start date was already set. This makes it so that it
does not matter which calendar picker is opened first.
Furthermore, it adds the ability to confirm the filter by
pressing "Enter" (all provided the both given dates are
valid). This should increase usability and should also be
closer to what users are used to from the old admin ui.

Includes the update to material ui version 5 from https://github.com/opencast/opencast-admin-interface/pull/277,
to avoid doing this for different material ui version.


Piriya-VM:
Works for us in OC 15 :)

PR #288, opened 2024-02-09
Fix tabbing in event details metadata

### LINKED ISSUE #233: Tab index not working/existing?
In the old Admin UI, you can use Tab to navigate through quite a number of UI elements, for example the fields in the metadata tab. At least that doesn't seem to be working anymore.

Arnei:
In the event/series details metadata tab, it was not possible to navigate between the different items using the "tab" key. This is now possible.

Fixes #233.


KatrinIhler (changes requested):
Unfortunately, this isn't working for me. I'm testing in Firefox with a proxy against stable.opencast.org. When creating new series or events, I can edit the title, but then no other metadata field can be edited, neither with tabbing nor with clicking into the field or clicking on the little edit icon on the right of the field. Also the metadata of existing events doesn't seem to be editable anymore as well. :(

gregorydlogan:
This seems to work correctly for me now.

PR #289, opened 2024-02-09
Fix typescript complaints about tabIndex

Arnei:
The react type definitions for the tabIndex prop want tabIndex to be a number, but we're using strings. This changes it so we use numbers. This should not change behaviour as technically using strings for tabIndex is fine, this is just to appease typescript.


owi92 (approved):
Don't see any issue with this change. And for once, glancing over the code, there really shouldn't be.
This is the last comment of my "approval bombing".
(See https://github.com/opencast/opencast-admin-interface/pull/242#pullrequestreview-1882338091 for more info on these approvals)

PR #295, opened 2024-02-20
Replace logout request with redirect

geichelberger:
Restore the old logout procedure by redirecting to `/j_spring_security_logout`, which is required for SAML2 logout.

https://github.com/opencast/opencast/issues/5612


Arnei (approved):
I can't tell or test if this solves the linked issue. But it does what is says and also seems to simplify our code a bit, and that is good enough for me.

PR #300, opened 2024-03-05
Fix user not having access to created event

Arnei:
When creating a new event (or series), the access tab will suggest to create it with the role `ROLE_USER_ADMIN`. If you are not an admin, this means you won't be able to access the created event later on.

Instead of always setting `ROLE_USER_ADMIN` as the default, this tries to get the user role from `/info/me.json` instead and suggest that as the default. Should be more in line with how it works in the old admin ui too


Piriya-VM:
Looks good to us 👍🏾

PR #301, opened 2024-03-05
Sort asset upload options by displayOrder

Arnei:
User can specify a display order in the event upload asset listprovider, but we do not honor it. With this change we honor it.

This contains #286 for the sake of developing and testing convenience. The actual change is only 10da02f96b917c6c0e670064a2543f0505eaaf17.


KatrinIhler:
Both parts of this work as expected -> merging :)

PR #304, opened 2024-03-06
Modernize redux: tableFiltersSlice

Arnei:
Helps with #213.

Switching to redux toolkit for handling table filters.


ziegenberg:
@Arnei can you rebase, I would do a review afterwards

src/utils/resourceUtils.ts
```diff
@@ -11,7 +11,11 @@ import { NewUser } from "../slices/userSlice";
 import { Recording } from "../slices/recordingSlice";
 import { UserInfoState } from "../slices/userInfoSlice";
 import { hasAccess } from "./utils";
+<<<<<<< HEAD:app/src/utils/resourceUtils.ts
+import { RootState } from "../store";
+=======
 import { MetadataCatalog } from "../slices/eventSlice";
+>>>>>>> main:src/utils/resourceUtils.ts
```
ziegenberg, line 18: I think that might need some fixing :)

ziegenberg (changes requested):

PR #307, opened 2024-03-15
Fix workflow operations not refreshing

### LINKED ISSUE #294: EventDetailsWorkflowTab does not properly rerender
Due to https://github.com/opencast/opencast-admin-interface/pull/219, the Workflow tab in the event details now rerenders frequently. However, the rerender looks terrible (to the point where we might have to issue an epilepsy warning) and causes the page to jump to the top. Ideally, the rerender should not be noticable at all. 

If possible wait with fixing this after https://github.com/opencast/opencast-admin-interface/pull/269 is merged, as a fix for this issue is highly likely to conflict with it.

Arnei:
In the event details workflow tab, this fixes workflow operations not refreshing (again, was broken by redux modernization). Furthermore, this fixes the rather harsh rerender whenever the workflow operations were fetched again.

Resolves #294.
[Bildschirmaufzeichnung vom 2024-03-15, 14-41-20.webm](https://github.com/opencast/opencast-admin-interface/assets/14070005/c0133d44-ef3e-4d70-b1a7-95459e95b6be)

### DIFF
--- a/app/src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowOperations.tsx
+++ b/app/src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowOperations.tsx
@@ -3,12 +3,12 @@ import Notifications from "../../../shared/Notifications";
 import {
 	getWorkflow,
 	getWorkflowOperations,
-	isFetchingWorkflowOperations,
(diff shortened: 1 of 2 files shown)


Piriya-VM:
Thanks for this fix. :)
It works. Now I can watch the status of workflow operations in real time :)

PR #308, opened 2024-03-15
Remove various ts expect errors

Arnei:
We still have a couple of comments that disable typescript errors. This is a collection of various low-effort fixes for said errors.

Purposefully ignores some errors that are already fixed in other pull requests or require work on a larger scale.


wsmirnow:
This PR breaks the build.

```
rm -rf build/ && CI=false npm run build

> admin-ui-picard@0.1.0 build
> react-app-rewired build

Creating an optimized production build...
One of your dependencies, babel-preset-react-app, is importing the
"@babel/plugin-proposal-private-property-in-object" package without
declaring it in its dependencies. This is currently working because
"@babel/plugin-proposal-private-property-in-object" is already in your
node_modules folder for unrelated reasons, but it may break at any time.

babel-preset-react-app is part of the create-react-app project, which
is not maintianed anymore. It is thus unlikely that this bug will
ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to
your devDependencies to work around this error. This will make this message
go away.
  
Failed to compile.

TS2345: Argument of type 'EventTarget | null' is not assignable to parameter of type 'Node | null'.
  Type 'EventTarget' is missing the following properties from type 'Node': baseURI, childNodes, firstChild, isConnected, and 43 more.
    116 | 			if (
    117 | 				containerAction.current &&
  > 118 | 				!containerAction.current.contains(e.target)
        | 				                                  ^^^^^^^^
    119 | 			) {
    120 | 				setActionMenu(false);
    121 | 			}
```

Arnei:
Thanks for testing @wsmirnow , should be fixed now.

PR #325, opened 2024-04-11
Add typing to NewEventWizard.tsx

Arnei:
Typescript for new event wizard and the components it uses.


ziegenberg:
Clicked through, and it still seems to work.

PR #326, opened 2024-04-18
Fix the release script for macOS

JulianKniephoff:
`date --utc` is not POSIX and macOS doesn't have it. `-u` is the standard equivalent.


ziegenberg (approved):
Looks right to me.

PR #331, opened 2024-04-24
Select correct workflows for Start Task

KatrinIhler:
Currently we show workflows for the tags `schedule` and `upload` in the Start Task modal, when we want to show those with `archive`.


gregorydlogan:
This fixes #335

PR #345, opened 2024-05-06
Simplify the proxies

JulianKniephoff:
This contains a bunch of simplifications in my quest to boil down the whole proxy setup to something more reasonable. See individual commits for more details.


ziegenberg:
It does what it says on the tin, and it works.

PR #346, opened 2024-05-06
Make the port the client proxy proxies to configurable

JulianKniephoff:
This is based on #345, so that one needs to go in first.

Otherwise it does what it says on the tin.


ziegenberg:
Works.

PR #347, opened 2024-05-06
Add simple CI test

lkiesow:
This patch adds a simple CI workflow to test if the dependencies can be downloaded and the project actually builds.


ziegenberg:
@lkiesow, would this be ok to be merged now?

lkiesow:
The build still fails with warnings. We need to fix that first

ziegenberg:
Could we add this with `CI=false` so we at least know it builds, and when we remove all warnings, we can set `CI=true`? So we can get gradually more CI.

ziegenberg:
And it already found a build error... :D

Will be fixed by #404

ziegenberg:
But as they are unrelated I will merge this now.

PR #348, opened 2024-05-06
Update Node.js to current LTS

lkiesow:
This patch updates the Node.js version used for creating releases to the current LTS version which is Node.js 20.


ziegenberg (approved):

ziegenberg:
I think this looks good to me. Let's use this for the next (imminent) release 🥳

PR #349, opened 2024-05-06
Fail on warnings when building releases

lkiesow:
This patch removes the forced `CI=false` which means that the build process will fail on compiler warnings.


ziegenberg:
@lkiesow As all the warnings on the build are now fixed (thx to #430) I think we can now merge this.

ziegenberg (approved):

PR #355, opened 2024-05-07
Change the default port for the development servers

JulianKniephoff:
5000 is used by Control Center on macOS, so this will likely annoy
at least a few developers.

This is based on #346, so that should be reviewed and merged first.


JulianKniephoff:
This is made obsolete by #398 (and indirectly #394).

PR #386, opened 2024-05-09
Fix unique key issue in WizzardStepper

ziegenberg:
Fixes a web console complaint about missing unique keys in the WizzardSteppers. I deem using .filter() a nicer solution than a ternary operator that checks if the hidden property of the current label object is false and rendering an empty fragment. This is probably also what made the console unhappy.

### DIFF
--- a/src/components/shared/wizard/WizardStepper.tsx
+++ b/src/components/shared/wizard/WizardStepper.tsx
@@ -69,8 +69,7 @@ const WizardStepper = ({
 			connector={false}
 			className={cn("step-by-step", classes.root)}
 		>
-			{steps.map((label, key) =>
-				!label.hidden ? (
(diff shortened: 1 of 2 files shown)


Arnei (changes requested):
This seems to break whatever arcane magic the WizardStepper is using. In the image below I'm in the "source" step, but the WizardStepper believes me to be in the "asset upload" step.

Bildschirmfoto vom 2024-05-14 13-57-06

ziegenberg:
Rebased.

Arnei:
Can't reproduce the web console warning anymore, so we can probably close this. @ziegenberg, what say you?

ziegenberg:
Yeah, it was fixed in another way. So, closing it is.

PR #387, opened 2024-05-09
Typing WizardStepperEvent

ziegenberg:


Arnei (approved):
Looks good to me.

Does make me wonder why `formik` and `setCompleted` are going unused and if that is intentional. Maybe worth a follow-up issue?

ziegenberg:
> Does make me wonder why `formik` and `setCompleted` are going unused and if that is intentional. Maybe worth a follow-up issue?

@Arnei Would you like to create the issue?

Arnei:
Sure

PR #389, opened 2024-05-10
always add current user to acl

### LINKED ISSUE #385: User can select ACL template and kick himself out the ACL
When uploading an event, the user can select an acl template. If doing so, it is possible that he is not in the acl anymore and cant upload the event. Happens easily for non admin user:

image

The current user setting the acl should always be in the acl and this should not be overwritten when selecting an acl template.

schuettloeffel-elsa:
Closes #385


ziegenberg:
Shouldn't this also be added to `app/src/components/users/partials/wizard/AclAccessPage.tsx` ?

app/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
```diff
@@ -251,7 +251,15 @@ const ResourceDetailsAccessPolicyTab : React.FC <{
...
+		// always add current user to acl since template could lock the user out
+		template = template.concat({
+			role: user.userRole,
+			read: true,
+			write: true,
+			actions: [],
+		});
 
 		setFormikFieldValue("policies", template);
 		setFormikFieldValue("template", templateId);
```
ziegenberg, line 265: I `app/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx` we do check the ACLs with `await dispatch(checkAcls(formik.values.acls));`. Shouldn't we do this here too?

schuettloeffel-elsa:
> Shouldn't this also be added to `app/src/components/users/partials/wizard/AclAccessPage.tsx` ?

No, when creating a user the user gets his own role in any case, even when no role was selected. Not sure where this happens, maybe in the backend?

app/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
schuettloeffel-elsa: Dont know, but better make an own issue for that and stick to this issues topic?

schuettloeffel-elsa:
Hei @ziegenberg, could you have another look into this pr? Would be cool if we could get this to the next release since this is kind of crucial for our oc setup

ziegenberg:
Would an admin be able to select an ACL template without being in the list as the current user?

schuettloeffel-elsa:
> Would an admin be able to select an ACL template without being in the list as the current user?

Selecting a template would always add your user to the list, but you can off course delete that entry.

Arnei (approved):
Works and makes sense to me.

I agree that this does not need to be added to `app/src/components/users/partials/wizard/AclAccessPage.tsx`. Unlike for events or series, ACL templates have no sense of ownership, meaning users don't need their role in the template to edit it.

ziegenberg:
I'm ok with the change, as it adds feature parity with the old Admin UI.

For later reference: This kind of business logic should ideally be placed in the backend and not the frontend. So we do not have to implement this kind of logic in all interfaces again.

ziegenberg (approved):

PR #392, opened 2024-05-13
Harmonize date rendering between read-only and read/write fields

JulianKniephoff:
Metadata fields, that is. This expands on f62e5dac7b5bd080b067b0b7cdc9e4bb74950d22 (itself a late addition to #277) and fixes a time drift between the read-only "Created" field and the non-edit-mode variant of the Start field (the values of which are synchronized) caused by the two methods of formatting interpreting the source date in different timezones.


Arnei (approved):
Works and makes sense to me, thanks!

Arnei:
@ziegenberg Please explain why you assigned me to this PR.

ziegenberg:
I've seen it on other opencast projects that the person merging was set at assignee. So I did this here too. That's all to it.

Arnei:
After they were already done merging? Interesting. Will ask in the Technical Meeting what the purpose behind that is, thanks for the info.

PR #393, opened 2024-05-14
Further simplify the static file server

JulianKniephoff:
This avoids having to poorly reimplement `express.static` using some internal request rewriting logic.


Arnei (approved):
Works!

PR #394, opened 2024-05-14
Remove the proxy server

JulianKniephoff:
Well, the outer one at least. For context: This app had two proxies:

One comes with `create-react-app` as a middleware in the dev server, proxying all the requests the frontend makes (except for the static JS, CSS, etc. assets this project is generating) to some other server.

In production, Opencast delivers the admin UI assets and the backend under the same URL, but in development this is not true. The development server delivers the assets and the backend is ... well it's complicated.

So far you either started a static file server which was acting as a mock backend, or you started **another** proxy server that piped the requests through to an Opencast instance of your choosing while tacking on some digest authentication.

Well, the dev server proxy can't (easily) do digest auth, but the fact that we insist on digest auth is a historical artifact, anyway, at least as far as I know. It can do basic auth, though, and it **is** a proxy, so it can do everything we need!

This also updates the `README` appropriately and cleans it up some in general (though not completely).


README.md
```diff
@@ -1,85 +1,64 @@
...
 
-You can then start the client in a different tab by running:
+Not all functionality of the admin UI works in this mode. If you need to test with real data,
+or need the ability to change it, you can rely on the proxy functionality of said development server,
+instead of running the static file server. Run:
 
-```shell
-npm run client
+```sh
+PROXY_TARGET=https://develop.opencast.org PROXY_AUTH=jdoe:opencast npm run client
```
Arnei, line 38: Running `PROXY_TARGET=https://develop.opencast.org PROXY_AUTH=jdoe:opencast npm run client` results in the browsers asking me to authenticate myself. Maybe we should use a user that, unlike jdoe, actually exists on develop in order to avoid confusion?

Arnei (changes requested):

README.md
```diff
@@ -1,85 +1,64 @@
...
--------
-
-Before starting, get the project dependencies by running  `npm ci` beforehand both in the root and `/app` directory.
-
-To test with real data run:
-
-```shell
-npm run proxy-server http://stable.opencast.org *opencast_digest_username* *opencast_digest_password*
+```sh
+npm start
```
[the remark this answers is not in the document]
JulianKniephoff: My bad, I shouldn't work on 3 versions of the same thing at the same time. 😅 Pushed a fix.

README.md
```diff
@@ -1,85 +1,64 @@
...
-on the port 5000. If this is used already (as it is on macOS)
-you can specify an alternative port in the `PORT` environment variable,
-for example:
-
-    PORT=5001 npm run proxy-server ...
-
-Note that you need to specify the same port when running the client,
-this time in the `PROXY_PORT` variable:
+```sh
+PORT=3001 PROXY_TARGET=http://localhost:5001 npm run client
```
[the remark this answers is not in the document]
JulianKniephoff: Does https://github.com/opencast/opencast-admin-interface/pull/394/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R41-R44 not explain it?

README.md
JulianKniephoff: I think the paragraph below makes it sufficiently clear that you should set this to sensible user credentials, though. 🤔

Arnei: Not sure why you would want to avoid `admin` as an example. Since admin has access to everything, it easily let's you test most things. Also `admin:opencast` is very much public knowledge.

JulianKniephoff: Because it is the default. You don't have to specify `PROXY_AUTH=admin:opencastz`

README.md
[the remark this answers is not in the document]
Arnei: Not really. Previously, `PROXY_TARGET` was pointing at my Opencast instance. Now, `PROXY_TARGET` is used to point at the development server. So what variable should I use to point at my Opencast instance, and in which command?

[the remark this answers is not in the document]
JulianKniephoff: Previously `PROXY_TARGET` did not exist. `PROXY_TARGET` is not used to point to the development server, but to anything that replies to the backend requests. That can be (and is by default) the static file server at `localhost:5000` (by default), but can also be `https://develop.opencast.org` as demonstrated by the command in the section above, or `http://localhost:8080` as stated by the remark below that command. 🤔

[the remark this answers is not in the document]
Arnei: Hmm, maybe my brain is just too stuck in ye olde ways of thinking.

README.md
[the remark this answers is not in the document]
Arnei: Fix confirmed

README.md
Arnei: True, but someone who is just reading the ReadMe wouldn't know that, would they?

JulianKniephoff: That is fair. I amended the explanation.

JulianKniephoff: And moved things around a bit more.

README.md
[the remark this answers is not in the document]
JulianKniephoff: That's fair, it's a bit of a misnomer. Luckily that command will disappear completely in #398. 😉

README.md
Arnei: I still don't understand why you are so adamant on making this command not "just work tm". Why force everyone who copies this command to alter or remove PROXY_AUTH? Espeically if we want people to be able to test pull requests as easy as possible?

Arnei: Seems I was looking at an out-of-date version, sorry.

Arnei (approved):

README.md
JulianKniephoff: I happens and I'm not making it particularly easy force-pushing every 5 seconds. 🙈

PR #398, opened 2024-05-16
Pull the static file server into the CRA proxy as well

JulianKniephoff:
With this, the outer npm-project is finally obsolete. See the updated `README` for how to use it and the proxy server. I abstained from restructuring the project any further to keep the PR focused, but a PR that will do nothing but move things around is in the pipeline.

Note, this is based on #394.


package.json
lkiesow, line 1: Removing this, we need to remove this from the CI as well:

https://github.com/opencast/opencast-admin-interface/blob/34dd129025a87002328e853e9b77b599fc7bc5d8/.github/workflows/test.yml#L20-L22

app/src/setupProxy.js
```diff
@@ -1,22 +1,40 @@
...
+        if (req.method !== "GET") {
+            req.originalMethod = req.method;
+            req.method = "GET";
+            setTimeout(testFiles, 1000, req, res, next);
+        } else {
+            testFiles(req, res, next);
+        }
+    };
+
+const testFiles = express.static(path.join(__dirname, "../../test"));
```
lkiesow, line 40: The path is missing the `/app`:
```suggestion
const testFiles = express.static(path.join(__dirname, "../../test/app"));
```

Without this change, running `npm start` will immediately end in an error.

package.json
JulianKniephoff: To be fair, this wasn't there when I first created this. ;D

app/src/setupProxy.js
```diff
@@ -1,22 +1,40 @@
...
+        target: process.env.PROXY_TARGET || "http://localhost:8080",
+        changeOrigin: true,
+        auth: process.env.PROXY_AUTH || "admin:opencast",
+    })
+    : (req, res, next) => {
+        if (req.method === "POST") {
+            res.status(201);
+        }
+
+        req.url = `/${req.method}/${req.originalUrl}`;
```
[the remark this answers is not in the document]
JulianKniephoff: This was probably caused by trying and reverting lots of different things. Fixed now.

app/src/setupProxy.js
JulianKniephoff: And this was probably missed when separating this from #399.

lkiesow (approved):

PR #399, opened 2024-05-16
Simplify the project structure

JulianKniephoff:
With the refactoring of the development servers done in #398
and indirectly #394, there is an opportunity to boil down
the project structure a bit.

This PR does nothing but move files around and make the necessary
changes to keep everything working. (Fingers crossed.)

Note that this is based on #398!


JulianKniephoff:
Converting to draft until #398 is merged.

.crowdin.yaml
```diff
@@ -4,5 +4,5 @@ base_path: .
 preserve_hierarchy: true
 
 files:
-  - source: '/app/src/i18n/org/opencastproject/adminui/languages/lang-en_US.json'
-    translation: '/app/src/i18n/org/opencastproject/adminui/languages/lang-%locale_with_underscore%.json'
\ No newline at end of file
+  - source: '/src/i18n/org/opencastproject/adminui/languages/lang-en_US.json'
+    translation: '/src/i18n/org/opencastproject/adminui/languages/lang-%locale_with_underscore%.json'
```
ziegenberg, line 8: This file could benefit from a newline at the end :)

.github/workflows/create-release.yml
```diff
@@ -20,16 +20,14 @@ jobs:
 
       - name: download dependencies
         run: npm ci
-        working-directory: ./app
 
       - name: create release tarball
         run: ../.github/build-release.sh
-        working-directory: ./app
```
ziegenberg, line 26: Are you sure the command `../.github/build-release.sh` does not need to be changed when we alter the working directory?

.gitignore
```diff
@@ -1,25 +1,2 @@
...
-.pnp.js
-
-# testing
-/coverage
-
-# production
 /build
-
-# misc
-.DS_Store
```
ziegenberg, line 15: Are we sure we don't want those any longer ignored? Especially `.DS_Store` and `.vscode`?

README.md
```diff
@@ -13,7 +13,6 @@ To get a local copy of the admin UI to test or develop on, you can do the follow
 git clone git@github.com:opencast/opencast-admin-interface.git opencast-admin-interface-demo
 cd opencast-admin-interface-demo
 git switch my-branch  # or otherwise check out, pull, merge, etc. whatever branch you want to test/hack on
```
ziegenberg, line 15: Not related to this change but as I just saw it...

```suggestion
git switch -c my-branch  # or otherwise check out, pull, merge, etc. whatever branch you want to test/hack on
```

At this point the branch `my-branch` probably does not exist yet, so we might need to create it.

ziegenberg (changes requested):

ziegenberg:
I thoroughly clicked through the admin UI, and everything seems to work. I did not come across new bugs I didn't see before. So when the few comments above are addressed, I think this can be merged.

.crowdin.yaml
JulianKniephoff: That wasn't me, but it's totally fair. 😊

.github/workflows/create-release.yml
JulianKniephoff: Good catch! 👀

README.md
JulianKniephoff: This is a fictitious command as the comment explains, I don't want to make assumptions about what the user wants to do. I also don't care that much though, so if you insist, I will add `-c`, but this gets a -0 from me. x)

.gitignore
JulianKniephoff: This is a philosophical discussion, I guess. I'm usually against these generated/template ignore files, which this one clearly was, ignoring files that will never ever come up in the context of the project. I also don't think it is the job of `.gitignore` to ignore files that we **anticipate** to be created by tools the developers might use outside the scope of this project. That's every dev's own responsibility IMO, and that's what `.git/info/exclude` or the global versions thereof are for.

Anyway, I **do** see the practical value of doing so for the most common things so things don't slip through the cracks. If people insist that we do it I can add some things back, but then I want those people to suggest a reasonable list of things.

Fun fact: `.vscode` was part of this repo up until this PR even though it was in **TWO** `.gitignore` files ... Do with that what you will. x)

ziegenberg (approved):
Let's give this a try. 🙈

PR #401, opened 2024-05-17
dont display ui when logged out

### LINKED ISSUE #397: UI is displayed after logout as if logged in
After logging out it is possible to use the browsers back button to display the UI again. This can lead to creation of too many cookies when using shibboleth login, and it is confusing to the user anyway. The UI should recognize when not logged in.

schuettloeffel-elsa:
Closes #397


schuettloeffel-elsa:
Ah, maybe better redirect to the loginpage instead..

app/src/slices/healthSlice.ts
```diff
@@ -57,6 +57,12 @@ type FetchHealthStatusResponse = {
 // Fetch health status and transform it to further use
 export const fetchHealthStatus = createAsyncThunk('health/fetchHealthStatus', async () => {
 	const res = await axios.get<FetchHealthStatusResponse>("/services/health.json");
```
gregorydlogan, line 59: You want to parse `/info/me.json` and look for `ROLE_ADMIN_UI` in the roles field.  It's possible to have an admin ui user who can see the UI, but *not* the health metrics.

gregorydlogan (changes requested):

app/src/slices/healthSlice.ts
schuettloeffel-elsa: Ok thanks, I gave it another try

schuettloeffel-elsa: Hei @gregorydlogan can you have another look at the new draft?

gregorydlogan:
This looks reasonable to me, haven't had a chance to test.  Take this out of draft mode when you're ready and we can take a look.

PR #406, opened 2024-05-21
fix i18n.options.fallbackLng

ziegenberg:
We used `en-US` as a source for all translations but had `en-GB` as a fallback. This is now fixed.


lkiesow (approved):
We use `en-US` everywhere else. Looks good.

PR #407, opened 2024-05-21
Better naming for Crowdin workflows

lkiesow:
Let all GitHub Actions workflow definitions for Crowdin…

- have a filename starting with `crowdin-…`
- Have a name starting with `Crowdin » `

### DIFF
--- a/.github/workflows/crowdin-download-translations.yml
+++ b/.github/workflows/crowdin-download-translations.yml
@@ -30,14 +30,7 @@ jobs:
 
       - name: add new translations
         run: |
-          git add /src/i18n/org/opencastproject/adminui/languages/
-
-      - name: update language list
-        working-directory: /src/i18n/org/opencastproject/adminui/languages/
-        run: |
-          echo -n '[ "' > locales.json
-          echo -n ??-??.json | sed 's/  */", "/g' >> locales.json
-          echo '" ]' >> locales.json
+          git add src/i18n/org/opencastproject/adminui/languages/
 
       - name: upload translations
         run: |
--- a/.github/workflows/crowdin-download-translations.yml
+++ b/.github/workflows/crowdin-download-translations.yml
@@ -16,24 +16,29 @@ jobs:
         run: |
           git config --global user.email 'crowdin-bot@opencast.org'
           git config --global user.name 'Crowdin Bot'
+
       - name: prepare crowdin client
         run: |
           wget --quiet https://artifacts.crowdin.com/repo/deb/crowdin3.deb -O crowdin.deb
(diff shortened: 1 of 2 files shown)


ziegenberg (approved):
Looks good to me.

ziegenberg (changes requested):
The line
```
crowdin download --config .crowdin.yaml -b admin-ui-picard
```
probably also needs changing, and the branch name updated to `main`.

That's what the last GHA error suggest: https://github.com/opencast/opencast-admin-interface/actions/runs/9153874134/job/25163529960

lkiesow:
Fixed conflicts and branch name

.github/workflows/crowdin-download-translations.yml
```diff
@@ -16,26 +16,31 @@ jobs:
...
       - name: add new translations
         run: |
           git add /src/i18n/org/opencastproject/adminui/languages/
+
       - name: update language list
         working-directory: /src/i18n/org/opencastproject/adminui/languages/
         run: |
           echo -n '[ "' > locales.json
           echo -n ??-??.json | sed 's/  */", "/g' >> locales.json
           echo '" ]' >> locales.json
```
ziegenberg, line 40: Do we need this step? Isn't the list of languages configured here `src/i18n/languages.ts`?

ziegenberg (changes requested):

.github/workflows/crowdin-download-translations.yml
```diff
@@ -16,26 +16,31 @@ jobs:
...
       - name: download translations
         env:
           CROWDIN_TOKEN: ${{ secrets.CROWDIN_TOKEN }}
         run: |
-          crowdin download --config .crowdin.yaml -b admin-ui-picard
+          crowdin download --config .crowdin.yaml -b main
+
       - name: add new translations
         run: |
           git add /src/i18n/org/opencastproject/adminui/languages/
```
ziegenberg, line 33: Should we have those absolute file paths here? Same with the next step

gregorydlogan: The upstream project uses `git commit -a` here.  Unsure which is less safe :D

ziegenberg: I'm just not sure if an absolute file path would work or if this should be relative to the current working directory...

ziegenberg: @lkiesow, could you quickly look at this? Then, we could merge it right after.

ziegenberg (approved):

ziegenberg:
Looking good now!

Issue #395, opened 2024-05-14
Allow configuring the base URL for API requests


lkiesow:
The Opencast Editor allows us to configure the base URL and authentication for API requests:

https://github.com/opencast/opencast-editor/blob/813a1abd4861c7ca14264d34a4c7a76d0c4882f2/public/editor-settings.toml#L19-L23

This makes it very easy to build and deploy static pages for every pull requests which then get their data from develop.opencast.org. If we can do that, it would be really easy to automatically deploy pull requests which in turn would make reviewing patches much easier.

PR #416, opened 2024-05-22
Automatically Deploy Demo Page

lkiesow:
This patch deploys the latest version of the admin interface to GitHub Pages (admin-interface.opencast.org). The mock data are deployed so that they can be used for `GET` requests.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,65 @@
+name: Publish Test Page
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v4
+
+      - name: get node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: download dependencies
+        working-directory: ./app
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,65 @@
...
+        working-directory: ./app
+        env:
+          CI: false
+        run: npm run build
+
+      - name: create pages directory
+        working-directory: ./app
+        run: mkdir gh-pages
+
+      - name: include admin intterface
```
ziegenberg, line 34:
```suggestion
      - name: include admin interface
```

ziegenberg:
Can't we use stable.opencast.org instead of the mocks?

lkiesow:
I don't think we can easily do that. We can just deploy a static page while `npm start` actually runs a proxy.
We would need to build in the option to change the domain requests are sent to.
Something I would really like to see. But for now, this is a simple solution we can already use.

And I would suggest using develop.opencast.-org, not stable.opencast.org to test against the latest state of Opencast

ziegenberg:
Are you talking about #395, right?

lkiesow:
> Are you talking about #395, right?

That issue describes what I would like to see

PR #457, opened 2024-05-24
Correctly import zh-CN and zh-TW and remove language flags

ziegenberg:
This fixes the garbled use of zh-CN and zh-TW. We used the flag from the one, the description of the other and only used one, so somehow got it mixed up.

The strings 简体中文 and 繁體中文 were approved by a native speaker. We shouldn't be stepping on anybody's toes too badly ...

Apart from that we also enable importing .json files for the translations.

_EDIT:_
Also removing the language flags now


ziegenberg:
@JulianKniephoff would you approve the changes in b817e4c345db8c28df7b26c431c20a07c5eea143?

JulianKniephoff (commented):
The code changes I gladly approve.

Regarding the language stuff I have some thoughts, but obviously I'm not a native, so take this with a grain of salt. I did some research right now, but of course I'm not immersed in these cultures, and know very little about conventions and idiomatic use of these languages/terms. However, while I don't want to question any one person in particular, I **do** want to put out there that being a native does not necessarily qualify someone to answer these questions. Natives can be "wrong" about their languages, especially in a situation that's politically loaded like this one. Because remember, too, that we are dealing with **two** languages here, but see also below. Anyway:

First of all, by now we should all know that [flags aren't langauges](https://www.flagsarenotlanguages.com). Of course whether or not we want to keep the flags is a bigger discussion outside the scope of this PR. I just wanted to mention it for completeness.

What we **should** discuss here is what our translations actually represent, especially when talking about variants like this. Officially `zh_CN` is Chinese as spoken/written in the PRC, and `zh_TW` is Chinese as used in Taiwan. This is also what the flags suggest. I don't know and can't really verify whether that's what's in the translation files, but as far as I'm concerned that is what we are intending to provide here: A localization that's appropriate for mainland China and another one that's appropriate for Taiwan. And people who don't fall directly into these categories probably know/have a preference about which one of those they should pick. (Same way as we don't have a Canadian English translation, but @gregorydlogan still has an opinion about which one he wants.)

And it is true (generally, probably simplifying a lot here) that that maps to simplified vs. traditional Chinese writing, but that's only part of it. However, if I understand correctly, 简体中文 and 繁體中文 specify exactly that, simplified and traditional Chinese **writing**. So are our two Chinese translations just the same texts but written in different writing systems? I don't think so. And even if, mapping that to flags would definitely be wrong then, and also I think it's kind of weird to have that while other translations have a rather clear connection to some territory.

Anyway, these specific terms might be weird for another reason:

> The government of Taiwan officially refers to traditional Chinese characters as 正體字; 正体字; zhèngtǐzì; 'orthodox characters'.[[9]](https://en.wikipedia.org/wiki/Traditional_Chinese_characters#cite_note-9) This term is also used outside Taiwan to distinguish standard characters, including both simplified, and traditional, from other [variants and idiomatic characters](https://en.wikipedia.org/wiki/Variant_Chinese_characters).[[10]](https://en.wikipedia.org/wiki/Traditional_Chinese_characters#cite_note-10) Users of traditional characters elsewhere, as well as those using simplified characters, call traditional characters 繁體字
> —<cite>[Wikipedia](https://en.wikipedia.org/wiki/Traditional_Chinese_characters#Terminology)</cite>

So I think it's weird to use 繁體中文 as the self-description of the `zh_TW` locale. IIUC it should be 正體中文, but what I'm really getting at is that we maybe should use the names these territories use for their own language, instead of just their writing system. Which, AFAIK, would be either 中文 for both, or 汉语 and 漢語 for PRC and Taiwan respectively.

TL;DR I propose changing this:

image

to either this (note again the similarity to English):

image

or this:

image

ziegenberg:
That's what Google offers when switching languages:
Screenshot from 2024-05-28 16-17-15

For traditional Chinese, they also offer a country selection:
Screenshot from 2024-05-28 16-17-29

geichelberger:
> Using 中文 for both provides the least value and the best would be to get rid of the flags all along. We could also argue why we have a German flag for the german language when there also could be a Swiss or an Australian flag used in that case...

🚫🦘

ziegenberg:
> > Using 中文 for both provides the least value and the best would be to get rid of the flags all along. We could also argue why we have a German flag for the german language when there also could be a Swiss or an Australian flag used in that case...
> 
> 🚫🦘

Autocorrection 🤨🙄🤬

JulianKniephoff:
I'm definitely in favor of removing the flags. Regarding the wording I voiced my concerns, but I don't want to block anything with them. If Google uses these terms as well, that can of course still be wrong, but so can I, and I unfortunately don't have the resources right now to get deeper into it. 🤷‍♀️

One thing I would be interested in is what language codes Google maps these strings to. 🤔

ziegenberg:
@JulianKniephoff I rebased and removed the flags. Can you have a look once more?

tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "node_modules"
   ],
   "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
```
JulianKniephoff, line 6: This line seems a little misplaced, both in this PR and within the file. think 🤔

src/components/Header.tsx
```diff
@@ -161,12 +161,8 @@ const Header = ({
...
-						<div
-							className="lang"
-							title={t("LANGUAGE")}
-							onClick={() => setMenuLang(!displayMenuLang)}
-						>
-							<img src={currentLanguage?.flag} alt={currentLanguage?.code} />
+						<div className="lang" title={t("LANGUAGE")} onClick={() => setMenuLang(!displayMenuLang)}>
+							{currentLanguage?.long}
 						</div>
 						{/* Click on the flag icon, a dropdown menu with all available languages opens */}
```
JulianKniephoff, line 167: What flag icon? 😉

src/components/Header.tsx
```diff
@@ -161,12 +161,8 @@ const Header = ({
 				<nav className="header-nav nav-dd-container" id="nav-dd-container">
 					{/* Select language */}
 					<div className="nav-dd lang-dd" id="lang-dd" ref={containerLang}>
-						<div
-							className="lang"
-							title={t("LANGUAGE")}
-							onClick={() => setMenuLang(!displayMenuLang)}
-						>
-							<img src={currentLanguage?.flag} alt={currentLanguage?.code} />
+						<div className="lang" title={t("LANGUAGE")} onClick={() => setMenuLang(!displayMenuLang)}>
```
JulianKniephoff, line 164: I would get rid of the `title` attribute in this case, this element is now self-describing. Also, can we get rid of the `LANGUAGE` translation key then?

**OR** we use that as the text for the selector, not the selected language.

Or we use it as `alt` for a generic language selector icon, like the typical globe or the "A文" thing you see sometimes. 🤔

JulianKniephoff (changes requested):
Some nitpicks in the code.

For everyone's convenience, this is what it looks like now:

image

tsconfig.json
ziegenberg: I will remove it

src/components/Header.tsx
ziegenberg: removed

src/components/Header.tsx
ziegenberg: I would got for the third option.

ziegenberg: @JulianKniephoff which of the two icons do you prefer?

`FaGlobe from "react-icons/fa"`

Screenshot from 2024-06-04 11-47-11

`FaLanguage from "react-icons/fa"`

Screenshot from 2024-06-04 11-47-38

ziegenberg: I will leave the `title` attribute in here, as it creates a nice pop-over

Screenshot from 2024-06-04 11-52-40

lkiesow:
The selector looks fine, except the icon which is different from all other tools we have (studio, editor, tobira) and three people told me (and I agree) that they didn't identify this as a language selector. It looks like something related to internet or network or something like that.

Screenshot from 2024-06-04 16-06-33

Can we just use what we use in the other tools?
Screenshot from 2024-06-04 16-05-12


Except for that, I'm fine with the new dialog.

src/components/Header.tsx
JulianKniephoff: I think I prefer the globe in this case 🤔

ziegenberg:
> The selector looks fine, except the icon which is different from all other tools we have (studio, editor, tobira) and three people told me (and I agree) that they didn't identify this as a language selector. It looks like something related to internet or network or something like that.
> 
> 
> Can we just use what we use in the other tools? 

@lkiesow, would you be okay with the second version, like here: https://github.com/opencast/opencast-admin-interface/pull/457#discussion_r1625709840? I would stay within the same iconography, and that's currently FontAwesome5. We can change this at a later stage.

@JulianKniephoff prefers the globe.

lkiesow:
> @lkiesow, would you be okay with the second version, like here: [#457 (comment)](https://github.com/opencast/opencast-admin-interface/pull/457#discussion_r1625709840)? I would stay within the same iconography, and that's currently FontAwesome5. We can change this at a later stage.

Sorry for the late reply. Lots of other patches drawing attention ;-)
That's definitely better. But judging by your poll in Matrix, you might have already decided to go with another icon.

lkiesow (approved):

PR #458, opened 2024-05-24
Add module declarations for svg and png ressources

ziegenberg:
Removes a bunch of `ts-expect-error`:

```
// @ts-expect-error TS(2307): Cannot find module '../img/*.svg' or
// @ts-expect-error TS(2307): Cannot find module '../../img/*.png' or
```

### DIFF
--- a/src/custom.d.ts
+++ b/src/custom.d.ts
@@ -1,0 +1,2 @@
+declare module "*.svg";
+declare module "*.png";
--- a/src/i18n/languages.ts
+++ b/src/i18n/languages.ts
@@ -43,8 +43,7 @@ import { zhCN } from "date-fns/locale/zh-CN";
  * dateLocale: is needed for translation in datepicker
  *
  * !!! If a translation file of a new language was added, please insert these language here, too !!!
- *
- * */
+ */
 const languages = [
 	{
 		code: "en-US",


ziegenberg:
@Arnei, would you consider this a valid approach?

Arnei:
Absolutely no expert on this, but if our goal is to get rid of ` @ts-expect-error` messages, I would consider this a valid solution.

Apart from getting typescript to shut up, this does not add anything of value I believe. We're basically declaring pngs as `any` and svgs as `string` which is ... technically correct I suppose, but not actually all that useful? If we want good typing we should probably take a look at libraries like [svgr](https://react-svgr.com/). But that again begs the question if really need all that jazz. Certainly not if just want typescript to stop complaining.

And you can probably shorten the module declarations down to just `declare module "*.png";`? Not sure.

geichelberger:
While switching to Vite, I had to remove the `ts-expect-error` annotation. Else, `tsc` was complaining.

https://github.com/opencast/opencast-admin-interface/pull/471

ziegenberg:
I rebased and fixed the merge conflicts.

@Arnei I also shortened the module declarations as suggested.

Arnei:
Actually, do we still need this? Or did this become unnecessary with the introduction of vite and stuff?

ziegenberg:
@Arnei We still have messages like `Cannot find module '../../img/*.png' or` and they are gone with this patch.

geichelberger:
svgs should be handled by [svgr](https://react-svgr.com/)

ziegenberg:
So, then let's close this.

PR #460, opened 2024-05-24
Remove superflous SCSS for old video editor

ziegenberg:
The new standalone editor replaced the integrated editor, and these styles are no longer needed. Only the `.video-player` class was used again for media asset video preview and required to stay.


Arnei (approved):

ziegenberg:
@Arnei I rebased and amended my commit slightly.  Only the `.video-player` class was used again for the media asset video preview and required to stay. Can you have another look at it?

Arnei (approved):
Clicked on everything I could think of, can confirm that the media asset video preview is working.

PR #461, opened 2024-05-24
Replace deprecated i18next-xhr-backend with i18next-http-backend

ziegenberg:
The new [i18next-http-backend](https://www.npmjs.com/package/i18next-http-backend) is based on the deprecated [i18next-xhr-backend](https://github.com/i18next/i18next-xhr-backend) and can mostly be used as a drop-in replacement.


ziegenberg:
Rebased onto current main

Issue #429, opened 2024-05-23
Don't vendor fontawesome


JulianKniephoff:
We have a (partial?) copy of fontawesome in `style/base/fontawesome`. That's not good. Additionally, we depend on the `font-awesome` package and use that **as well**!

ziegenberg:
How about only using `react-icons`? I would be working on such a PR at the moment.

JulianKniephoff:
Sure, I'm not married to Fontawesome. If it helps us get rid of this Frankenstein dependency I'm all for it! :D

Arnei:
> How about only using react-icons?

That's something I would bring up at the technical meeting or somewhere. I remember people having *opinions* on icons.

JulianKniephoff:
That's fair, but IIRC that was mostly in the context of the recently redesigned apps. Increasing consistency with those doesn't hurt, though, of course.

Arnei:
Resolved by  #1356.

PR #462, opened 2024-05-24
remove partial? vendored frankstein-dependency Fontawesome and replace with react-icons

ziegenberg:
I thought about doing this in a more gentle way by replacing it piece by piece. But ripping it out all at once is the easiest option.

This `font-awesome` package with version 4.7.0, was published 8 years ago. We can do better than this. 'react-icons' also provides FontAwesome icons, so let's use that.

Related to #429


Arnei:
#1356 has superceded this, so I think it is ok to close this. Do reopen in case you don´t agree.

Issue #465, opened 2024-05-25
Video upload


Melanie-Koller:
The part in brackets next to "Multipart: Presenter Part 1" should be removed.
In addition, the description should be in German everywhere.

![Bildschirmfoto 2024-05-23 um 16 05 00](https://github.com/opencast/opencast-admin-interface/assets/57887263/c783512a-cafd-4c87-80ba-96644512ae51)

It should look like this:

![Bildschirmfoto 2024-05-23 um 15 56 00](https://github.com/opencast/opencast-admin-interface/assets/57887263/ee71e796-c80c-4a3a-8142-fc9d88edbf0d)

ziegenberg:
Fixed by #467

PR #467, opened 2024-05-27
Remove asset type + flavor from source upload

Arnei:
Removes superfluous information in the source upload and asset upload of the "Create Event" dialog. The information in question concerns the asset type (e.g. track) and flavor.

Reasoning:
- The current formatting of the info obfsucates the dialog, making it hard to parse at a glance.
- The info is not needed. Technical users will either know this by heart or can easily look it up in the rare case they need it, and for less technical users the title and description for each asset should be more than sufficient. (And in case they are not, title and description should be improved).

Looks like:
Bildschirmfoto vom 2024-05-27 09-56-58

Helps with #465.


ziegenberg (approved):

ziegenberg:
Works as advertised.

PR #471, opened 2024-05-27
Replace react-script with vite

geichelberger:
Vite was chosen because of https://github.com/opencast/opencast-editor/pull/1265

> Vite was chosen because:
>
>    Straightforward migration
>    Relatively similar set of opinions as CRA
>    Optimized for single page applications like ours
>    Popular

I didn't reimplement the dev mode with the mocks.

PR #472, opened 2024-05-28
Automatically deploy pull requests

lkiesow:
This patch automatically deploys pull requests with static mock data to https://test.admin-interface.opencast.org similar to how the test deployment of the Opencast Editor works.

Unfortunately, we cannot set a path prefix for mock data which means that all deployments share the lattest set of mock data. But since they shouldn't change very often, that's probably okay for this to get started.


Arnei (approved):
The workflow reads reasonably, mostly because it reads very similar to the respective editor workflow, so we should be good.

Seeing as how this would be quite difficult to test without access to the required github secrets, how about we go ahead and merge this? We can still hotfix it if anything breaks.

Also this is not necessarily a change request, but why do we not do a opencast developer/committer membership check like in the editor?

lkiesow:
> Also this is not necessarily a change request, but why do we not do a opencast developer/committer membership check like in the editor?

Because this cases the code from a pull request to be checked out and executedin an environment which has access to secrets like the private deployment key. A malicious actor can steal that. A malicious actor can aalso use this to publish any code, binary, … under an opencast.org domain. Again, not great.

GitHub has a limited built-in protection in that this will only run automatically for users which have an accepted contribution merged into the default branch. But that also means a contributtion from 10 years ago with no way for us to remove that user even if we know that their account has been compromised.

lkiesow:
> Seeing as how this would be quite difficult to test without access to the required github secrets, how about we go ahead and merge this? We can still hotfix it if anything breaks.

+1 from me. I justt verified tthat #471 did not break this. We already discussed this PR on Tuesday. If you waant to merge tthis, please go ahead.

PR #483, opened 2024-05-28
Test deployments without mocks

geichelberger:
This change allows to configure the Axios client to use a different base
URL and set the authentication header.

NODE_ENV=development
VITE_TEST_SERVER_URL=https://develop.opencast.org/
VITE_TEST_SERVER_AUTH="admin:opencast"

Fixes https://github.com/opencast/opencast-admin-interface/issues/395


geichelberger:
There will be more documentation when it's ready for review. I will refactor it after https://github.com/opencast/opencast-admin-interface/pull/472 is merged,

ziegenberg:
Can the mock data be re-used for Unit and integration tests? So, should we throw it away with this PR or keep it for later use?

geichelberger:
The data is not gone. It's still in the commit history.

geichelberger:
To have data that is not used and, therefore, not maintained makes no sense. And with the argumentation that we may need it in the future, we should not remove unused stuff at all.

geichelberger:
Should we document the environment variables? They should only be used in the workflows, and the other workflow-specific variables are not documented either.

README.md
```diff
@@ -50,6 +50,12 @@ PROXY_TARGET=http://localhost:8080 PROXY_AUTH=jdoe:aligator3 npm start
 Similarly, if you want to change the port the development server itself runs at,
 you can specify an alternative port in the `PORT` environment variable.
 
+If you aim to test against a remote server without using a proxy, you have the option to configure the target server with the `VITE_TEST_SERVER_URL`, and the `VITE_TEST_SERVER_AUTH` environment variables while using the node development mode:
+
+```sh
+NODE_ENV=development VITE_TEST_SERVER_URL="https://develop.opencast.org" VITE_TEST_SERVER_AUTH="admin:opencast" npm start
```
gregorydlogan, line 56: I'm not sure how common this will be, in testing against localhost I could not get this to work until I put a proxy with a wide open CORS config in between.  So `VITE_TEST_SERVER_URL="http://localhost:8080"` did *not* work, but `VITE_TEST_SERVER_URL="http://localhost" <- now with proxy!` did.  Would it be worth explaining that here, or am I sufficiently out of touch with frontend dev that this is something that's obvious?

geichelberger: I think CORS is Daily Business nowadays, and the Vite proxy would be a better fit for local scenarios.

src/index.tsx
```diff
@@ -31,6 +32,11 @@ const theme = createTheme({
 	}
 })
 
+if (import.meta.env.DEV && import.meta.env.VITE_TEST_SERVER_URL) {
+	axios.defaults.baseURL = import.meta.env.VITE_TEST_SERVER_URL || ""
+	axios.defaults.headers.common['Authorization'] = import.meta.env.VITE_TEST_SERVER_AUTH && ("Basic " + window.btoa(import.meta.env.VITE_TEST_SERVER_AUTH));
```
Arnei, line 37: Checkstyle complaint

Arnei (changes requested):
Wow, this has been utterly forgotten >.>. Looks good to me though, just one checkstyle complaint.

Arnei (approved):

PR #494, opened 2024-05-30
Add .dockerignore

lkiesow:
This patch adds a `.dockerignore` file to exclude unnecessary files like the git history from the container images.


Arnei (approved):

ziegenberg:
As it has been approved, merging it now

PR #430, opened 2024-05-23
Fix build warnings

JulianKniephoff:
Fixes #350.

See individual commit messages for more details.

PR #496, opened 2024-05-31
Set CI=true in all workflows

lkiesow:
As all the warnings on the build are now fixed (thx to #430), we no longer need to set `CI=false` to actively suppress warnings to be picked up by our CI tests. This should help with us not introducing new warnings.


ziegenberg (approved):
Looking good to me!

Issue #510, opened 2024-06-01
Fix dependency updattes


lkiesow:
The following updates seem to have problems:

- [x] https://github.com/opencast/opencast-admin-interface/pull/443
- [x] https://github.com/opencast/opencast-admin-interface/pull/498
- [x] https://github.com/opencast/opencast-admin-interface/pull/499
- [x] https://github.com/opencast/opencast-admin-interface/pull/504

Take a look. Fix the problem.

ziegenberg:
#504 was superseded by #514

ziegenberg:
#498 was superseded by #512

ziegenberg:
#499 was superseded by #513

ziegenberg:
#442 was superseded by #511

Issue #455, opened 2024-05-24
Date Filter


Melanie-Koller:
The date filter is working. But if i choose the date filter in the list, there is only 1 event.

![Bildschirmfoto 2024-05-24 um 08 20 45](https://github.com/opencast/opencast-admin-interface/assets/57887263/0546a998-4217-45b9-9fd5-727734591f2e)

![Bildschirmfoto 2024-05-24 um 08 21 16](https://github.com/opencast/opencast-admin-interface/assets/57887263/de454909-65a0-4dfe-8b44-a6ed97e08158)

Melanie-Koller:
If i choose the date filter with the calendar and select only one day, it is not working. It only works, if i write the date.

![Bildschirmfoto 2024-05-25 um 08 19 37](https://github.com/opencast/opencast-admin-interface/assets/57887263/6bb2e9cc-f074-4a14-9339-9c520ee08cee)

Arnei:
This has since been fixed by the linked PRs I believe.

PR #513, opened 2024-06-03
Update dependency muixdatepickers

Arnei:
Update dependency  @mui/x-date-pickers. Depends on #512.

Helps with #510.


ziegenberg (approved):
Datepicker works, but filtering by dates is as borked as before (see #455). Still, I think this can be merged though.

PR #514, opened 2024-06-03
Update date-fns to 3.6.0

Arnei:
Update dependency date-nfs to the newest version.
The new major version changed some import paths which had to be fixed.

Helps with #510.


ziegenberg:
Works as advertised.

PR #515, opened 2024-06-03
Some vite nitpicks

JulianKniephoff:
This does several things:

* It swaps out `@vitejs/plugin-react` for the `-swc` variant to speed up the build even further.
* Remove a deprecation warning. (I thought my job was done with #430 but alas ... 🥲)
* Clean up some other small things that were overlooked in #471.

See individual commits and their messages for more details.


ziegenberg:
It still builds, it runs and the changes make sense. (I had a branch with the same fixes on my machine)

ziegenberg (approved):

PR #546, opened 2024-06-04
External links in new pages

### LINKED ISSUE #525: [UX]: Make external links open in new page
All links that point to somewhere other than admin-ui are targeting the same page, therefore,
it becomes really annoying sometimes if you want to check something for example in REST API Docs quick and come back (you need to go back at least 2 times to get to admin-ui again!)

(PR Will follow!)

ferishili:
This PR fixes #525
There are a few other places like feedback and feeds that are not included here, because they were not provided by default settings.


Arnei (approved):
I tend to open these as a new tab anyway, so this makes sense to me!

Issue #486, opened 2024-05-28
Padding on buttons


lkiesow:
Comparing the old and new UI, the padding of buttons looks much better in the old one. We have a lot of free space and should use it. There is no need to make the button as small as possible.

![Screenshot from 2024-05-28 22-50-11](https://github.com/opencast/opencast-admin-interface/assets/1008395/56e9f061-591c-469c-b6ff-7aaff0af5f4b)

marwyg:
#554 makes the buttons look like this now:

![grafik](https://github.com/opencast/opencast-admin-interface/assets/5593139/d54d2542-2980-4fa0-9798-52990f325780)

PR #554, opened 2024-06-04
Fix dialog button sizes and margin

marwyg:
Fixes https://github.com/opencast/opencast-admin-interface/issues/486

Sets the buttons min-width to 100px and adjusts the spacing to the edges, so they line up with the elements above.


ziegenberg (approved):
Looking good now.

ziegenberg:
I removed all the labels, except for 'visual-clarity' as we should stick the best matching one the issue.

Issue #536, opened 2024-06-04
Checkboxes on Events tab are not labelled


luniki:
Form elements must have labels. The checkboxes on the Events tab (one per row and the check-all checkbox at the top) do not have them. This could be an implicit or explicit `<label>`, an `aria-label` or `title` attribute.

This has critical impact on blind/low vision people.

PR #558, opened 2024-06-04
Add labels to checkboxes on Events tab.

luniki:
Fixes https://github.com/opencast/opencast-admin-interface/issues/536

### DIFF
--- a/src/components/shared/Table.tsx
+++ b/src/components/shared/Table.tsx
@@ -210,6 +210,7 @@ const Table = ({
 								<input
 									type="checkbox"
 									onChange={(e) => onChangeAllSelected(e)}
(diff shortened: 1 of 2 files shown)


Arnei (approved):
Does what it says.

Since atm we have zero `aria-label`s in this project, do you have any recommendations on a unified approach towards adding them where they belong (and avoid adding them where they don't belong or are not needed)?

luniki:
> Does what it says.
> 
> Since atm we have zero `aria-label`s in this project, do you have any recommendations on a unified approach towards adding them where they belong (and avoid adding them where they don't belong or are not needed)?

I cant' cite an exhaustive list. There is of course an easy rule: All form elements have to be labelled in any way.

Issue #447, opened 2024-05-24
Admin UI should look like the last login 


Melanie-Koller:
Every time i login there is a standard view with 10 videos per page and standard sorting. I like to have the same view like the last login.

PR #560, opened 2024-06-04
Persist table pagination settings

### LINKED ISSUE #544: Max entries per page resets on language switch
Changing the maximum entries per page\
- on the events page (`/#/events/events`) or
- on the series page (`/#/events/series`) \

to something else than `10` and your are going to switch language (i.e. to `100`) the max. entires per page are reset to `10`.

Expected to keep the value when switching languages in the UI.

Arnei:
Persists changes to table pagination information, most notably the setting for how many items per page are displayed.

Since this information is stored and applied to ALL tables, setting the limit from 10 to 20 for the events table will also set it to 20 for the series table and all other tables. If we say that is no good, and we want the settings to be per table instead, I'll have to create a different, more extensive PR.

Fixes #544, #447.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -37,6 +37,7 @@ import autoMergeLevel2 from "redux-persist/lib/stateReconciler/autoMergeLevel2";
 // Configuration for persisting states in store
 const eventsPersistConfig = { key: "events", storage, whitelist: ["columns"] }
 const seriesPersistConfig = { key: "series", storage, whitelist: ["columns"] }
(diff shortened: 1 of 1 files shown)


reiferschris (approved):
I can confirm that at least https://github.com/opencast/opencast-admin-interface/issues/544 is fixed by this. I don't know how to properly test logout/login with the container test setup but after "logging out" and the reload that happens after I still have the pagination settings saved

lkiesow:
I'm merging this for you. Thanks for the review

PR #563, opened 2024-06-04
Fix misplaced themes details menu navigation buttons

### LINKED ISSUE #551: Misplaced themes details menu navigation buttons
In Themes details, the navigation buttons are not in place!

In local
Screenshot 2024-06-04 at 12 03 57

on https://develop.opencast.org/
Screenshot 2024-06-04 at 12 06 02

dennis531:
image

Closes #551

### DIFF
--- a/src/styles/components/modals/_nav.scss
+++ b/src/styles/components/modals/_nav.scss
@@ -31,7 +31,7 @@
 
     > button {
         $width: 100px;
-        width: $width;
+        min-width: $width;
         display: inline-block;
         text-align: center;
         position: relative;


ferishili:
@dennis531 would that be also good to nowrap the text in the nav items? In this way it looks even better. What do you think?

dennis531:
> @dennis531 would that be also good to nowrap the text in the nav items? In this way it looks even better. What do you think?


`white-space: nowrap;` causes long texts to become completely invisible. I would suggest min-width as an alternative.

reiferschris:
There is still a continuous line below the "arrow" that points to the current tab that doesn't exist on the old admin-ui.

dennis531:
Solution with min-width:
image

ziegenberg:
I like the `min-width` solution 👍🏻

Arnei:
Looks like people are happy with this, merging.

Issue #542, opened 2024-06-04
Admin Interface should use ARIA landmarks


luniki:
It is a good practice benefitting people using a screenreader to ensure that all content is contained within a navigational region. That is using `header`, `nav`, `main` and `footer` elements and additionally using the ARIA counterparts `role="navigation"` and so on.

The current Admin Interface already employs a `header` and `footer` but should move the `.action-nav-bar` and the `.main-view` into a `nav` and `main` element. Furthermore using the matching ARIA roles makes this more robust for some screen readers.

PR #564, opened 2024-06-04
Move all HTML content into ARIA landmarks.

luniki:
Fixes https://github.com/opencast/opencast-admin-interface/issues/542

### DIFF
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -1,0 +1,14 @@
+import React from "react";
+
+/**
+ * Component that renders the nav bar
+ */
+const NavBar: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+	return (
+		<section className="action-nav-bar" role="navigation">
+			{children}
+		</section>
+	);
+};
+
+export default NavBar;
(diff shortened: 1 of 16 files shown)


src/components/NavBar.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+/**
+ * Component that renders the nav bar
+ */
+const NavBar: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+	return (
+		<section className="action-nav-bar" role="navigation">
```
Arnei, line 8: Does it actually make sense to assign `role="navigation"` to this and call this `NavBar`? Not necessarily your fault, the css className would be misleading. And while this element contains navigation links like "Events" and "Series", and a button that opens the actual main menu, it also contains buttons for sorting the table and opening dialogs. In that sense, it is more like a second, multi-purpose header. So `role="navigation"` might actually be confusing?

luniki: What would you suggest?

Arnei: Not sure. This was kind of more of a comprehension question :'D.

Maybe I'm approaching this wrong from the wrong perspective anyway. Maybe it is perfectly normal and expected for users of screenreaders to find a section with role "navigation", which then contains `nav` tags. And maybe it's just weird to me as someone who has barely ever used screenreaders. Ultimately I don't know, and wanted an explanation so that I may know in the future.

Issue #527, opened 2024-06-04
Comments styling issues


stefanosgeo:
The submit button in the comments goes out of the modal body 

![Screenshot from 2024-06-04 11-02-55](https://github.com/opencast/opencast-admin-interface/assets/3663204/95c4974f-8a9f-4c16-921e-bbf99fdc6065)

Also the resolve checkbox in an existing comment is, IMO, too big in compare with the "resolve" text on the right 

![Screenshot from 2024-06-04 11-03-23](https://github.com/opencast/opencast-admin-interface/assets/3663204/543023fa-5d04-4809-af3f-c1be98c101a3)

stefanosgeo:
I am not sure if this is the same css rule/class,  but the same happens if you try to change the location to a scheduled event in the scheduling tab.

![Screenshot from 2024-06-04 11-12-42](https://github.com/opencast/opencast-admin-interface/assets/3663204/d8a0bd60-f8d2-49db-aa3f-38a5949defb8)

marwyg:
Related to https://github.com/opencast/opencast-admin-interface/issues/540
I created a PR for the first one: https://github.com/opencast/opencast-admin-interface/pull/565
And I am seeing now, that the CSS I removed in my PR messed up the 2nd page :P I will add more commits to https://github.com/opencast/opencast-admin-interface/pull/565 so it fixes the 2 first views you posted.

marwyg:
Fix for the scheduling tab is out: https://github.com/opencast/opencast-admin-interface/pull/572

Issue #526, opened 2024-06-04
Click on an entry in a cell makes OC crash


kaspot:
No matter on what “link” you click in the table in the event overview: Peter Planner, Mock Series, etc. you get a blank screen and are not able to recover. You have to clear the cookies, to be able to reload the interface. 

![IMG_1478](https://github.com/opencast/opencast-admin-interface/assets/1519714/42ae8631-c006-4de4-9e50-42e8ca505d98)
![attatchment AbhcWK](https://github.com/opencast/opencast-admin-interface/assets/1519714/f1fafe40-718a-4fbb-a1dd-c177c6e44371)

marwyg:
Not that anyone gets confused: #565 has nothing to do with this Issue here. I accidentally mentioned the wrong issue.

dennis531:
I cannot reproduce this error. When I click on a link, the corresponding filter will be set. The edit table view works too. Could you check if the error still exists?

PR #565, opened 2024-06-04
Fix button misplacmenet in the comment section and cleanup css

### LINKED ISSUE #540: Misplaced submit button in Event Details -> Comments section
The submit button is off!
Under Event -> Details -> Comments

Screenshot 2024-06-04 at 11 33 40

marwyg:
Fixes #540 

This PR adds some CSS flex to rearrange the button and the selectbox. Furthermore it cleans up some css that seem to have no effect and makes the selectbox a little bit larger (because why not, there is plenty of room).

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/EventDetailsCommentsTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/EventDetailsCommentsTab.tsx
@@ -309,7 +309,42 @@ const EventDetailsCommentsTab = ({
 										}
 									></textarea>
 
-									{/* submit button for comment reply (only active, if text has been written) */}
+                  {/* 'resolved' checkbox */}
+									{hasAccess(
+										"ROLE_UI_EVENTS_DETAILS_COMMENTS_RESOLVE",
+										user
+									) && (
+										<>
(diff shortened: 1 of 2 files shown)


jduehring:
Looks good to me and works as expected.

marwyg:
My removed CSS seems to break the view, when you click on "reply". See #526 as reference. I will make another commit for this PR here to enhance styling of the "reply" view.

marwyg:
I kept the big checkbox against this comment: https://github.com/opencast/opencast-admin-interface/issues/527#issue-2332984970 because of accessibility reasons. What do you think?

Nonetheless, the reply view looks like this now:
grafik

marwyg:
It's ready for review

marwyg:
I thought about this. But I am not sure if I might break something and I found out, that it's possible to change order with CSS ;) But if you prefer the other way I try it out.

Arnei:
I would avoid ordering with CSS unless necessary in order to keep with convention and keep it simple.

marwyg:
I reordered the elements in the TSX now.

Arnei (approved):

PR #571, opened 2024-06-04
Fix notification when changing series theme

### LINKED ISSUE #384: After assigning a theme to a series no message is presented
After assigning a theme to a series no message is presented and it remains unclear if the save button worked:

image

Should be consistent to e.g. acl where a message is presented after saving:

image

jduehring:
This PR fixes #384 . Like in the old admin UI, the following notification will be displayed after hitting the save button: 

image


ziegenberg (approved):
This is fixed now.

Issue #408, opened 2024-05-21
Rendering components directly into document.body is discouraged


ziegenberg:
We get the following browser console warning:

```
render(): Rendering components directly into document.body is discouraged, since its children are often manipulated by third-party scripts and browser extensions. This may lead to subtle reconciliation issues. Try rendering into a container element created for your app.


tsx index.tsx:33
```

This should be easy to fix.

PR #573, opened 2024-06-04
Fix render warning

geichelberger:
This fixes following warning:

```
Warning: render(): Rendering components directly into document.body is discouraged, since its children are often manipulated by third-party scripts and browser extensions. This may lead to subtle reconciliation issues. Try rendering into a container element created for your app.
```


ziegenberg:
Fixes #408

ziegenberg (approved):
I confirm, the warning is gone

PR #575, opened 2024-06-04
Sort specific lists in dropdown menu alphabetically

### LINKED ISSUE #567: Sort list in drop down fields
The list in drop down fields isn't sorted. Examples are the language or series selection when creating a new event. I would expect the list in the drop down to be sorted in ascending order.

grafik

owi92:
This adds a function to check whether the options for a dropdown menu are custom ordered. If they are not, the resulting list is ordered alphabetically. Otherwise the ordering is left as is (this would be the case for the "license" field for example).

Closes #567

### DIFF
--- a/src/utils/dropDownUtils.ts
+++ b/src/utils/dropDownUtils.ts
@@ -1,5 +1,6 @@
 import { TFunction } from "i18next";
 import { DropDownType } from './../components/shared/DropDown';
+import { isJson } from "./utils";
 /*
  * this file contains functions, which are needed for the searchable drop-down selections
  */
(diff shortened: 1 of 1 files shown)


snoesberger:
I can confirm that the sorting is working. I tested it for the event language and series drop down field in the modify pop-up. The licence field isn't sorted, because this is handled by the opencast config (see comment https://github.com/opencast/opencast-admin-interface/issues/567#issuecomment-2147413835)

src/utils/dropDownUtils.ts
```diff
@@ -46,6 +47,14 @@ export const formatDropDownOptions = (
...
 	t: TFunction
 ) => {
+	/**
+	 * This is used to determine whether any entry of the passed `unformattedOptions`
+	 * contains an `order` field, indicating that a custom ordering for that list
+	 * exists and the list therefore should not be ordered alphabetically.
+	 */
+	const hasCustomOrder = unformattedOptions.some((item: any) => 
+		isJson(item.name) && JSON.parse(item.name).order !== undefined);
+
```
lkiesow, line 57: If I understand this correctly, this will only indicate if any of the items has a field `order`.

src/utils/dropDownUtils.ts
```diff
@@ -111,5 +120,7 @@ export const formatDropDownOptions = (
 		}
 	}
 
-	return formattedOptions;
+	return hasCustomOrder
+		? formattedOptions
```
lkiesow, line 124: If any item has a field order, this will not use the order field for sorting, but stop sorting this at all. Or am I mistaken.

lkiesow (changes requested):

src/utils/dropDownUtils.ts
owi92: Yes, that is correct. I suppose this could check instead if all items have that field - or are you saying that checking for that simply isn't enough to imply custom ordering? If that is the case, I'm honestly not sure how else to do it.

src/utils/dropDownUtils.ts
owi92: Right, good point. I didn't touch this because I somehow thought they would already be ordered as configured, but now that you mention it I checked, and that does not seem to be the case.

lkiesow (approved):

PR #579, opened 2024-06-04
Update Deployment Comment

lkiesow:
Instead of writing a new comment every time the static test site gets deployed, this patch will cause the first comment to be updated similar to the container image deployment.

This means that if you open a pull requests, the first two comments should usually hold all information about the latest deployment.


ziegenberg:
I don't know if this is working every time. But let's merge it anyway and see if it does anything...

ziegenberg (approved):

PR #581, opened 2024-06-04
Simplify code for Notifications.tsx

Arnei:
Since we are now using redux toolkit this can be simpler and use typescript.


ziegenberg (approved):
I tested several cases where notifications are displayed and they still work. So I think this is fine to merge.

PR #582, opened 2024-06-04
Update react and get rid of mui styles

geichelberger:
This updates React 18 from 17 and removes mui styles because it's deprecated and does not work with React 18.

I don't know if the way I replaced makeStyle is the best solution.


src/utils/wizardUtils.ts
```diff
@@ -1,24 +1,23 @@
...
 // Base style for Stepper component
-export const useStepperStyle = makeStyles({
+export const stepperStyle = {
 	root: {
 		background: "#eeeff0",
 		height: "100px",
 		padding: "24px",
 	},
-});
+} as const;
```
Arnei, line 10: I don't think `as const` is actually necessary.

Arnei (changes requested):
Finally, we can upgrade to React 18, thanks for that. Minor nitpick below.

Arnei (approved):

Issue #562, opened 2024-06-04
changing the size of the embed code should remove the text after successfully copying to the clipboard


ziegenberg:
While reviewing #537, I discovered that changing the embed size after clicking the "Copy to clipboard" button does not remove the text "Embedding code has been copied to the clipboard".

**What currently happens**

1. Click the chain icon to open the embed code modal
2. Select a size
3. Click "Copy to clipboard"
4. The text "Embedding code has been copied to the clipboard: SIZE" appears above the button
5. Select another size
6. The text changes and only the size is updated

**What should happen**

1. Click the chain icon to open the embed code modal
2. Select a size
3. Click "Copy to clipboard"
4. The text "Embedding code has been copied to the clipboard: SIZE" appears above the button
5. Select another size
6. The text changes should disappear to encourage the user to click the button again

PR #586, opened 2024-06-04
Reset copy success message

geichelberger:
This resets the success message after changing the size of the embed code.

Fixes https://github.com/opencast/opencast-admin-interface/issues/562


ziegenberg (changes requested):
Nice! That works :)

ziegenberg (approved):

ziegenberg:
Should we also clear the clipboard at the same instant? This prevents the user from accidentally pasting the wrong embed code with the wrong size.

PR #587, opened 2024-06-04
Submitter undefined

### LINKED ISSUE #359: Submitter undefined
The submitter of a workflow can never be undefined, but the new interface states exactly that:

Screenshot from 2024-05-07 18-43-09

This is likely due to the submitter being the admin user which has no email address?

lkiesow:
The submitter of a workflow can never be undefined, but the new interface states exactly that. The reason for that is that the interface tries to access non-existing data. This patch fixes the problem.

This fixes #359


gregorydlogan (approved):
Works for me

PR #588, opened 2024-06-04
Enforce minimum column width

### LINKED ISSUE #450: Header looks strange
The header looks strange, if i have videos on the list with a long title.

Bildschirmfoto 2024-05-24 um 07 05 22

gregorydlogan:
This PR enforces a minimum column width to prevent the sorting arrow from dropping down a line.

Fixes #450

### DIFF
--- a/src/styles/components/_tables.scss
+++ b/src/styles/components/_tables.scss
@@ -52,6 +52,7 @@
 
     th {
         &.sortable {
+            min-width: 80px;
             cursor: pointer;
             &:hover {
                 background: linear-gradient(to bottom, $off-white, darken(#f1f3f5, 3%));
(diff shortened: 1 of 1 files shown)


ziegenberg:
Setting the min-width to 80px does not seem to be enough:

image

The location still drops

ziegenberg (changes requested):
Is there a way to make this safe with even longer translations?

gregorydlogan:
I came up with the 80 value by expanding things until (i thought...) nothing dropped. Honestly not sure how to make the minimum auto set - I'm guessing we need to check all the languages, find the longest, and then add a bit extra?

gregorydlogan:
Bringing that value up to 150(!) gets you no more dropping sorts, but in the case where the window width is smaller than `rows * 150 + table/window dressing`  you get

image

The minimum window size that triggers this is 1490px, which is... not that small.  Unclear what the correct course of action is here.  It looks goofy to have the page do this, but the dropping sorters also look silly.

Issue #534, opened 2024-06-04
Stats (Today, Scheduled, Recording, …) do not meet minimum color contrast ratio


luniki:
This is the same problem as in #532 

The text in the Stats container does not meet the minimum color contrast ratio.

Currently the ratio is about 2:1 but should be at least 4.5:1.

This has a serious impact on people with low vision.

PR #589, opened 2024-06-05
Enhance color contrast ratio of stats.

luniki:
Fixes https://github.com/opencast/opencast-admin-interface/issues/534


Arnei (approved):
I like being able to read things.

Issue #532, opened 2024-06-04
Non-active tabs ("Events"/"Series") does not meet minimum color contrast ratio


luniki:
There is the tab headers "Events" and "Series". The non-active tab header does not meet the minimum color contrast ratio.

Currently the ratio is about 2:1 but should be at least 4.5:1.

This has a serious impact on people with low vision.

PR #590, opened 2024-06-05
Enhance color contrast ratio of tab headers.

luniki:
Fixes https://github.com/opencast/opencast-admin-interface/issues/532


Arnei (approved):
I like being able to read things.

Issue #548, opened 2024-06-04
Not yet started WO show "Invalid date" in their metadata 


Shinymonk:
![NotYetStartedWOShowInvalidDate](https://github.com/opencast/opencast-admin-interface/assets/4223485/a1585779-7212-4033-91d7-331ecd821755)

Issue #464, opened 2024-05-25
Edit the metadata of multiple events


Melanie-Koller:
If i like to change the date of multiple events with the same date, there is written "invalid date". And it is only possible to change the date with the calendar. I can't write the date.

![Bildschirmfoto 2024-05-25 um 08 33 39](https://github.com/opencast/opencast-admin-interface/assets/57887263/73d5039f-fc24-41e6-b5f5-b14e5a3407b4)

Arnei:
#606 should make it so that "Invalid date" is not wrongly shown anymore.

On the most current version of the admin ui I am able to change the date both by calendar and by typing, so maybe that part of this issue was already fixed incidentally.

PR #606, opened 2024-06-05
Avoid rendering "Invalid date"

Arnei:
We've got multiple issues with people complaining about "Invalid date". Often enough, that string is simply rendered because there (rightly) is no date available. Expected behaviour is to render the empty string, so this does that in a bunch of places.


reiferschris:
https://github.com/opencast/opencast-admin-interface/issues/548
I can confirm that this issue is fixed by this PR

 #464 Also seems to be solved by this, selecting multiple events with different dates/times doesn't show invalid date anymore

Issue #428, opened 2024-05-23
Closing of modal with ESC does not work


ziegenberg:
Open any modal and try to close it with `ESC` right away.

**Expected outcome**: The modal is closed.

**Actual outcome**: Nothing happens.

PR #619, opened 2024-06-05
Add esc hotkey to close modals

owi92:
This adds `esc` to the hotkeys config, and uses it to close all modals.
At least I think I got all modals, but I wouldn't be completely surprised if I missed a few anyway.
This is a lot of repeated code and it would be nice to have one or two modal components that can be reused, but that is a larger refactoring task for another time.

Closes https://github.com/opencast/opencast-admin-interface/issues/428

### DIFF
--- a/src/components/events/partials/modals/DeleteEventsModal.tsx
+++ b/src/components/events/partials/modals/DeleteEventsModal.tsx
@@ -5,6 +5,8 @@ import { getSelectedRows } from "../../../../selectors/tableSelectors";
 import { connect } from "react-redux";
 import { useAppDispatch } from "../../../../store";
 import { deleteMultipleEvent } from "../../../../slices/eventSlice";
+import { useHotkeys } from "react-hotkeys-hook";
+import { availableHotkeys } from "../../../../configs/hotkeysConfig";
 
 /**
  * This component manages the delete bulk action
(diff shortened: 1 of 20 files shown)


ziegenberg:
You missed the hotkey modal :D (open it via `h`)

ziegenberg:
It does not work on the location tab (`#/recordings/recordings`)

ziegenberg:
It does not work on the users tab:
- for user (`#/users/users`)
- for groups (`#/users/groups`)
- for acls (`#/users/acl`)

owi92:
So many modals 😵‍💫
Hope we got them all now. Thanks for checking!

ziegenberg:
And it does not work on the configuration tab (`#/configuration/themes`)

owi92:
> And it does not work on the configuration tab

That is because `esc` doesn't unfocus the text input there. Once that is manually unfocused, it works. But I'll see if I can fix this here as well.

ziegenberg:
> Edit: the same is true for users, groups and acl btw. `esc` will only close these modals when nothing else is focused, but the input fields aren't loosing focus with `esc`.

Would you like to fix this in this PR or open another one?

owi92:
> Would you like to fix this in this PR or open another one?

I would prefer to do this in another PR, as I don't think I'll get to it today.

PR #621, opened 2024-06-05
Added option so unselect a theme in series view

### LINKED ISSUE #383: Cant unassign theme from series
It is not possible to unassign a theme from a series. In the browsers insepect menu you can see that no request is fired after selecting the "no option" theme and hitting the save button:

image

jduehring:
This PR fixes #383 . 

You can now select "No theme" option and click the save button. In doing so a delete request will be sent to remove the theme from the series and the respective notification is shown. 

image


Arnei (approved):
Works and makes sense to me.

Issue #577, opened 2024-06-04
Filter and pagination do not work well together


lkiesow:
- Open admin interface with > 20 events
- Select to display 10 events per page
- Go to page 3
- Search for existing event

Opencast should show one page with one event.

Opencast shows no events, claims to be on page 1, but actually still seems to use the offset from page 3.

Changing the filter, we should always just jump to page 1.

ferishili:
See #211
And specially: https://github.com/opencast/opencast-admin-interface/issues/211#issuecomment-1788798999

Issue #528, opened 2024-06-04
Date filter doesn't have the same look and feel as other fields/filters


snoesberger:
The date picker for the date filter has a different size than the other drop down fields. This means that the filter overlaps other UI components. And the design is different from the other filter fields.

![grafik](https://github.com/opencast/opencast-admin-interface/assets/103581881/58a00073-665a-4d66-b2d3-b0491e1269b9)

PR #627, opened 2024-06-05
Fix filter functionality and performance

### LINKED ISSUE #211: [new Admin UI] filter function does not work properly yet
On the one hand, it takes time until the event or series I am looking for is displayed.

And on the other hand, nothing is shown to me when filtering, for example, if I have previously used pageing and, for example, had called up the second results page before the search.

**Environment Information**

OS: Debian bookworm
Browser: Firefox
Opencast version: 14.4

ferishili:
This PR fixes #211 and covers #577,

The filter functionality problem (on page > 1) has been solved by going to first page and load resources!

The performance issue has been solved by introducing debounce mechanism on loading resources.
This also covers Filter profiles!


ferishili:
> in general fixed - just looks a little bit wobbly
> 
> grafik

Right, but the problem with datepicker in the filter was not covered by the initial issue #211,
Therefore, datepicker mispalcement needs to be fixed in a separate issue.

ferishili:
https://github.com/opencast/opencast-admin-interface/issues/528 here is the issue regarding datepicker UI

snoesberger:
Works mostly, but not if I'm using the start date as filter. If I go to page 3 and enter a start date filter, filtering is not working. In the browser console I do get the following error:
```
MUI X: A component is changing the uncontrolled value of a picker to be controlled.
Elements should not switch from uncontrolled to controlled (or vice versa).
Decide between using a controlled or uncontrolled valuefor the lifetime of the component.
The nature of the state is determined during the first render. It's considered controlled if the value is not `undefined`.
More info: https://fb.me/react-controlled-components
```

ferishili:
Thanks for the testing and your feedback, I will fix it right away

ferishili:
As an info, this error also happens in the current main branch, therefore, it would say, it is separate issue!

ferishili:
Fix for MUI error is available in #652

lkiesow:
Probably makes sense to merge https://github.com/opencast/opencast-admin-interface/pull/654 first

Arnei:
All other concerns seem to have been addressed (or moved), so I'll be merging this.

PR #634, opened 2024-06-05
Fix license strings not being translated in metadata dropdown

### LINKED ISSUE #559: License field content displays inconsistently in editor vs metadata
MetadataShowsALLRIGHTS_WhileEditorFullText

ALLRIGHTS_WithMetadata

Arnei:
For example, "ALLRIGHTS" would be shown instead of "All rights reservered". This is fixed now, by hopefully properly parsing listproviders that put JSON arrays as their values instead of strings.

Fixes #559.


lkiesow:
Tested this with custom labels and that works as well:

```properties
list.name=LICENSES
…
CC0={"label":"EVENTS.LICENSE.CC0", "order":8, "selectable": true}
LK={"label":"Lars' license", "order":8, "selectable": true}
LK2=Lars' second license
```

PR #635, opened 2024-06-05
Collapse roles in user table

### LINKED ISSUE #351: Collapse roles in user table
The roles column in the user table can get extremely long in a production system. The [episode ID based access control via roles](https://github.com/opencast/opencast/pull/5056) will turn this even more into a problem.

Screenshot from 2024-05-06 20-35-18

While we can't really do much about custom roles like in this screenshot, we could esily do something about the new episode ID based roles and just show something like:

> ROLE_X, ROLE_Y and 12 events

lkiesow:
The roles column in the user table can get extremely long in a production system. This patch collapses UI, API and capture agent roles.

Screenshot from 2024-06-05 13-54-02

This fixes #351

### DIFF
--- a/src/components/users/partials/UsersRolesCell.tsx
+++ b/src/components/users/partials/UsersRolesCell.tsx
@@ -7,14 +7,34 @@ const UsersRolesCell = ({
     row
 }: any) => {
 	const getRoleString = () => {
-		let roleString = "";
+		let displayRoles = [];
+		let roleCountUI = 0;
+		let roleCountAPI = 0;
+		let roleCountCaptureAgent = 0;
 
-// @ts-expect-error TS(7006): Parameter 'role' implicitly has an 'any' type.
-		row.roles.forEach((role) => {
-			roleString = roleString.concat(role.name + ", ");
-		});
+		for (const role of row.roles) {
+			if (role.name.startsWith('ROLE_UI')) {
+				roleCountUI++;
+			} else if (role.name.startsWith('ROLE_API')) {
+				roleCountAPI++;
+			} else if (role.name.startsWith('ROLE_CAPTURE_AGENT')) {
+				roleCountCaptureAgent++;
+			} else {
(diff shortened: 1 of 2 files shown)


src/components/users/partials/UsersRolesCell.tsx
```diff
@@ -7,14 +7,34 @@ const UsersRolesCell = ({
...
+			} else if (role.name.startsWith('ROLE_CAPTURE_AGENT_')) {
+				roleCountCaptureAgent++;
+			} else {
+				displayRoles.push(role.name);
+			}
+		}
 
-		return roleString;
+		if (roleCountUI > 0) {
+			displayRoles.push(`${roleCountUI} UI roles`);
```
Arnei, line 28: These should probably be translation strings?

Arnei: Resolved

src/components/users/partials/UsersRolesCell.tsx
```diff
@@ -7,14 +7,34 @@ const UsersRolesCell = ({
...
-// @ts-expect-error TS(7006): Parameter 'role' implicitly has an 'any' type.
-		row.roles.forEach((role) => {
-			roleString = roleString.concat(role.name + ", ");
-		});
+		for (const role of row.roles) {
+			if (role.name.startsWith('ROLE_UI')) {
+				roleCountUI++;
+			} else if (role.name.startsWith('ROLE_API')) {
+				roleCountAPI++;
+			} else if (role.name.startsWith('ROLE_CAPTURE_AGENT_')) {
```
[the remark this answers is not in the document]
lkiesow: Good catch. Fixed.

src/components/users/partials/UsersRolesCell.tsx
lkiesow: Now it is.

Arnei (approved):

PR #639, opened 2024-06-05
Fix incorrect error message changing time on scheduled event for the same day

### LINKED ISSUE #608: Incorrect error message when changing the date/time of scheduling
When I change the start or end time of a scheduled event for the current day, the error message "The schedule could not be updated: You cannot schedule an event to be in the past." is displayed even though the start or end time is still in the future.

grafik

jduehring:
This PR fixes #608 . 

Before this PR, the start and end date have always been modified before the conflict-check took place. This lead to the incorrect error message. This should be fixed now. 

In addition to this I added a check that not only the `endDate`, but also the `startDate` should not be in the past when editing scheduled events. In my opinion that would make more sense?


ziegenberg:
Hi @jduehring, could you address the outstanding comment(s) and rebase and fix the merge conflicts?

snoesberger:
> Actually I believe all remaining places where this occured have been fixed with this commit: [93dc23b](https://github.com/opencast/opencast-admin-interface/commit/93dc23bc79cb2d55d1e017424a9285015f1295cf)
> 
> Can you confirm this?

I think [93dc23b](https://github.com/opencast/opencast-admin-interface/commit/93dc23bc79cb2d55d1e017424a9285015f1295cf) makes this PR obsolete as it fixes also the issue https://github.com/opencast/opencast-admin-interface/issues/608.

Regarding that it only checks if the `endDate` is in the future. From my point of view this is ok, because it makes it possible to change the `endDate` of an event that is already running. Perhaps a user wants to stop a recording earlier than originally planned.

jduehring:
> > Actually I believe all remaining places where this occured have been fixed with this commit: [93dc23b](https://github.com/opencast/opencast-admin-interface/commit/93dc23bc79cb2d55d1e017424a9285015f1295cf)
> > Can you confirm this?
> 
> I think [93dc23b](https://github.com/opencast/opencast-admin-interface/commit/93dc23bc79cb2d55d1e017424a9285015f1295cf) makes this PR obsolete as it fixes also the issue #608.
> 
> Regarding the other point of this PR that it only checks if the `endDate` is in the future. From my point of view this is ok, because it makes it possible to change the `endDate` of an event that is already running. Perhaps a user wants to stop a running recording earlier than originally planned.

Agreed, I will close this PR!

Issue #612, opened 2024-06-05
Framing in "create event" window broken 


kruemelm0nsta:
Header in the "create event" window should be in a single line

![grafik](https://github.com/opencast/opencast-admin-interface/assets/131258682/4c473d7e-7891-429c-819e-b82c9d557b7c)

Depending on the language text doesn't fit in header, this is the case for German and Greek

![grafik](https://github.com/opencast/opencast-admin-interface/assets/131258682/618c4445-77e2-4258-8be2-ba26644d58ad)

kruemelm0nsta:
same on "event details"

![grafik](https://github.com/opencast/opencast-admin-interface/assets/131258682/ce96ac3c-06e5-4d3c-a894-d2bee7060b00)

in German, Greek, Spanish, French, Gallego, Italian, Polish, Slovanian, Swedish

Arnei:
Initial complaint has been addressed. If additional complaints still persist, please file a new issue.

PR #644, opened 2024-06-05
Fix wizard stepper that drifts out of window

marwyg:
Fixes the wizard steps that are going out of the window in #612.

The text of the wizard steps include word breaking in non blank spaces now. I don't know if this is a good solution, what do you think?

### DIFF
--- a/src/styles/components/_steps.scss
+++ b/src/styles/components/_steps.scss
@@ -110,4 +110,15 @@
             background: radial-gradient(ellipse at center,  rgba(238,239,240,1) 35%,rgba(146,160,171,1) 40%);
         }
     }
+
+    &.MuiStepper-horizontal {
+      .MuiStep-horizontal {
+        word-break: break-word;
+        padding: 0;
+        flex: 1 1 min-content;
+      }
+      .MuiStepButton-horizontal {
+        padding: 24px 0;
+      }
+    }
 }
--- a/src/styles/components/_steps.scss
+++ b/src/styles/components/_steps.scss
@@ -112,13 +112,13 @@
     }
 
     &.MuiStepper-horizontal {
-      .MuiStep-horizontal {
-        word-break: break-word;
-        padding: 0;
-        flex: 1 1 min-content;
-      }
-      .MuiStepButton-horizontal {
-        padding: 24px 0;
-      }
+        .MuiStep-horizontal {
+            word-break: break-word;
+            padding: 0;
+            flex: 1 1 min-content;
+        }
+        .MuiStepButton-horizontal {
+            padding: 24px 0;
+        }
     }
 }


marwyg:
This is how it looks with the change:

grafik

and here with even one more step:

grafik

src/styles/components/_steps.scss
```diff
@@ -110,4 +110,15 @@
             background: radial-gradient(ellipse at center,  rgba(238,239,240,1) 35%,rgba(146,160,171,1) 40%);
         }
     }
+
+    &.MuiStepper-horizontal {
+      .MuiStep-horizontal {
+        word-break: break-word;
+        padding: 0;
+        flex: 1 1 min-content;
```
Arnei, line 118: What does this achieve?

src/styles/components/_steps.scss
```diff
@@ -110,4 +110,15 @@
...
+      }
+      .MuiStepButton-horizontal {
+        padding: 24px 0;
```
Arnei, line 121: What does this achieve?

Arnei (changes requested):
At some point we will have to accept that we only have two lines for our text:

Bildschirmfoto vom 2024-06-05 15-33-19

kruemelm0nsta:
maybe we could work on the wordbreaking? (minor priority)

src/styles/components/_steps.scss
marwyg: Which line do you mean? Or do you mean all of them?
- So word-break makes, that words with no blanks starting to wrap like 'Zusammenfassung'
- The padding removes the spacing between the text to the edges, so the text has more room. In combination with the removal of the padding in MuiStepButton there is no padding left and the spacing is fully managed by the flexbox.
- The flex: 1 and 1 means that the elements shall grow and shrink equally. The min-content sets the initial width of the elements to the content of the smallest one.

I tested this again in Firefox and Chrome. I am no expert in CSS, so I am not 100% sure, but the removal of some of the lines will change things when you remove or add things in combination with the other changes I made or when there are more or less elements in the wizard.

src/styles/components/_steps.scss
marwyg: The padding is to remove the padding from the buttons like described in the above comment. It's to remove all padding left and right of the wizard elements to give the text more space.

marwyg:
So I tested the CSS changes again after Arnes comments and maybe it would be better to just remove the wordbreaking? I added one extra wizard step with a long text and it still works without wordbreaking:

grafik

So there is room for one additional step currently and if in the future someone should add more steps, maybe this person can deal with the styling then? I tried to test some languages, but when clicking on the above link, there seem only 3 languages work? German, English, Dutch?

kruemelm0nsta:
for the translation issue @ziegenberg is on it

src/styles/components/_steps.scss
Arnei: I see it now, thanks for the explanation.

Arnei:
> So I tested the CSS changes again after Arnes comments and maybe it would be better to just remove the wordbreaking? I added one extra wizard step with a long text and it still works without wordbreaking:
> 
> grafik
> 

Not fond of this personally, as we have to sacrifice the even spacing between the buttons for it. But I'll gladly let people outvote me on that.

src/styles/components/_steps.scss
marwyg: There are other flex configurations that I am overwriting.

https://github.com/opencast/opencast-admin-interface/assets/5593139/5d4ae7f9-ceca-4ba2-9401-64064fe46954

marwyg:
Even spacing sounds good, too :) We also can leave it as it is. That's fine by me.

src/styles/components/_steps.scss
Arnei: My bad, you are of course correct. Thank you for taking the time to make a video.

ziegenberg:
Translation issue should be fixed by now.

JulianKniephoff (approved):

JulianKniephoff:
Should we have issues tracking the open discussion points or are people satisfied with this?

Arnei:
Might have to ping the people you want to answer that directly, I think the only one with major critique was me ^^'.

JulianKniephoff:
Fair 😛

PR #646, opened 2024-06-05
Remove dead code

Arnei:
Removes left over code from modernizing redux
(which is still not completely done).


ziegenberg (approved):
Clicked through the relevant parts of the UI, saw no additional errors or warnings and it still works.

PR #652, opened 2024-06-05
Fix MUI error when selecting a date as for StartDate filter

### LINKED ISSUE #636: Seleting a value for Datepicker in the filters throws error
When selecting a value for the Startdate in the filter section, it throws an error in the console log, which needs to be take care of:

Screenshot 2024-06-05 at 14 14 32

ferishili:
This PR fixes #636,

Upon selecting the StartDate filter and adding selecting a date, an error occurs in MUI level, which can be seen in the devtool console!

It is solved simply by making the endDate and startDate value so called "uncontrolled?" by making it optional as to force passing null!

### DIFF
--- a/src/components/shared/TableFilters.tsx
+++ b/src/components/shared/TableFilters.tsx
@@ -469,7 +469,7 @@ const FilterSwitch = ({
 						autoFocus={true}
 						inputRef={startDateRef}
 						className="small-search start-date"
-						value={startDate}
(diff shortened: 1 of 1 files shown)


lkiesow:
Probably makes sense to merge https://github.com/opencast/opencast-admin-interface/pull/654 first

Arnei (approved):
Odd fix, but can't argue with results!

PR #654, opened 2024-06-05
Fix style of filters date pickers

dennis531:
Close https://github.com/opencast/opencast-admin-interface/issues/528


ziegenberg (approved):
Looks better now

PR #657, opened 2024-06-05
Fix unnecessary green ticks in Metadata wizard

### LINKED ISSUE #629: necessity of green hooks on Presenter and Contributor in Event Details
leftover or have they a function?
grafik

ferishili:
This PR fixes #629,

The problem comes from comparing arrays like other types, which in this case of `!==` is always true.
The correct way of comparing array is to convert them into json string and then check if they are equal!

With this PR, you will see the green tick only after changing the value of the inputs,  which is the intended behavior!

### DIFF
--- a/src/components/shared/wizard/RenderMultiField.tsx
+++ b/src/components/shared/wizard/RenderMultiField.tsx
@@ -269,7 +269,7 @@ const ShowValue = ({
 				{showCheck && (
 					<i
 						className={cn("saved fa fa-check", {
-							active: initialValues[field.name] !== field.value,
+							active: JSON.stringify(initialValues[field.name] ?? []) !== JSON.stringify(field.value ?? []),
 						})}
 					/>
 				)}


geichelberger:
In the PR https://github.com/opencast/opencast-admin-interface/pull/398, the ticks were still working as expected and were broken in PR https://github.com/opencast/opencast-admin-interface/pull/399 as those got merged after the latest release develop is not broken. Maybe 399 introduced the changes or something that got merged in between those PRs.

Arnei (approved):
Regardless of how this came to be broken, this does fix the issue in a reasonable manner I think. So this gets an approval from me.

lkiesow:
+1 for merging the pull request like this. It's definitely an improvement over having random green check marks.

That being said, I think we would get just rid of them. They existed before we had a save button to indicate that a modified field had been saved. Now, I think they are just confusing and not helpful at all.

Arnei:
> That being said, I think we would get just rid of them. They existed before we had a save button to indicate that a modified field had been saved. Now, I think they are just confusing and not helpful at all.

Disagree on that. I think they are useful as feedback that your input has been recognized. They could also indicate that your input was valid, but alas our validation is ... lacking right now.

geichelberger:
Indicating which field changed is nice; the question is whether adding a green tick is the way to go.

lkiesow:
Let's move this discussion to a separate issue.

Any objections to merging this @Arnei, @geichelberger, @ziegenberg?

geichelberger:
+1 merging

geichelberger (approved):

ziegenberg:
It's fine with me

Arnei:
Fine by me!

lkiesow:
Then let's merge this.

PR #659, opened 2024-06-05
Fix Set Filter Profile dialog behavior and styles

### LINKED ISSUE #618: Improve saved filters dialog
osfilter

The save filter dialog isn't very intuitive. Hitting save doesn't save anything but switches to a form with a title and description. This could be changed to "add" or something similar. 
Hitting cancel also shouldn't delete the filter, but when you click on the edit icon of a specific filter and then click on cancel the filter disappears.

ferishili:
This PR fixes #618,

As it is stated in issue, this wizard/dialog doesn't behave as it should specially when clicking cancel button (it closes the dialog and in edit mode removes the profile) => that is no go!

This PR fixes this bad behavior by bringing the `currentlyEditing` variable to live once again and set its type as `FilterProfile` or null. So on `cancelEditProfile` we now only check if it was in editing mode and create a record of that instead of letting it be removed for good!

NOTE: A new translation key under "TABLE_FILTERS > PROFILES > ADD" is added. which replace the save text on the button when in adding new profile mode!

Apart from that, the style of the form in add/edit mode gets a proper look now!

### DIFF
--- a/src/components/shared/TableFilterProfiles.tsx
+++ b/src/components/shared/TableFilterProfiles.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from "react-i18next";
 import cn from "classnames";
 import { getFilterProfiles } from "../../selectors/tableFilterProfilesSelectors";
 import {
-	cancelEditFilterProfile,
+	FilterProfile,
 	createFilterProfile,
 	removeFilterProfile,
 } from "../../slices/tableFilterProfilesSlice";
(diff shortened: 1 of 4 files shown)


Arnei:
I can get a saved filter to be displayed in the list anymore, if i close the edit menu with the x button:
[Bildschirmaufzeichnung vom 2024-06-06, 14-01-34.webm](https://github.com/opencast/opencast-admin-interface/assets/14070005/2d6db76b-0649-4ad9-9740-376175ddfd82)

Arnei (changes requested):
Works as advertised and code looks good to me too. Just found the one thing in the comment above.

ferishili:
Thanks, I will fix it. That faulty behavior was meant to be fixed as well.

Arnei (approved):

PR #660, opened 2024-06-05
Disable statistics tab by default

### LINKED ISSUE #357: Statistics tab enabled by default
- Open the menu on the left side of the admin interface
- Statistics shows up as last option
  - It shouldn't unless it is specifically configured
- Click on Statistics
  - The whole UI breaks

lkiesow:
By default the statistics view is broken and you need additional configuration. Therefor, it should be disabled by default in the admin interface.

Unlike the old interface, this patch does not read the actual data from statistics providers, but uses a configuration option to determine if the statistics feature should be an option in the UI.

The configuration and documentation is added in pull request https://github.com/opencast/opencast/pull/5909

This fixes #357

### DIFF
--- a/src/components/shared/MainNav.tsx
+++ b/src/components/shared/MainNav.tsx
@@ -17,7 +17,10 @@ import {
 import { fetchThemes } from "../../slices/themeSlice";
 import { fetchFilters, fetchStats } from "../../slices/tableFilterSlice";
 import { setOffset } from "../../actions/tableActions";
-import { getUserInformation } from "../../selectors/userInfoSelectors";
+import {
+	getOrgProperties,
+	getUserInformation
+} from "../../selectors/userInfoSelectors";
 import { hasAccess } from "../../utils/utils";
 import { fetchServices } from "../../slices/serviceSlice";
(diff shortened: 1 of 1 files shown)


src/components/shared/MainNav.tsx
```diff
@@ -68,6 +71,9 @@ const MainNav = ({
 	let navigate = useNavigate();
 
 	const user = useAppSelector(state => getUserInformation(state));
+	const orgProperties = useAppSelector(state => getOrgProperties(state));
+
+	const statisticsEnabled = orgProperties['admin.statisttics.enabled']?.toLowerCase() === 'true';
```
gregorydlogan, line 76: Sigh.  *Now* is when I notice the misspelling here, not before I merged opencast/opencast#5909

gregorydlogan (changes requested):

src/components/shared/MainNav.tsx
gregorydlogan: I've fixed the OC side of the spelling issue with opencast/opencast#5918

lkiesow: Fixed this pull request

PR #664, opened 2024-06-06
increase contrast ratio in wizard modal

### LINKED ISSUE #640: Modal wizard: multi step bar lacks of contrast
Text has a contrast ratio of 2.32:1 and the bullet points as well.

rlucke:
fix #640
increase contrast ratio of steps items in the wizard modal.

Before:
Screenshot 2024-06-06 093717

After:
Screenshot 2024-06-06 093648


rlucke:
> So this is probably totally bias from me being used to a lower contrast, but the difference in contrast between step items and the stuff below (e.g. metadata) seems rather krass to me. Maybe we generally want to have darker fonts across the board?

I guess that would be another issue. For this particular element, all I can say is that having light gray on even lighter gray is a bad idea. Not just for accessibility reasons, but also in terms of UX design.

Arnei:
Aight, let's go with this.

PR #626, opened 2024-06-05
Fix handling of long strings in modal windows and table

marwyg:
Should fix: https://github.com/opencast/opencast-admin-interface/issues/549

This PR fixes issues with long strings in metadata for example the title. Especially the title creates problems when opening modal windows like the event details or the tables.

For testing: Please add data everywhere where you can: Users, Series, Events, etc, etc. and add long strings. The Modal Windows / Wizards should look fine if you do so. Also, when you save the data, take a look at the tables and search for your entry, the entry should cause no problems in the table. Important: It makes a difference when you put in a string with blanks or not, so maybe you want to test both cases like:

`LONGDATALONGDATALONGDATALONGDATALONGDATALONGDATALONGDATALONGDATALONGDATALONGDATA`
and
`LONGDATA LONGDATA LONGDATA LONGDATA LONGDATA LONGDATA LONGDATA LONGDATA LONGDATA LONGDATA`

PR #666, opened 2024-06-06
Fix layout of file upload in add event dialog

lkiesow:
This patch fixes the table layout of the upload section in the add event dialog which had overlapping elements.

Screenshot from 2024-06-06 01-05-45

Not the actions again have their own small column.

Screenshot from 2024-06-06 11-34-11

The problem was caused by #626


kruemelm0nsta:
grafik

looks OK again

Arnei (approved):

PR #667, opened 2024-06-06
Fix file upload labels 

### LINKED ISSUE #661: Long file names break the add event dialog
If you select to upload a file with a long name, the elements will run out of the dialog:

Screenshot from 2024-06-06 00-06-27

lkiesow:
The file upload dialog in the add events dialog does not work well with
long filenames and even renders the filename twice. This patch fixes
that.

Screenshot from 2024-06-06 12-00-49

This patch is based on https://github.com/opencast/opencast-admin-interface/pull/666 which fixes the broken table in that dialog.

This fixes #661

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -148,10 +148,6 @@ div.main-view.stub, div.full-col > div.stub {
     text-overflow: ellipsis;
   }
 
-  table {
-    table-layout: fixed;
-  }
-
   .editable {
     word-wrap: anywhere;
   }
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -493,9 +489,9 @@ input.disabled, select.disabled {
 
 // Use HTML5 input field for file upload
 .file-upload > input[type="file"] {
-  width: inherit !important;
+  width: 300px;
```
Arnei, line 492:
```suggestion
  width: 100%;
```
How about percentage width? Allows us to use more of the available space for displaying file name if file names become somewhat long. Also should not cause any overflow.

Arnei (changes requested):

src/styles/main.scss
lkiesow: Fixed

Arnei (approved):

PR #671, opened 2024-06-06
Use button tag for stat filter items

### LINKED ISSUE #624: Filters cannot be accessed with the keyboard
Screenshot 2024-06-05 123632

Filters in the header area cannot be accessed with the keyboard

rlucke:
To access the stat filters with the keyboard they need to be a button or have tabIndex={0} and need a key listener.
Better a11y implementaion is always the simple one. So the div is now a button and with some css it looks the same. 

fixes #624


ziegenberg:
Hi @rlucke, could you fix the merge conflicts? I'd like to review your patch.

Arnei (approved):

PR #676, opened 2024-06-06
Move build instructions to Actions workflow

lkiesow:
This patch removes the build shell script in favor of having all build instruction in the GitHub Actions workflow. This is hopefully less confusing for developers to work with since everything related to creating a release is now at one place.


ziegenberg:
Looks reasonable to me.

PR #680, opened 2024-06-07
Add text to icon buttons in the Edit Table dialog.

### LINKED ISSUE #599: Icon Buttons in "Edit Table View" dialog must have discernible text
All the icon buttons in the "Edit Table View" dialog do not have discernible text.

This could be:

- inner text
- `aria-label` attribute
- `aria-labelledby` pointing to element with text

This has critical impact on blind people.

luniki:
Fixes #599.

It's all about the green and red buttons in this dialog:

image


lkiesow (approved):
This is definitely an improvement, and I'm merging the patch like this.
In the future, we might also want to add tooltips though?

PR #682, opened 2024-06-08
Correctly display “00” in event details scheduling

### LINKED ISSUE #611: Event Scheduling: "00" not correctly displayed
The scheduling times of a planed event are not displayed correctly if "00" is selected:
grafik

lkiesow:
Since `00` is a `falsy` value, it wasn't used as value in the scheduling tab input fields. This patch fixes the problem by only checking for `null` and `undefined` instead.

This fixes #611


reiferschris (approved):
I can confirm that this works as it should :)

PR #684, opened 2024-06-08
Fix admin interface crashing when rendering filters

### LINKED ISSUE #526: Click on an entry in a cell makes OC crash

lkiesow:
This fixes the admin interface crashing if there is no preset label for a given filter value. This will be the case, for example, if you filter for usernames.

Note that this is only reproducible with the mock data right now since there seems to be a bug with the production data that prevents rendering the value completely in that case, so you never reach this code 🙈

This fixes #526


Arnei (approved):
> "This sound like something typescript could have prevented".
> *sees @ts-expect-error*
> "I see".

This does work for now though, so let's get it in.

PR #685, opened 2024-06-08
Enable source maps

lkiesow:
Source maps are really valuable when debugging errors. They are disabled by default. This pull request enables them.


ziegenberg (approved):
Looks reasonable.

PR #686, opened 2024-06-09
Scheduling information in event details for non-scheduled event

### LINKED ISSUE #413: Opening event without scheduling data produces multiple stack traces
When opening an event without scheduling data (e.g. an upload), the scheduling tab in the event details modal is shown:

Screenshot from 2024-05-22 20-13-26

The following error is logged to the console:

Screenshot from 2024-05-22 20-15-19

When opening the scheduling tab, it has no content:

Screenshot from 2024-05-22 20-14-46

The following error is logged to the console:

Screenshot from 2024-05-22 20-14-58

----

Interestingly, when opening an event **with** scheduling data first and then opening an event without scheduling data, we "only" get the 404 for the REST call for the `scheduling.json`, and the event modal does not show the scheduling tab:

Screenshot from 2024-05-22 20-26-59

lkiesow:
This patch fixes the problems that

- the scheduling tab is shown for non-scheduled event in event details
- a stack trace is logged as error when no scheduling information could be retrieved

This fixes #413

### DIFF
--- a/src/components/events/partials/modals/EventDetails.tsx
+++ b/src/components/events/partials/modals/EventDetails.tsx
@@ -129,7 +129,7 @@ const EventDetails : React.FC<{
 			accessRole: "ROLE_UI_EVENTS_DETAILS_SCHEDULING_VIEW",
 			name: "scheduling",
 			hidden:
-				!hasSchedulingProperties && hasAnyDeviceAccess(user, captureAgents),
+				!hasSchedulingProperties || !hasAnyDeviceAccess(user, captureAgents),
 		},
 		{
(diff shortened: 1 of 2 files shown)


Arnei (commented):
Can confirm that the scheduling tab only shows up scheduled events now. The request for scheduling information is still sent either way, so a stack trace is still logged for non-scheduled events (as a debug message instead of an error message now). Is that what you intended?

lkiesow:
Yes, that's intentional. Not sure if you could prevent the request in the first place. But right now, the 404 just shows that there is no schedule which is okay to log as debug information but not as an error. And I just tried to get this fixed as quickly as possible. We can still improve this later.

Arnei (approved):

PR #687, opened 2024-06-09
Make themes optional

### LINKED ISSUE #523: Make themes optional
Not all workflows/installations support theming which can be confusing as a user. As an admin, it would be great if I can hide these features in the admin interface.

This means hiding the systems option in the left-hand menu:
Screenshot from 2024-06-04 10-25-11

And hiding the themes options in the create event dialog:
Screenshot from 2024-06-04 10-24-55

lkiesow:
Not all workflows/installations support theming which can be confusing as a user. This patch makes them optional in the admin interface and disable them by default. Admins can easily enable them in Opencast's organization configuration.

This fixes #523

### DIFF
--- a/src/components/events/partials/modals/SeriesDetails.tsx
+++ b/src/components/events/partials/modals/SeriesDetails.tsx
@@ -9,7 +9,7 @@ import {
 	getSeriesDetailsThemeNames,
 	hasStatistics as seriesHasStatistics,
 } from "../../../../selectors/seriesDetailsSelectors";
-import { getUserInformation } from "../../../../selectors/userInfoSelectors";
+import { getOrgProperties, getUserInformation } from "../../../../selectors/userInfoSelectors";
 import { hasAccess } from "../../../../utils/utils";
 import SeriesDetailsAccessTab from "../ModalTabsAndPages/SeriesDetailsAccessTab";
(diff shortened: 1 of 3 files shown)


stefanosgeo:
According to the #523 , themes tab in series modal should also be hidden. In this PR themes are hidden in the left hand menu but not in the series.

lkiesow:
> …themes tab in series modal should also be hidden.

Where did you find that? In which dialog precisely?
Right now, a series should still show a themes tab in the series details if a theme is already set. Do you think we should hide that tab in that case as well? May make sense…

stefanosgeo:
> Where did you find that? In which dialog precisely?

In the deploy URL, there are two series listed: the first with a theme and the second without. I selected the first one and saw the themes tab.

The behavior of showing or hiding the themes tab based on whether a series already has a theme was not mentioned in either the issue or the PR. Therefore, I did not test it further. :roll_eyes: Sorry about that.

Additionally, I’m not sure if it’s a good idea to always hide the themes tab (in series). The additional series information (theme) should be displayed. I suggest keeping the tab with the condition you described.

lkiesow:
Okay, then let's leave it like this. And sorry about the missing information. I just made the decision while I implemented this.

Arnei (approved):
Works and makes sense to me.

PR #688, opened 2024-06-09
Center tooltips on main menu items

lkiesow:
The tooltips on the main menu items start really far away from the icons. This patch makes them activate only on the icon.

<center>

</center>

### DIFF
--- a/src/components/shared/MainNav.tsx
+++ b/src/components/shared/MainNav.tsx
@@ -233,92 +233,92 @@ const MainNav = ({
 							{/* todo: more than one href? how? roles? (see MainNav admin-ui-frontend)*/}
 							{hasAccess("ROLE_UI_NAV_RECORDINGS_VIEW", user) &&
 								(hasAccess("ROLE_UI_EVENTS_VIEW", user) ? (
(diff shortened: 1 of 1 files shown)


stefanosgeo:
It does what it should, but there are (?) issues. For example when using the dev tools for responsive design, the tooltips are still not aligned properly. Additionally, from the screenshot(text "Event" overlapping with status tabs, it appears there are other problems as well.

Screenshot from 2024-06-10 11-54-20

lkiesow:
You are absolutely right. Can you file the responsive design problems as separate issues?
This pull request just aims to be a small improvement for the tooltips in the main menu.
Adressing the overlapping texts on small screens is likely a bit more work.

stefanosgeo:
> You are absolutely right. Can you file the responsive design problems as separate issues? This pull request just aims to be a small improvement for the tooltips in the main menu. Adressing the overlapping texts on small screens is likely a bit more work.

I will. Then this PR addresses and fixes the described problem.

lkiesow:
Thanks for reviewing this. Since it was reviewed and later discussed at the technical meeting, I'm merging this.

PR #690, opened 2024-06-11
Using correct config key to hide imprint and privacy policy links

gregorydlogan:
This PR uses the correct config key to disable the imprint and privacy policy links

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -22,6 +22,7 @@ const Footer: React.FC = () => {
 	const lastModified = user?.ocVersion?.['last-modified']
 		? new Date(user.ocVersion['last-modified']).toISOString().substring(0, 10)
 		: 'unknown';
+	const aboutEnabled = orgProperties['org.opencastproject.admin.display_about']?.toLowerCase() === 'true';
 
 	return (
 		<footer id="main-footer">
@@ -40,7 +41,7 @@ const Footer: React.FC = () => {
 							)}
 						</li>
 					)}
-					{!!orgProperties && !!orgProperties["admin.display_about"] && (
+					{aboutEnabled && (
 						<>
 						<li><Link to="/about/imprint">{t("ABOUT.IMPRINT")}</Link></li>
 						<li><Link to="/about/privacy">{t("ABOUT.PRIVACY")}</Link></li>
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -40,7 +40,7 @@ const Footer: React.FC = () => {
 							)}
 						</li>
 					)}
-					{!!orgProperties && !!orgProperties["org.opencastproject.admin.display_about"] && (
+					{!!orgProperties && !!orgProperties["admin.display_about"] && (
 						<>
 						<li><Link to="/about/imprint">{t("ABOUT.IMPRINT")}</Link></li>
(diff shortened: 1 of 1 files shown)


src/components/Footer.tsx
```diff
@@ -40,7 +40,7 @@ const Footer: React.FC = () => {
 							)}
 						</li>
 					)}
-					{!!orgProperties && !!orgProperties["org.opencastproject.admin.display_about"] && (
+					{!!orgProperties && !!orgProperties["admin.display_about"] && (
```
lkiesow, line 43: Right now, the other one is in the configuration of `r/16.x`. Do you want to change this?

```
❯ grep -rn admin.display_about org.opencastproject.organization-mh_default_org.cfg
140:prop.org.opencastproject.admin.display_about=false
```
and
```
❯ curl -s https://develop.opencast.org/info/me.json | gron | grep admin.display_about
json.org.properties["org.opencastproject.admin.display_about"] = "false";
```

Do you plan on changing the configuration in Opencast?

gregorydlogan: So there are two issues here:
- The config key we're looking for is prefixed with `prop.org.opencastproject.` in the config file, but the `orgProperties` object removes that previx.  The presence of `prop.org.opencastproject.admin.display_about` in the org config file is enough to cause this check in the front end to be true, *regardless* of the value of the config key.  Setting this key to false does not turn the UI elements off.
- The backend config does not have the default value commented out.  This was addressed with opencast/opencast#5918.  I can see the argument that the commenting out should also be applied to r/15.x, I just didn't file it that way.  Low impact IMO, and at this point applying that change might just cause more conflicts for end users.  The default at least matches the value, even if it doesn't work properly without this patch :D

lkiesow: If I add `console.info(orgProperties)`, I get:
Screenshot from 2024-06-12 00-18-16

So, the property uses the long name.
The problem is that the value is a string and not a boolean.
That means any non-empty string is `true`.

Something like this should actually fix the problem:

```diff
diff --git a/src/components/Footer.tsx b/src/components/Footer.tsx
index de264430b0..34449449de 100644
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -22,6 +22,7 @@ const Footer: React.FC = () => {
 	const lastModified = user?.ocVersion?.['last-modified']
 		? new Date(user.ocVersion['last-modified']).toISOString().substring(0, 10)
 		: 'unknown';
+	const aboutEnabled = orgProperties['admin.display_about']?.toLowerCase() === 'true';
 
 	return (
 		<footer id="main-footer">
@@ -40,7 +41,7 @@ const Footer: React.FC = () => {
 							)}
 						</li>
 					)}
-					{!!orgProperties && !!orgProperties["admin.display_about"] && (
+					{aboutEnabled && (
 						<>
 						<li><Link to="/about/imprint">{t("ABOUT.IMPRINT")}</Link></li>
 						<li><Link to="/about/privacy">{t("ABOUT.PRIVACY")}</Link></li>
```

Btb. I actually prefer the short key, but we need to ensure that the one in here and in Opencast's org config match.

gregorydlogan: Ah, ty.  I've got it working now.  I'm going to use the long key, though I prefer the short one as well.  One day maybe we'll go through and rework all the keys, but for now it's one less PR to handle prior to 16.0 :D

lkiesow (approved):

PR #691, opened 2024-06-11
All metadata should be in the metadata step of the add event wizard

### LINKED ISSUE #487: All metadata should be in the metadata step of the add event wizard
The “source” step of the add event wizard also contains metadata. This doesn't really make sense since there is a whole step dedicated to metadata. We should move “Start date” in there.

The problem: This field should only be shown for a file upload. Not for a scheduled event.
The solution: Make “Source” step 1 and metadata step 2. We then know if we need a start date once we are in the metadata tab.

Screenshot from 2024-05-28 23-01-01

lkiesow:
The “source” step of the add event wizard also contains metadata. This doesn't really make sense since there is a whole step dedicated to metadata. We should move “Start date” in there.

- The problem: This field should only be shown for a file upload. Not for a scheduled event.
- The solution: Make “Source” step 1 and metadata step 2. We then know if we need a start date once we are in the metadata tab.

This fixes #487


mtneug:
Two other fields missing are "location" and "duration" (used for calculating the end date in the Admin UI). I set this for file uploads if I know these files are from a recording (e.g. CA did the recording, but editing was done locally). Is this in the scope of this PR, or should I create another issue?

lkiesow:
You mean, you want these two additional fields in the recording metadata section? I think that's unrelated enough to not wait for that when merging this pull request. Nevertheless, I'll try taking a peek at the issue since I now kind of know the code and I hope it's not too hard to do.

mtneug:
> You mean, you want these two additional fields in the recording metadata section? I think that's unrelated enough to not wait for that when merging this pull request. Nevertheless, I'll try taking a peek at the issue since I now kind of know the code and I hope it's not too hard to do.

Fair. I created issue #742 to track this.

src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
```diff
@@ -84,6 +81,7 @@ const NewSourcePage = <T extends RequiredFormProps>({
 		dispatch(fetchRecordings("inputs"));
 
 		// validate form because dependent default values need to be checked
+    console.info(formik);
```
Arnei, line 84: Leftover code

src/components/events/partials/ModalTabsAndPages/NewSeriesMetadataPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Field, FormikProps } from "formik";
+import RenderField from "../../../shared/wizard/RenderField";
+import WizardNavigationButtons from "../../../shared/wizard/WizardNavigationButtons";
+import RenderMultiField from "../../../shared/wizard/RenderMultiField";
+import { MetadataCatalog } from "../../../../slices/eventSlice";
+
+/**
+ * This component renders the metadata page for new events and series in the wizards.
+ */
+const NewSeriesMetadataPage = <T,>({
```
Arnei, line 12: This is a lot of code duplication just because we want to add start date to the event metadata tab. I would prefer if we could avoid this.

Arnei (changes requested):
Works. But I'm not totally happy with it ^^'.

Does the start date field require special attention? If not, I would prefer if start date was grouped together with the other metadata. Having it in its own special box might set wrong expectations, i.e. "Is 'recording metadata' a different metadata catalog from 'metadata'?".
Bildschirmfoto vom 2024-07-11 08-57-11

src/components/events/partials/ModalTabsAndPages/NewSeriesMetadataPage.tsx
lkiesow: I've tried that first, but there was more special handling between both since not only the metadata is different, but also the control elements. At the ed it seemed to complicated to me and I gave up. So… yes, I at least thought  about this.

lkiesow:
Fixed the minor issues and merged the different metadata sections. Now they shouldn't seem like different catalogs any longer.

Arnei:
> Now they shouldn't seem like different catalogs any longer.

Can confirm, looks good to me now.

I *was* going to complain that we should need extra code for rendering start_date, as it is part of the standard metadata catalog and therefore should be able to be rendered just like any other metadata field. But it turns the reason we cannot do that is ultimately the backends fault, as the relevant endpoint (admin-ng/event/new/metadata) explicitly excludes start_date (among other fields). Whether or not this is reasonable likely requires additional consideration, so I'm ok with leaving this as it is for now.

Arnei (changes requested):
If you schedule a new event and have asset upload configured, the "Previous" button on the metadata tab does nothing. It should bring me back to the "Source" tab.

lkiesow:
Rebased to resolve merge conflicts and fixed the next/previous problem.

Arnei (changes requested):
Can confirm, it is fixed.

Did find another small issue while testing though, sorry :)
The validation for the main metadata is not happening anymore. Effectively this means you could create an event without a title.

lkiesow:
Rebased to fix resolve conflicts.

> The validation for the main metadata is not happening anymore.

Fixed.

Arnei:
Can confirm, validation is happening again.

But the "Previous" button on the metadata tab does nothing again. Seems to only occur if I want to schedule an event.
Here is a video of me clicking around:
[Bildschirmaufzeichnung vom 2024-07-26, 14-48-04.webm](https://github.com/user-attachments/assets/e3f03fd2-384f-4f2a-bab7-138555683725)

PR #695, opened 2024-06-13
Modernize redux: tableSlice

Arnei:
**Hello potential reviewer! Even if the number of code changes seems daunting to you and you don't know what this is about, please consider testing this anyway. This patch should not change any behaviour in the UI, so testing if everything still works as expected would be greatly appreciated!**

Helps with https://github.com/opencast/opencast-admin-interface/issues/213.

Switching to redux toolkit for our central table reducers.

This is the last big PR on this topic. After this, I would consider our migration to redux toolkit complete (finally), although there certainly is still room for improvement.


gregorydlogan:
LGTM

Issue #533, opened 2024-06-04
Edit multiple events of the same series are no more grouped by day of the week


stefanosgeo:
In Opencast 14 editing multiple events form the same series (select x events from a specific series->actions->edit scheduled events)  would group the events by day of the week 
![Screenshot from 2024-06-04 11-18-28](https://github.com/opencast/opencast-admin-interface/assets/3663204/2fbb217c-d936-45f7-b779-cc3f6c94cebe)

This is not happening any more 
![Screenshot from 2024-06-04 11-18-45](https://github.com/opencast/opencast-admin-interface/assets/3663204/7e2ca31a-7ca9-4d3d-aeef-ba0e4254e2bc)

Bug or feature?

reiferschris:
Definitively not feature ;) 
IMHO

Arnei:
This was fixed, closing.

PR #697, opened 2024-06-14
Group events in bulk edit scheduled events tab

Arnei:
Aims at fixing #533.

You can bulk edit scheduled events in the admin ui. However, they were not grouped by weekday anymore, but each event was listed individually, which is not really bulk editing at all. This groups the events by weekday again.

Known issues:
- ~The old ui also allow bulk editing of capture agent inputs. This was and still is missing from the new ui. Want to bring this back, but will have to investigate (might need backend changes).~
   - Seems the old admin ui did not fully support inputs either? Would leave that for a separate PR in that case.
- ~The old ui would also allow bulk editing for metadata title and series, regardless of weekday. This is still missing. Planning on bringing that back.~
  - Should be fixed now.

Currently looks like:
Bildschirmfoto vom 2024-06-14 15-37-19

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/EditScheduledEventsEditPage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/EditScheduledEventsEditPage.tsx
@@ -111,6 +111,11 @@ const EditScheduledEventsEditPage = <T extends RequiredFormProps>({
 		return result;
 	}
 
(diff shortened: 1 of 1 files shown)


lkiesow (approved):
Tested this and seems to work fine.

@JulianKniephoff, a month ago, you requested a review on this. Do you still want to do that review? Or can we merge this?

PR #698, opened 2024-06-17
Allow tracks in event asset upload

### LINKED ISSUE #694: Upload subtitle for existing events via Asset Upload not possible
How to reproduce it:

1. Enable ` EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.SUBTITLES={"id": "track_subtitles_option", "type":"track", "flavorType": "captions","flavorSubType": "source", "tags": "generator:unknown", "displayOrder":3, "accept": ".vtt"}` in (`/etc/opencast/lisproviders/event.upload.asset.options.properties`)
2. New Admin UI
3. Event Details --> Assets  -- > Option `Add Asset` is missing.

add_asset_new_admin

In the old Admin UI looks like this:

add_asset_old_admin

Opencast Version: 15.5

Arnei:
Old logic dictates that media of type track cannot be in the asset upload tab of the event details or new event wizard. Since subtitles are now tracks, this logic finally breaks.

Instead of relying on checking for type track to differentiate between options for source upload and asset upload, this commit instead uses the key in the list provider that is there for this exact purpose.

To review this, you will need an Opencast with Asset Upload configured. Check the Opencast docs on how to do that https://docs.opencast.org/develop/admin/#configuration/admin-ui/asset-upload/#manual-asset-upload. This changes behaviour for both the event details and the create event dialog.

Fixes #694.

Edit: Furthermore, this also adds support for `showForNewEvents` and `showForExistingEvents`, two keys that specifically specify whether an asset upload option should show up for new events or existing events. Previously, an asset upload option would always show for new and existing events, but now this can be configured in detail.


snoesberger:
I have tested this PR with our test environment. I now have two steps in the event creation dialogue where I can upload a subtitle file:
grafik
grafik

If I add different subtitle files in both steps, only the one in the second step is added to the event. It should only be possible to add a subtitle file at one point in the create event dialogue.

Adding a subtitle file to an existing event in the event modal works fine.

mtneug:
I still favour adding subtitles on the second page so I would like to have the option to configure Opencast in that way. But having the same file picker on two pages is obviously wrong.

JulianKniephoff:
Just nagging @Arnei here that CI is broken since your last conflict resolution. Should be an easy fix.

mtneug:
So you would have:

* `EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION` = new and existing events
* `EVENTS.EVENTS.EXISTING.UPLOAD_ASSET.OPTION` = existing events
* `EVENTS.EVENTS.NEW.SOURCE.UPLOAD` = new events, but on the first page

I find this confusing. What about having only one key prefix for everything and adding properties to the JSON that determine where this gets included?

```json5
{
  ...
  "group": "main", // main = first page, secondary = second page, but I'm not tight to this property
  "showForNewEvents": true,
  "showForExistingEvents": false,
  ...
}
```

Arnei:
My idea was that keys with `NEW` would only be for new events, and keys with `EXISTING`  would only be for existing events. So `EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION` would then only be for new events, not new and existing.

But I can still see why that would be confusing. Adding properties to the JSON also seems good.

snoesberger:
Ok, I hadn't in mind being able to configure the behaviour of the asset upload.

> I do agree though that it should be possible to configure subtitles in the asset upload for existing events, without also adding them to the asset upload step for creating new events. 

I also agree with this and would prefer the solution with the properties in JSON format suggested by @mtneug as I think it is easier to understand.

> That I would consider a bug, I would expect both files to be added. Took a quick look, and this seems to be an issue in the backend for multiple files with the same flavor. Personally, I would consider this bug a seperate issue from this PR as this PR did not add the bug, merely brought it to light. But I'm open to persuasion :)

I agree, we should create a separate issue for this.

gregorydlogan (approved):
Works as expected.  Like #1006, this should be held until just prior to the next release to prevent breakage, correct?

PR #699, opened 2024-06-18
Set initial sort column to "title"

### LINKED ISSUE #610: Event view table breaks formatting after a second or two on main branch
This must have been introduced by a recent change, and might not be apparent on larger screens. This is on the up-to-date main branch. See screen capture:

https://github.com/opencast/opencast-admin-interface/assets/94838646/97c037c2-23fb-4630-a934-ba01347f643f

Arnei:
When initially loading the admin ui, the events table will  reorder itself after a few moments. This avoids the reordering by initializing table sorting parameters with "title" for the "sort by column" parameter.

Fixes #610

### DIFF
--- a/src/reducers/tableReducers.ts
+++ b/src/reducers/tableReducers.ts
@@ -50,7 +50,7 @@ const initialState = {
 	resource: "",
 	pages: [],
 	columns: [],
-	sortBy: "",
+	sortBy: "title",
 	predicate: "",
 	reverse: "ASC",
 	rows: [],


Arnei:
Note: Sort by date instead.

Arnei:
The latter, custom sort settings are currently not preserved.

lkiesow:
This doesn't seem to work. I've just tested 2e4b2da using `npm start` and I see three stages when initializing the table

- No data (`00:00:04`)
- Event “Espresso” as first event (`00:00:09`)
- Event “Test Scheduling” as first event (`00:00:13`)

Here is the (slowed down) recording:
https://github.com/opencast/opencast-admin-interface/assets/1008395/0eb57f72-6d31-47ad-9ba0-88531944c52e

Arnei:
Damn it, you're right. Curse you descending sorting order. I'll see what I can do.

Arnei:
Should be better now.

lkiesow (approved):

PR #706, opened 2024-06-19
Ignore top-level `.vite` output folder

JulianKniephoff:


lkiesow:
I'm fine with the change since it shouldn't hurt, but I've never seen that folder being generated. Is there a specific condition for this to happen?

JulianKniephoff:
Interesting, now I can't reproduce it either. 👀

JulianKniephoff:
No idea how I did this. Apparently normally this folder should live in `node_modules`, and that's where I find it reproducibly now. Closing. 🤷‍♀️

Issue #708, opened 2024-06-19
Remove `rollup-preserve-directives`


JulianKniephoff:
This was added in #709 to work around https://github.com/vitejs/vite/issues/15012. Once that is resolved, we can probably get rid of it.

PR #709, opened 2024-06-19
Get rid of Vite build warnings

JulianKniephoff:
There are quite a few errors like the following in the output of `npm run build`:

    Error when using sourcemap for reporting an error: Can't resolve original location of error.

They all relate to the `@mui` package(-s). This seems to be related to https://github.com/vitejs/vite/issues/15012.
I don't fully understand it, but apparently Rollup, which is used by Vite internally somehow,
doesn't preserve enough information to generate source maps properly.

This patch adds a plugin to do exactly that. We should probably remove it in the long run,
if the upstream issue(-s) ever get fixed. I created #708 to track that.

/cc https://github.com/opencast/opencast-admin-interface/pull/685

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,7 @@
         "@types/uuid": "^9.0.8",
         "@vitejs/plugin-react-swc": "^3.7.0",
         "prop-types": "^15.8.1",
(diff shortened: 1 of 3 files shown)


lkiesow (approved):
Seems to work :+1:

Issue #713, opened 2024-06-19
Change ESLint configuration


JulianKniephoff:
Probably to [`@opencast/eslint-config-ts-react`](https://www.npmjs.com/package/@opencast/eslint-config-ts-react).

This would increase consistency between our frontend projects, and also let us get rid of a number of symptoms of us using what is basically unmaintained code:

* We currently use `eslint-config-react-app`, which is unmaintained.
* That holds back our ESLint version at the 8 release branch.
* It also forces the `@eslint/eslintrc` compatibility package on us.
* It also makes us depend on `@babel/plugin-proposal-private-property-in-object` because of a bug in a transitive dependency, which will never be fixed.
* It holds back our version of `@typescript-eslint/typescript-estree` at a version that is not compatible with the TypeScript version we use. This is probably fine (for now), but generates a fat warning during linting.

Arnei:
I for one have no attachment whatsoever to `eslint-config-react-app` and am all in favor of using our own.

Issue #712, opened 2024-06-19
Live linting/type checking


JulianKniephoff:
Linting and type checking currently only happens when creating a build. This also happens in CI, so we will catch errors eventually, but it would be a better DX if we tightened that feedback loop.

Unfortunately, Vite seems to think it idiomatic to keep that stuff out of the build system, so it's questionable how we want to implement it. This position is supported by `vite-plugin-eslint` being all but dead, basically. There is `vite-plugin-checker`, which Vite also mentions on their website, but I haven't looked into it, yet.

Of course we could also provide IDE integration (if todays IDEs aren't smart enough to pick up on the ESLint config already anyway) or custom scripts as npm-scripts or like we do in Tobira. 🤔

PR #714, opened 2024-06-19
"Restore" ESLint support

JulianKniephoff:
I removed our leftover config in #700 with a promise to bring ESLint back.
Well this is it, for now!

This introduces an ESLint config file in the new flat file format,
which will be required for ESLint `>= 9`. It is already supported in version 8, though,
which we are still stuck on because of the (unmaintained) config we are/were using.
I briefly tried to port this project over to `@opencast/eslint-config-ts-react`,
but the code said "no," basically. ;P

This has a few unfortunate consequences. See #713. I plan to address these in due time, though.

Note that this does **not** introduce automatic/live checking of the code
during development. I only added a call to the linter to the build command,
in turn making CI fail when there are warnings.
I created #712 to track the live thing if we want that.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,0 +1,9 @@
+import { FlatCompat } from "@eslint/eslintrc";
+
+const config = [
+    { ignores: ["build"] },
+    ...new FlatCompat({ baseDirectory: import.meta.dirname })
+        .extends("eslint-config-react-app"),
+];
+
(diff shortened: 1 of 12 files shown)


Arnei (approved):
Achieves what it sets out to do. I look forward to seeing the unfortunate consequences fixed as well

PR #715, opened 2024-06-19
Fix type error for missing organization property

lkiesow:
If an organization property does not exist, the admin interface completely breaks in some cases. This seems to happen on admin-interface.opencast.org right now:

```
Uncaught TypeError: u.toLowerCase is not a function
    qr Footer.tsx:25
    React 8
    D scheduler.production.min.js:13
    $ scheduler.production.min.js:14
Footer.tsx:25:80
```

What's interesting is that this only seems to happen with the compiled code. Starting the development server, calling `toLowerCase()` just works.

In any case, this patch fixes the issue by providing the default value first, so that `toLowerCase()` is always called on a string.

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -22,7 +22,7 @@ const Footer: React.FC = () => {
 	const lastModified = user?.ocVersion?.['last-modified']
 		? new Date(user.ocVersion['last-modified']).toISOString().substring(0, 10)
 		: 'unknown';
-	const aboutEnabled = orgProperties['org.opencastproject.admin.display_about']?.toLowerCase() === 'true';
+	const aboutEnabled = (orgProperties['org.opencastproject.admin.display_about'] || 'false').toLowerCase() === 'true';
 
 	return (
 		<footer id="main-footer">
--- a/src/components/events/partials/modals/SeriesDetails.tsx
+++ b/src/components/events/partials/modals/SeriesDetails.tsx
@@ -62,7 +62,7 @@ const SeriesDetails = ({
 
 	const user = useAppSelector(state => getUserInformation(state));
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
-	const themesEnabled = (orgProperties['admin.themes.enabled']?.toLowerCase() || 'true') === 'true';
+	const themesEnabled = (orgProperties['admin.themes.enabled'] || 'true').toLowerCase() === 'true';
 
 	// information about each tab
 	const tabs = [
--- a/src/components/events/partials/wizards/NewSeriesWizard.tsx
+++ b/src/components/events/partials/wizards/NewSeriesWizard.tsx
@@ -35,7 +35,7 @@ const NewSeriesWizard: React.FC<{
 	const user = useAppSelector(state => getUserInformation(state));
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
 
-	const themesEnabled = (orgProperties['admin.themes.enabled']?.toLowerCase() || 'true') === 'true';
+	const themesEnabled = (orgProperties['admin.themes.enabled'] || 'true').toLowerCase() === 'true';
 
 	const initialValues = getInitialValues(metadataFields, extendedMetadata, user);
 
--- a/src/components/shared/MainNav.tsx
+++ b/src/components/shared/MainNav.tsx
@@ -74,8 +74,8 @@ const MainNav = ({
 	const user = useAppSelector(state => getUserInformation(state));
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
 
-	const statisticsEnabled = orgProperties['admin.statistics.enabled']?.toLowerCase() === 'true';
-	const themesEnabled = (orgProperties['admin.themes.enabled']?.toLowerCase() || 'true') === 'true';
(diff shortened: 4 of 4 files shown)


Arnei (approved):

src/components/Footer.tsx
```diff
@@ -22,7 +22,7 @@ const Footer: React.FC = () => {
 	const lastModified = user?.ocVersion?.['last-modified']
 		? new Date(user.ocVersion['last-modified']).toISOString().substring(0, 10)
 		: 'unknown';
-	const aboutEnabled = orgProperties['org.opencastproject.admin.display_about']?.toLowerCase() === 'true';
+	const aboutEnabled = (orgProperties['org.opencastproject.admin.display_about'] || 'false').toLowerCase() === 'true';
```
JulianKniephoff, line 25:
```suggestion
	const aboutEnabled = (orgProperties['org.opencastproject.admin.display_about'] ?? 'false').toLowerCase() === 'true';
```

src/components/events/partials/modals/SeriesDetails.tsx
```diff
@@ -62,7 +62,7 @@ const SeriesDetails = ({
 
 	const user = useAppSelector(state => getUserInformation(state));
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
-	const themesEnabled = (orgProperties['admin.themes.enabled']?.toLowerCase() || 'true') === 'true';
+	const themesEnabled = (orgProperties['admin.themes.enabled'] || 'true').toLowerCase() === 'true';
```
JulianKniephoff, line 65:
```suggestion
	const themesEnabled = (orgProperties['admin.themes.enabled'] ?? 'true').toLowerCase() === 'true';
```

src/components/events/partials/wizards/NewSeriesWizard.tsx
```diff
@@ -35,7 +35,7 @@ const NewSeriesWizard: React.FC<{
 	const user = useAppSelector(state => getUserInformation(state));
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
 
-	const themesEnabled = (orgProperties['admin.themes.enabled']?.toLowerCase() || 'true') === 'true';
+	const themesEnabled = (orgProperties['admin.themes.enabled'] || 'true').toLowerCase() === 'true';
```
JulianKniephoff, line 38:
```suggestion
	const themesEnabled = (orgProperties['admin.themes.enabled'] ?? 'true').toLowerCase() === 'true';
```

src/components/shared/MainNav.tsx
```diff
@@ -74,8 +74,8 @@ const MainNav = ({
 	const user = useAppSelector(state => getUserInformation(state));
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
 
-	const statisticsEnabled = orgProperties['admin.statistics.enabled']?.toLowerCase() === 'true';
-	const themesEnabled = (orgProperties['admin.themes.enabled']?.toLowerCase() || 'true') === 'true';
+	const statisticsEnabled = (orgProperties['admin.statistics.enabled'] || 'false').toLowerCase() === 'true';
+	const themesEnabled = (orgProperties['admin.themes.enabled'] || 'true').toLowerCase() === 'true';
```
JulianKniephoff, line 78:
```suggestion
	const statisticsEnabled = (orgProperties['admin.statistics.enabled'] ?? 'false').toLowerCase() === 'true';
	const themesEnabled = (orgProperties['admin.themes.enabled'] ?? 'true').toLowerCase() === 'true';
```

JulianKniephoff (changes requested):

src/components/Footer.tsx
lkiesow: This would only have an effect if the default value is `true`. Since the default value here is `false` this should behave identical, shouldn't it?

I mean, the difference is essentially just what happens if the key exists but is empty:

```ts
('' || 'false) => 'false'
('' ?? 'false') => ''

('false' === 'true') => false
('' === 'true') => false
```

If the default value is `'true'`, the effect would differ and the empty string would be considered a valid value equal to `false`.  The old code and old behavior in the case `…` was set to an empty string was:

https://github.com/opencast/opencast-admin-interface/blob/e6cb18f8f4efc0054474ba5cab20898ee2180642/src/components/events/partials/modals/SeriesDetails.tsx#L65

This means that the empty string was treated as no value. Your code would change that behavior. This could cause confusion since an existing configuration may change in meaning. Technically, that shouldn't happen within a minor version of Opencast. If we change that, we should also make sure that's happening consistently in all the code.

Long story short: If there is no good reason to change the behavior, so far you didn't provide any reasoning for the change, I suggest leaving it as is.

JulianKniephoff: That's fair, I was a bit quick with these suggestions.

For the record: I don't like using `||` for this purpose, and I also think it's weird that an empty string is handled differently from `" "` for example. (Or from `"foobar"` for that matter ...). But I guess that's outside the scope of this PR. 🤷‍♀️

JulianKniephoff (commented):
Chaning my review to a "comment;" see https://github.com/opencast/opencast-admin-interface/pull/715#discussion_r1651199936

PR #718, opened 2024-06-20
fix a11y issues in header nav

### LINKED ISSUE #601: Language selector is not accessible with the keyboard
You cannot set the language using the keyboard. The element cannot be accessed using the keyboard. If the drop-down menu is open, a language can be selected.

### LINKED ISSUE #603: Elements in the header are not recognizable in case of language output
The Opencast logo, which is supposed to be the home button, is read out as “link Opencast logo”, so what the link does is not recognizable.
The media module and studio buttons are only read out as “link”.

### LINKED ISSUE #605: Some elements in the header cannot be reached with the keyboard
- System warnings and notifications
- Help
- User menu

rlucke:
use button tag instead of div
reorder tooltip
add some css changes to keep layout

fixes #605 
fixes #601 
fixes #603

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -180,8 +180,8 @@ const Header = ({
                     otherwise the app crashes */}
 					{!!orgProperties &&
 						!!orgProperties["org.opencastproject.admin.mediamodule.url"] && (
-							<Tooltip  title={t("MEDIAMODULE")}>
-								<div className="nav-dd">
+							<div className="nav-dd">
+								<Tooltip  title={t("MEDIAMODULE")}>
 									<a
 										href={
 											orgProperties["org.opencastproject.admin.mediamodule.url"]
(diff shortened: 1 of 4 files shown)


Arnei (approved):
Can navigate to all buttons in the header via tabbing now, and layout is kept. Nice.

Arnei:
As a note, adding translations in PR has no effect, as crowdin can and will override any changes not made directly to the "source" file (lang-en_US.json). If you would like to help with translations, please request access to our crowdin project at [crowdin.com/project/opencast-admin-interface](https://crowdin.com/project/opencast-admin-interface).

Arnei:
Also this is probably obvious, but I do not understand how this addresses #603. Could you explain?

rlucke:
> Also this is probably obvious, but I do not understand how this addresses #603. Could you explain?

The correct position of the ToolTip in the template enables the screenreader to say more than just "link".

PR #719, opened 2024-06-20
Allow admins to save ACLs with no write roles

### LINKED ISSUE #717: Allow admins to save ACLs with no write roles
As an admin, I couldn't restore an ACL to the previous *valid* state, since the interface complains about no write role being set. This doesn't matter for admins since they always have access. There is no need for them to set additional roles.

Screenshot from 2024-06-19 18-04-43

dennis531:
Closes #717

### DIFF
--- a/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
+++ b/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
@@ -183,7 +183,7 @@ const ResourceDetailsAccessPolicyTab = ({
 		let allRulesValid = true;
 
 		values.policies.forEach((policy) => {
-			if (policy.read && policy.write || user.isAdmin) {
+			if ((policy.read && policy.write) || user.isAdmin) {
 				roleWithFullRightsExists = true;
 			}
 
--- a/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
+++ b/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
@@ -177,13 +177,13 @@ const ResourceDetailsAccessPolicyTab = ({
 
 	/* checks validity of the policies
 	 * each policy needs a role and at least one of: read-rights, write-rights, additional action
-	 * there needs to be at least one role, which has both read and write rights */
+	 * if not admin, there needs to be at least one role, which has both read and write rights */
 	const validatePolicies = (values: { policies: TransformedAcl[] }) => {
 		let roleWithFullRightsExists = false;
 		let allRulesValid = true;
 
 		values.policies.forEach((policy) => {
-			if (policy.read && policy.write) {
+			if (policy.read && policy.write || user.isAdmin) {
 				roleWithFullRightsExists = true;
 			}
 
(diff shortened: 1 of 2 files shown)


src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
```diff
@@ -182,15 +182,15 @@ const ResourceDetailsAccessPolicyTab : React.FC <{
...
+	 * if not admin, there needs to be at least one role, which has both read and write rights */
 // @ts-expect-error TS(7006): Parameter 'values' implicitly has an 'any' type.
 	const validatePolicies = (values) => {
 		let roleWithFullRightsExists = false;
 		let allRulesValid = true;
 
 // @ts-expect-error TS(7006): Parameter 'policy' implicitly has an 'any' type.
 		values.policies.forEach((policy) => {
-			if (policy.read && policy.write) {
+			if (policy.read && policy.write || user.isAdmin) {
```
Arnei, line 193: While this probably also works, I suggest using `hasAccess = (role: string, userInfo: UserInfoState)` in `utils/utils.ts` instead, for consistency.

src/slices/aclSlice.ts
```diff
@@ -173,8 +176,8 @@ export const checkAcls = (acls) => async (dispatch: AppDispatch) => {
 			check = false;
 		}
 
-		// check if there is at least one policy with read and write rights
-		if (acls[i].read && acls[i].write) {
+		// if not admin, check if there is at least one policy with read and write rights
+		if (acls[i].read && acls[i].write || user.isAdmin) {
```
Arnei, line 180: Same as above

Arnei (changes requested):

src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
lkiesow: How would you use `hasAccess(…)` here? I don't see how this could be used in a sensible way. Sure, something like this would work:
```ts
if (policy.read && policy.write || hasAccess('', user)) {
```
But that's also just a `user.isAdmin` internally with additional confusing parameters?

Arnei: From what I've seen in the code, checking for admin rights is never done by directly accessing the `isAdmin` property, but always by calling `hasAccess([...])` like so:
```
hasAccess("ROLE_ADMIN", user)
```
Is that because the information in the `isAdmin` property is unreliable? Maybe, I don't actually know. I'm mainly proposing this in the interest of code consistency.

lkiesow: The method `hasAccess(…)` itself uses `isAdmin`:
```ts
src/utils/utils.ts:123:  return !!(userInfo.isAdmin || userInfo.roles.includes(role));
```
and `isAdmin` does exactly the same role check:
```ts
src/slices/userInfoSlice.ts:120:  state.isAdmin = userInfo.roles.includes("ROLE_ADMIN");
```

This means that `hasAccess("ROLE_ADMIN", user)` is essentially

```ts
userInfo.roles.includes("ROLE_ADMIN") || userInfo.roles.includes("ROLE_ADMIN")
```

That works, but doesn't make that much sense. If we have the simple check for an admin and that is exactly what we want, I find it pretty unintuitive to intentionally complicate this by using another wrapper around that.

PR #723, opened 2024-06-20
 Disabled buttons need more contrast #641 

### LINKED ISSUE #641: Disabled buttons need more contrast
The save button has a contrast ratio of 1.59:1 and the remove button has 1.9:1.

I strongly recommand another layout for disabled buttons. Setting opacity to 50% must lead to a poor contrast ratio.

By the way, github does it even worse ;)

rlucke:
add a inactive color for buttons

fixes #641


lkiesow:
I don't think this is a good idea. After your patch you purely rely on color which can also be a problem for people. More than that, the disabled button is not an element which needs or even can be used. I think the low contrast here is intentional. You want to get an inactive element out of the users way after all.

lkiesow:
We discussed this patch at the technical meeting and agreed that it's actually better for accessibility and visual clarity to have the disabled buttons with a deliberate low contrast. This makes it clear to users that these are indeed disabled.

PR #729, opened 2024-06-21
Custom roles in ACL policy

### LINKED ISSUE #658: Custom roles pattern not respected in ACL tab
The custom roles pattern seems to not be respected when trying to add roles that are not in the Opencast database (e.g. LDAP users).

Issue seems similar to opencast/opencast#5016.

Custom Roles Pattern Attempt

dennis531:
# Changes
- Add create option function to dropdown component
- Add button "new custom role" to create a policy with a new custom role

# Alternative 
We could show only one button "new policy" and enable the creation in the dropdown

# Screenshot
grafik

Close #658

### DIFF
--- a/src/components/shared/DropDown.tsx
+++ b/src/components/shared/DropDown.tsx
@@ -8,7 +8,7 @@ import {
 	filterBySearch,
 	formatDropDownOptions,
 } from "../../utils/dropDownUtils";
-import Select, { Theme } from "react-select";
+import Select, { Props } from "react-select";
 import CreatableSelect from "react-select/creatable";
 
 /**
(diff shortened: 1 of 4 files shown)


Arnei (approved):
Code works and looks reasonable.

dennis531:
I prefer to have only one button, as the new create dropdown can also be used to select existing ACLs. In my opinion, this separation would rather confuse users and lead to lower usability. 

We could also use the create dropdown for editing roles to avoid the workaround of removing policies.

dennis531:
> I prefer to have only one button, as the new create dropdown can also be used to select existing ACLs. In my opinion, this separation would rather confuse users and lead to lower usability.
> 
> We could also use the create dropdown for editing roles to avoid the workaround of removing policies.

Done as described.

src/components/shared/DropDown.tsx
```diff
@@ -53,28 +56,40 @@ const DropDown = <T,>({
...
-			options={formatDropDownOptions(
-				filterBySearch(searchText.toLowerCase(), type, options, t),
-				type,
-				required,
-				t
+		<div>
+			{creatable ? (
+				<CreatableSelect
+					{...commonProps}
+					noOptionsMessage={() => "No matching results."}
```
Arnei, line 85: Should be a translatable

Arnei (changes requested):
Works and looks good to me. Really like this over having an extra button. One minor nitpick below.

src/components/shared/DropDown.tsx
dennis531: I added the property to the false component. It should be correct now.

Arnei (approved):

Issue #643, opened 2024-06-05
Some input fields are missing keyboard accessibility/convenience


owi92:
Many input fields in several modals can be "unfocused" with the `esc` and `enter` keys. However this isn't the case for every field since some of them are missing the necessary `onKeyDown` handlers.
This is the case for input fields in the modals for "add user", "add group", "add access policy" and "add theme" (but I'm sure there are others as well).
Fixing this would increase accessibility, convenience and coherence.

Edit: This came up in #619 and was probably also mentioned in serveral a11y related issues.

owi92:
Well so far I've only got a somewhat hacky feeling band-aid solution for this. That would be to add the following prop to every `<Field>` missing the keyboard functionality:
```
onKeyDown={(event: KeyboardEvent) => {
	if (event.key === "Enter" || event.key === "Escape") {
		(document.activeElement as HTMLInputElement).blur();
	}
}}
```
I would honestly prefer another solution but haven't got the time right now, so I'll unassign myself for now.

ziegenberg:
@owi92 Thx for reporting! :)

PR #733, opened 2024-06-24
Add keyboard keys to remove focus on input fields

dennis531:
Adds a keyboard event handler to remove the focus from input elements when the `esc` or `enter` key are pressed. This pull request is based on https://github.com/opencast/opencast-admin-interface/issues/643#issuecomment-2150046215.

Close https://github.com/opencast/opencast-admin-interface/issues/643

### DIFF
--- a/src/components/shared/Field.tsx
+++ b/src/components/shared/Field.tsx
@@ -9,7 +9,7 @@ export const Field = (props: FieldAttributes<any>) => {
     <FormikField
       {...props}
       onKeyDown={(event: KeyboardEvent) => {
-        // Remove focus
+        // Handler for basic html inputs to remove focus, if no custom component is passed
         if (event.key === "Enter" || event.key === "Escape") {
           (event.currentTarget as HTMLInputElement).blur();
         }
--- a/src/components/shared/Field.tsx
+++ b/src/components/shared/Field.tsx
@@ -1,0 +1,19 @@
+import { Field as FormikField } from "formik";
+import { FieldAttributes } from "formik/dist/Field";
+
+/**
+ * Wrapper for the Formik Fields
+ */
+export const Field = (props: FieldAttributes<any>) => {
+  return (
+    <FormikField
+      {...props}
+      onKeyDown={(event: KeyboardEvent) => {
+        // Remove focus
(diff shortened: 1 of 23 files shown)


dennis531:
> I don't understand the point of this pull request. There is already a key handler in RenderField.

This is true, but it doesn't apply to all field components, as not all of them have a component property. [In these cases, a simple HTML input will be used](https://formik.org/docs/api/field), which does not support Enter and ESC keys (at least in Firefox). For example, in the create user dialog, the RenderField or RenderMutlifield is not passed as component prop. 

To avoid redundant code, I created this wrapper that adds this key down handler to all Field components. If a component prop with value such as RenderField is passed, this handler will be ignored in the RenderField component.

src/components/shared/Field.tsx
```diff
@@ -0,0 +1,19 @@
+import { Field as FormikField } from "formik";
+import { FieldAttributes } from "formik/dist/Field";
+
+/**
+ * Wrapper for the Formik Fields
+ */
+export const Field = (props: FieldAttributes<any>) => {
```
lkiesow, line 7: How does this relate to
https://github.com/opencast/opencast-admin-interface/blob/172a3bf55585f8162a8ade14389507b18b7680c0/src/components/shared/wizard/RenderField.tsx#L38-L46

That seems to do something similar. Do they conflict? Should they be merged? Or is this just fine?

lkiesow:
Then, do we still need the code in `RenderField` at all?

dennis531:
> Then, do we still need the code in `RenderField` at all?

Yes, because the handler in `RenderField` manages the specific state of this component, which is different from the basic HTML input logik.

Arnei (approved):
While I can see developers accidentally importing the wrong Field component in the future, this does seem like the most reasonable solution to the given problem. Works and can be merged after the current merge conflicts are fixed imo.

Arnei:
For stuff like the "Add User" oder "Create group" modals, it might be better if the `enter` key worked more like the `tab` key in that pressing it would automatically focus the next field. Not something that needs to be changed by this PR, certainly, I'm just musing here. I also have not thought about all UX or a11y implications yet.

PR #734, opened 2024-06-24
 Modal does not have a focus trap #600 

### LINKED ISSUE #600: Modal does not have a focus trap
It should not be possible to reach elements outside of a modal with the tab key.

rlucke:
add focus trap to modals

fixes #600

### DIFF
--- a/src/components/configuration/partials/wizard/ThemeDetailsModal.tsx
+++ b/src/components/configuration/partials/wizard/ThemeDetailsModal.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { useTranslation } from "react-i18next";
 import ThemeDetails from "./ThemeDetails";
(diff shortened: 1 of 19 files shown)


Arnei (changes requested):
From my testing this seems to generally work, but there still are some problems. Below is an assortment of observations. Testing done with Firefox 121.

I can escape the trap if I press tab before the modal has finished loading, but that should not constitute a common scenario.

I can escape the "Start Task" modal if i mash tab for long enough. Same with "Recording Details" modal. There might be more, I did not go through all of them.

If anything this PR goes to show that we really want our own Modal wrapper component thingy :D

Out of interest, why did you choose to implement the focus trap yourself instead of using a library like focus-trap-react?

Arnei:
This has not seen work in quite some time, closing it in favor of #1339

Issue #625, opened 2024-06-05
Einbettungscode Dialog a11y issues


rlucke:
- dimension buttons cannot be accessed with the keyboard
- screen reader reads out the table underneath the modal
- Focusing after pressing the copy button causes the screen reader to read out the embed code and does not give the feedback that the code has been copied to the clipboard -> feedback should be in a live area
- is there a good reason why the embed code is in a textarea?

dennis531:
> * Focusing after pressing the copy button causes the screen reader to read out the embed code and does not give the feedback that the code has been copied to the clipboard -> feedback should be in a live area

Still open, as I didn't know how to solve this problem without hacks.

Arnei:
All issues should be fixed now

PR #735, opened 2024-06-24
Improve accessablility in embeded code dialog

dennis531:
Changes:
- Add tab navigation through size buttons
- Wrap status message in live area to support screen readers

Related #625


dennis531:
Thank you for the detailed explanation! 

I didn't consider the better solution with the buttons. 

I wasn't sure about the roles, as the [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) states that this role should only be used sparingly. 

I am closing this issue, as it was solved with #736.

PR #737, opened 2024-06-25
Hide type and flavor of add asset items

wsmirnow:
To be consistent, the Add-Asset dialog shouldn't show the type and flavor of the items, same as Add-event dialog.

Before:
image

After:
image

The new event summary page should also hide the asset type and flavor.

Before:
new-event-summary-before

After:
image


Arnei (approved):
Makes sense to me!

Issue #409, opened 2024-05-22
UI seems to ignore some configs


jduehring:
I stumbled upon two config changes that are working on the old admin-ui but seem to have no impact with the new one. 

1. When publishing to an additional publication channel, lets say with the name "LMS-Download", then after editing `/listproviders/publication.channels.properties` to list and name the publication channels the name does not show up and the order is ignored as well.  Instead only "Custom" is shown. 

Old Admin-UI: 
![image](https://github.com/opencast/opencast-admin-interface/assets/32971521/66a96986-d829-4086-8883-093ce44c1ead)

New: 
![image](https://github.com/opencast/opencast-admin-interface/assets/32971521/10171bce-5619-4f3e-b66b-9263dcb50dab)

2. Same for `/listproviders/acl.default.create.properties` : 
Changing `#keep_on_template_switch_role_prefixes=ROLE_USER_
` works with the old UI but not with the new one. 

I haven't yet confirmed if this affects everything in `/listproviders` .

PR #738, opened 2024-06-25
Respect publication channels listprovider more

Arnei:
While we were already getting information on publications from the respective listprovider for the event details, we were not putting all of it to use, for example ordering. Also, the event table did not have access to this information at all. This should fix that.

Helps with #409.

Screenshots, as the effects are rather unnoticeable without making changes to the listprovider.
Bildschirmfoto vom 2024-06-25 12-29-25
Bildschirmfoto vom 2024-06-25 12-29-41

This PR over in the main repository may also be of interest: https://github.com/opencast/opencast/pull/5956


snoesberger:
I have tested this PR on our side with an Annotation Tool publication. Now the name is displayed correctly in the publications (before it was just named "Custom"): 
grafik

gregorydlogan:
@Arnei build correctly fails due to tab character ;)

owi92 (approved):
Code looks reasonable. I haven't done any testing because I don't know my way around the listprovider.
But since Sascha has done so, I think this should be fine to merge.

PR #740, opened 2024-06-25
Hide tooltip on hover

### LINKED ISSUE #722: Tooltip for main menu items covers other main menu item
If I am hovering "Systems", but then decide I want to go to "Capture" after all, I
- Need to know it is there, as the tooltip is hiding the icon
- Need to make the tooltip go away to be able to click on the icon (clicking on the tooltip does the same as clicking on the item it belongs to)

Bildschirmfoto vom 2024-06-20 15-34-00

dennis531:
Close #722

### DIFF
--- a/src/components/shared/MainNav.tsx
+++ b/src/components/shared/MainNav.tsx
@@ -241,14 +241,14 @@ const MainNav = ({
 							{hasAccess("ROLE_UI_NAV_RECORDINGS_VIEW", user) &&
 								(hasAccess("ROLE_UI_EVENTS_VIEW", user) ? (
 									<Link to="/events/events" onClick={() => loadEvents()}>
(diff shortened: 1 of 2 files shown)


Arnei:
Might not be the best way to go about this. In the editor, tooltips explicitly do not disappear even if you click on them, allowing you to mark the contents. Iirc this was done for usability and accessibility reasons which should also be important to us here in the admin ui.

lkiesow (changes requested):
-1 to this solution. Having non-interactive tooltips is bad for accessibility. In particular, trying to use this in combination with screen magnifications is bad. A simple alternative solution for #722 would be `placement='right'`.

dennis531:
Thanks for the suggestion. The main navigation tooltips are placed on the right side. I hope this improves the usability.

lkiesow:
Looks good

PR #749, opened 2024-06-26
Add a very restricted form of `no-tabs` to the ESLint config

JulianKniephoff:
This is in preparation to adopting `@opencast/eslint-config-ts-react`,
which includes `no-tabs`.

We use tabs for indentation in some places, but also for other things.
The indentation tabs are easy to get rid of automatically, if a bit annoying.
It is the **other** instances that are not so easily picked out,
and especially distinguished from valid tab usages (think literals).

So I'm getting these out of the way first.


src/components/shared/TableFilterProfiles.tsx
```diff
@@ -168,7 +168,7 @@ const TableFiltersProfiles = ({
 											<Tooltip title="profile.description">
 												<button
 													onClick={() => chooseFilterProfile(profile.filterMap)}
-                        className="button-li	ke-anchor"
+                        className="button-li ke-anchor"
```
Arnei, line 171: Should there even be a whitespace breaking up the class name here?

Arnei (changes requested):
Looks reasonable and builds. Got one question:

src/components/shared/TableFilterProfiles.tsx
JulianKniephoff: Probably. 👀

Arnei (approved):

PR #750, opened 2024-06-26
Asset-Upload fixed

wsmirnow:
This patch fixes the asset upload.

### DIFF
--- a/src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
@@ -171,6 +171,7 @@
 		"EVENTS_CREATED": "The event has been created",
 		"EVENTS_UPLOAD_STARTED": "The event is being uploaded… {{ progress }}%",
 		"EVENTS_NOT_CREATED": "The event could not be created",
+		"EVENTS_UPDATED": "The event have been successfully updated",
 		"EVENTS_UPDATED_ALL": "All events have been successfully updated",
 		"EVENTS_NOT_UPDATED": "The event could not be saved",
 		"EVENTS_NOT_UPDATED_ID": "The event {{ id }} could not be saved",
--- a/src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
@@ -171,7 +171,7 @@
 		"EVENTS_CREATED": "The event has been created",
 		"EVENTS_UPLOAD_STARTED": "The event is being uploaded… {{ progress }}%",
 		"EVENTS_NOT_CREATED": "The event could not be created",
-		"EVENTS_UPDATED": "The event have been successfully updated",
+		"EVENTS_UPDATED": "The event has been successfully updated",
 		"EVENTS_UPDATED_ALL": "All events have been successfully updated",
 		"EVENTS_NOT_UPDATED": "The event could not be saved",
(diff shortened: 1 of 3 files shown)


src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -171,6 +171,7 @@
 		"EVENTS_CREATED": "The event has been created",
 		"EVENTS_UPLOAD_STARTED": "The event is being uploaded… {{ progress }}%",
 		"EVENTS_NOT_CREATED": "The event could not be created",
+		"EVENTS_UPDATED": "The event have been successfully updated",
```
Arnei, line 174:
```suggestion
		"EVENTS_UPDATED": "The event has been successfully updated",
```

src/slices/eventDetailsSlice.ts
```diff
@@ -1564,24 +1564,47 @@ export const updateAssets = createAsyncThunk('eventDetails/updateAssets', async
...
+		"downloadSourceflavorsExist": string,
+		"download-source-flavors": string,
+	} = {
+		"downloadSourceflavorsExist": String(assetFlavors.length > 0),
+		"download-source-flavors": assetFlavors,
+	};
+
+	let processing: {
+		workflow: string | undefined,
+        configuration: typeof uploadAssetWorkflowConfiguration,
```
Arnei, line 1597: Should be using tabs instead of spaces for indentation.

src/slices/eventDetailsSlice.ts
```diff
@@ -1564,24 +1564,47 @@ export const updateAssets = createAsyncThunk('eventDetails/updateAssets', async
...
+			let uploadAssetFlavor = [option.flavorType, option.flavorSubType].join("/");
+			if (assetFlavors.length > 0) {
+				assetFlavors = [assetFlavors, uploadAssetFlavor].join(",");
+			} else {
+				assetFlavors = uploadAssetFlavor;
+			}
 		}
 	});
 
+	let uploadAssetWorkflowConfiguration: {
```
Arnei, line 1587: Minor stylistic nitpick: It is preferred to use `const` instead of `let` whenever possible.

Arnei (changes requested):
Works and makes sense to me.

Ideally we would not have to make assumptions about magic variables in workflows and have the backend sort it out somehow, but arguably restoring the upload to working order is more important and the assumptions we have to make are not too annoying.

Minor nitpicks below:

src/slices/eventDetailsSlice.ts
```diff
@@ -1564,24 +1564,47 @@ export const updateAssets = createAsyncThunk('eventDetails/updateAssets', async
...
 
 	let assets: {
-		workflow: string | undefined,
 		options: UploadAssetOption[],
 	} = {
-		workflow: uploadAssetWorkflow,
 		options: [],
 	};
 
+	let assetFlavors: string = "";
```
Arnei, line 1572: It is not necessary to type everything explicitly. For example, in this case typescripts type inference would have no problem inferring that this should be a string.

src/slices/eventDetailsSlice.ts
wsmirnow: done

src/slices/eventDetailsSlice.ts
wsmirnow: Thank you for the info. I've updated the PR.

wsmirnow:
Thank you @Arnei. I've updated the PR with your suggestions. Please continue review.

Arnei (approved):

PR #793, opened 2024-07-02
Replace title attribute in footer elements with tooltips

lkiesow:
The footer still used `title` attributes instead of the tooltips we use everywhere else. This patch fixes that.


lkiesow:
Sorry, my bad. Fixing a typing complaint, I actually broke this patch. It's fixed now.

Arnei (approved):

PR #800, opened 2024-07-03
Fix required metadata not being required

### LINKED ISSUE #317: Metadata validation does not respect "required" for any field other than "title"
When creating a new event or series, certain metadata fields can be *required*, meaning they have to have a value before the user can proceed to the next step. Usually, only the `title` field for the common metadata catalog is required, but users can change that in the backend.

Unfortunately, our validation for metadata fields is hardcoded and as such only checks if there is a value for title. Should another field be marked as required, this will be correctly shown in the ui, but will not be respected by validation. For example, `description* may also be marked as required, but the user can proceed to the next step without having entered any description.

To properly solve this, we would need to make our static validation dynamic, based on the metadata catalogs schemas from the backend.

Arnei:
In the backend, metadata catalogs can be configured as needed. One thing that can be configured for each metadata field is if is considered required. A user should not be able to leave a required field empty (or set it to 

empty). Our validation was ignoring said configuration, merely enforcing the title field.

This fixes the issue by dynamically creating a validation schema based on configuration for the existing metadata fields. Although it really feels like Yup does not want me to do this.

Fixes #317.
Includes #798.

Bildschirmfoto vom 2024-07-03 16-25-45
Bildschirmfoto vom 2024-07-03 16-25-56

### DIFF
--- a/src/components/events/partials/wizards/NewEventWizard.tsx
+++ b/src/components/events/partials/wizards/NewEventWizard.tsx
@@ -7,7 +7,7 @@ import NewMetadataPage from "../ModalTabsAndPages/NewMetadataPage";
 import NewAccessPage from "../ModalTabsAndPages/NewAccessPage";
 import NewProcessingPage from "../ModalTabsAndPages/NewProcessingPage";
 import NewSourcePage from "../ModalTabsAndPages/NewSourcePage";
-import { NewEventSchema } from "../../../../utils/validate";
+import { NewEventSchema, MetadataSchema } from "../../../../utils/validate";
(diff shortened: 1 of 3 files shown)


gregorydlogan:
Does what it says on the tin.  Does not address the series side of things, which has the same options (at least in the config file).  Is that something you want to address here since I'm assuming the logic would be more or less copy+paste?

gregorydlogan (approved):

Arnei:
Series side addressed  (I think. Have not looked at this code in a while.)

PR #812, opened 2024-07-04
Fix default value in extended Metadata

Arnei:
The default value for metadata fields in the extended metadata tab was always `false` instead of the intended default value (usually `""`). This patch fixes that.


src/utils/resourceUtils.ts
```diff
@@ -125,7 +125,7 @@ export const getInitialMetadataFieldValues = (
 		for (const metadataCatalog of extendedMetadata) {
 			if (!!metadataCatalog.fields && metadataCatalog.fields.length > 0) {
 				metadataCatalog.fields.forEach((field) => {
-					let value = false;
+					let value: string | string[] | boolean = field.value;
```
lkiesow, line 128: This still doesn't seem correct to me. According to [the docs](https://docs.opencast.org/r/16.x/admin/#configuration/metadata/#part-3-catalog-fields-configuration), you can configure the type of each field (`boolean`, `text`, …). This would convert a field of type `text` with the value `false` to a boolean, wouldn't it? At the same time, types like `date` are kept as string. That seems weird. But maybe I'm missing something?

Arnei: Honestly, I don't quite get it either. I only tried to fix the issue I introduced myself here, so I mainly just put the code back "as it was".  Why we would need to convert strings to booleans, and not other types like dates, and only for the extended metadata is beyond me. On the other hand, the conversion-to-boolean code currently never runs anyway, as we don't accept default values for metadata fields in the first place. So maybe it is better to remove the code entirely and have it behave more like the code for the main metadata catalog?

owi92:
Rebasing this apparently nullified your changes, so I think this can be closed.

PR #813, opened 2024-07-04
Fix unique key prop

Arnei:
No idea if this was actually causing errors, but this fixes the console warnings at least.


lkiesow:
When and where did you get warnings?

Arnei:
Ah, sorry for not mentioning that. When opening the Assets tab in the event details modal. And also when looking at the info for a specific media file in the Assets tab in the event details modal.

lkiesow (approved):

PR #814, opened 2024-07-04
Fix video preview displaying wrong video

### LINKED ISSUE #466: Admin interface displays media preview from different events
It is very good to have a preview in the media details:

Bildschirmfoto 2024-05-25 um 09 33 08

But if you look at the preview from another video, there is the preview from the last video:

Bildschirmfoto 2024-05-25 um 09 33 42

Arnei:
The video preview player in the event assets details would try to display the media from the assets previously visited, which would lead to the player displaying no or the wrong video. This fixes that by ensuring that the player reloads when the url changes.

Fixes #466.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/EventDetailsAssetMediaDetails.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/EventDetailsAssetMediaDetails.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useRef } from "react";
 import Notifications from "../../../shared/Notifications";
 import {
 	getAssetMediaDetails,
@@ -24,6 +24,14 @@ const EventDetailsAssetMediaDetails = ({
 
 	const media = useAppSelector(state => getAssetMediaDetails(state));
 	const isFetching = useAppSelector(state => isFetchingAssetMediaDetails(state));
+
+	const videoRef = useRef<HTMLVideoElement>(null);
+
+	// Make sure to reload the video player when the url changes, React will not do that for us
+	// This is necessary for proper rerendering when switching between assets
+	useEffect(() => {
+		videoRef.current?.load();
+	}, [media.url]);
 
 	return (
 		<div className="modal-content">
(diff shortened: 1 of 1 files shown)


src/components/events/partials/ModalTabsAndPages/EventDetailsAssetMediaDetails.tsx
```diff
@@ -366,8 +374,8 @@ const EventDetailsAssetMediaDetails = ({
 								{/* video player */}
 								<div className="video-player">
 									<div>
-										<video id="player" controls>
-											<source src={media.url} type="video/mp4" />
+										<video ref={videoRef} id="player" controls>
```
lkiesow, line 377: This also displays for subtitle or audio tracks. Maybe we can do something like:
- if `media.has_video` → `<video …`
- else if `media.has_audio` → `<audio …`
- else hide this section

lkiesow (changes requested):
Looks good in general. I'm not sure if the mime type is needed. Testing this locally, videos do not work for me in either Firefox or Chromium. But that may also be caused by them being loaded from develop directly?

src/components/events/partials/ModalTabsAndPages/EventDetailsAssetMediaDetails.tsx
```diff
@@ -366,8 +374,8 @@ const EventDetailsAssetMediaDetails = ({
 								{/* video player */}
 								<div className="video-player">
 									<div>
-										<video id="player" controls>
-											<source src={media.url} type="video/mp4" />
```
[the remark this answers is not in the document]
Arnei: Good catch, thanks.

src/components/events/partials/ModalTabsAndPages/EventDetailsAssetMediaDetails.tsx
Arnei: Reasonable, but I would leave that for a different PR. This is just about loading the correct media file at all.

PR #816, opened 2024-07-08
Fix up registration modal

Arnei:
This contains various changes to the adopter registration modal that should bring it in line with how it used to work and just generally make it look a bit better. For a list of changes see the commit messages.

### DIFF
--- a/src/components/shared/RegistrationModal.tsx
+++ b/src/components/shared/RegistrationModal.tsx
@@ -652,13 +652,16 @@ const RegistrationModal = ({
 										</div>
 										<br />
 
-										<p>{t("ADOPTER_REGISTRATION.MODAL.SUMMARY_STATE.STATS_HEADER")}</p>
-										{formik.values.allowsStatistics &&
+										{formik.values.allowsStatistics ?
+										<>
+											<p>{t("ADOPTER_REGISTRATION.MODAL.SUMMARY_STATE.STATS_HEADER")}</p>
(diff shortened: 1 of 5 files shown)


lkiesow:
Not sure if it's part of what you intended to fix with this pull request, but I noticed that one major issue is that the statistical information are not displayed at all:

Screenshot from 2024-07-09 22-06-05

This is how this looked in the old interface:

Screenshot from 2024-07-09 22-06-56

Arnei:
Thanks for the info, did not notice statistical information is displayed here. Will look into it.

lkiesow:
While there is now a section “statistical information”, that section is just empty. That doesn't seem right?
Screenshot from 2024-07-18 23-28-55

gregorydlogan:
The reason Lars isn't seeing the statistical data is because he told Opencast not to send it.  If you turn it on you see the data as expected.  The bug here is that the whole UI component (Statistical Information, the textarea) should disappear if stats aren't being generated.  Probably.  It's still going to send the empty json object, but it won't contain data.

lkiesow:
> The reason Lars isn't seeing the statistical data is because he told Opencast not to send it. If you turn it on you see the data as expected.

@gregorydlogan, that's an incorrect assumption. I've selected to send user statistics, but this doesn't work unless you selected it in a previous registration attempt. I've filed the issue as #858. I've also added a video demoing the problem.

PR #817, opened 2024-07-08
Show workflow operations when clicking on status in events table

### LINKED ISSUE #391: Clicking on status in event table should open workflow operations
If an event fails, what basically every admin does is to click
- event details
- workflows
- select latest workflow details
- scroll down
- click on operations
- find the failed operation

It would be so much easier if you could just click on the status to always open the event details with the operations of the latest workflows. In fact, this would be helpful not only for `Failed` but also for `Running` and `Finished`.

But then you lose the option to quickly filter by status you say? We already have that via the counter if need be, but filtering for something like `Finished` isn't all that helpful anyway. So, I'm quite happy if that gets replaced with something useful.

dennis531:
Open the workflow operations of the last workflow.

Changes:
- Move navigation logic of event details modal to the redux store to allow modals to be opened from different locations.
- When clicking on status in the events table, open the operations of the last workflow. If no workflow exists in the event, show the workflow overview.

Close #391

### DIFF
--- a/src/components/events/partials/EventsStatusCell.tsx
+++ b/src/components/events/partials/EventsStatusCell.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { useTranslation } from "react-i18next";
 import { useAppDispatch } from "../../../store";
 import { Tooltip } from "../../shared/Tooltip";
-import { EventDetailsPage } from "./modals/EventDetails";
 import { Event } from "../../../slices/eventSlice";
 import {
 	fetchWorkflowDetails,
@@ -10,6 +9,8 @@ import {
 	fetchWorkflows,
 	openModal
 } from "../../../slices/eventDetailsSlice";
+import { EventDetailsPage } from "./modals/EventDetails";
+import { hasScheduledStatus } from "../../../utils/eventDetailsUtils";
 
 /**
  * This component renders the status cells of events in the table view
@@ -22,11 +23,16 @@ const EventsStatusCell = ({
 	const { t } = useTranslation();
 	const dispatch = useAppDispatch();
 
-	const openWorkflowOperations = () => {
+	const openStatusModal = () => {
+		// Open scheduling modal for scheduled and recording events
+		if (hasScheduledStatus(row)) {
+			return dispatch(openModal(EventDetailsPage.Scheduling, row));
+		}
+
 		dispatch(fetchWorkflows(row.id)).unwrap()
 			.then(async (workflows) => {
+				// Open workflow overview modal if no workflows available
 				if (!workflows.entries.length) {
-					// Open workflow overview modal
 					return dispatch(openModal(EventDetailsPage.Workflow, row));
 				}
 
@@ -45,7 +51,7 @@ const EventsStatusCell = ({
 		<Tooltip title={t("EVENTS.EVENTS.TABLE.TOOLTIP.STATUS")}>
 			<button
 				className="button-like-anchor crosslink"
-				onClick={() => openWorkflowOperations()}
+				onClick={() => openStatusModal()}
 			>
 				{t(row.displayable_status)}
 			</button>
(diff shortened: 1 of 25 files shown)


lkiesow (changes requested):
I really like this :heart: 
But there are a few minor issues:
- The tooltip still says it will filter events
- I've tested this with running, finished and failed events and it works fine. Clicking on scheduled events does nothing though. Maybe we could open the event details scheduling tab for those?
- I haven't tested with events in state recording. I would probably treat them similar to scheduled events and open the scheduling tab.

src/components/events/partials/EventsStatusCell.tsx
```diff
@@ -20,24 +24,30 @@ const EventsStatusCell = ({
...
+				}
+
+				// Show operations of last workflow
+				const lastWorkflow = workflows.entries[workflows.entries.length-1];
+
+				// Load necessary workflow data
+				await dispatch(fetchWorkflowDetails({ eventId: row.id, workflowId: lastWorkflow.id }));
+				await dispatch(fetchWorkflowOperations({ eventId: row.id, workflowId: lastWorkflow.id }));
+
+				dispatch(openModal(EventDetailsPage.Workflow, row, 'workflow-operations'));
```
Arnei, line 42: Code generally looks reasonable and makes sense to me.

It would be great if we did could move this code (or part of it) into the related event details component somehow, so we can avoid having to know how the event details components works in this unrelated component. Phrased differently, the EventStatusCell should not need to know which GET requests to send in order to populate the workflow details. But I understand if that this might be difficult, so no biggie if it's not possible.

dennis531: I also don't like the fetches at this place. I had tried to remove them, but I found out that nested tabs load the content independently of each other. As a result, the data was not loaded when navigating to the previous navigation item. 

I'll look at it again and try to find a better solution.

dennis531:
> * I've tested this with running, finished and failed events and it works fine. Clicking on scheduled events does nothing though. Maybe we could open the event details scheduling tab for those?
> 
> * I haven't tested with events in state recording. I would probably treat them similar to scheduled events and open the scheduling tab.

There was a bug. In these cases, the workflow overview should be opened. I think the schedule tab would be more useful because these events have no workflows.

src/components/events/partials/EventsStatusCell.tsx
dennis531: I hope the current solution looks better. Now the data is fetched in the modal components again. The main problem is that we need to pass which data (e.g. workflow, error etc.) should be shown in the modal. The implemented function only allows to select an event and a workflow. It still feels a bit hacky.

Arnei: Arguably that is fine? If we want to open a specific workflow tab in the event details from anywhere, and have the, workflow tab load stuff by itself, we will have to pass the workflowId somehow.

src/components/events/partials/modals/EventDetails.tsx
```diff
@@ -74,6 +75,8 @@ const EventDetails = ({
 	const { t } = useTranslation();
 	const dispatch = useAppDispatch();
 
+	const workflowId = useAppSelector(state => getModalWorkflowId(state))!;
```
Arnei, line 78: If we're only fetching the workflowId here to pass it on, the components could just fetch it themselves, no?

Arnei (changes requested):

src/components/events/partials/modals/EventDetails.tsx
dennis531: Yes, I will change it.

Arnei (approved):

lkiesow (changes requested):
I just realized that the tooltips are still incorrect:

Screenshot from 2024-07-16 18-56-29

This does no longer add a filter

dennis531:
> I just realized that the tooltips are still incorrect:

I think your language is set to `en-GB`?

I only changed the language for `en-US`. How should we handle the translations in this case?

lkiesow:
upps, my bad. I had indeed selected `en-GB`

PR #818, opened 2024-07-08
Fix style of saved filter sets

### LINKED ISSUE #809: Extremely poor contrast on stored filter button
Screenshot from 2024-07-03 23-14-15

### LINKED ISSUE #811: Broken alignment of icons in stored filter menu
The menu for stored filters does not have a proper alignment of its elements. It should probably look and act more like a small table.

Screenshot from 2024-07-03 23-16-21

dennis531:
Changes:
- Fix class name
- Use flex box
- Allow arbitrary word breaks

Close #811, close #809

### DIFF
--- a/src/components/shared/TableFilterProfiles.tsx
+++ b/src/components/shared/TableFilterProfiles.tsx
@@ -168,7 +168,7 @@ const TableFiltersProfiles = ({
 											<Tooltip title="profile.description">
 												<button
 													onClick={() => chooseFilterProfile(profile.filterMap)}
-                        className="button-li	ke-anchor"
+                        	className="button-like-anchor"
 												>
 													{profile.name.substr(0, 70)}
 												</button>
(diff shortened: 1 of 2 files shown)


lkiesow (approved):
Way better than before :rocket:

src/components/shared/TableFilterProfiles.tsx
```diff
@@ -168,7 +168,7 @@ const TableFiltersProfiles = ({
 											<Tooltip title="profile.description">
 												<button
 													onClick={() => chooseFilterProfile(profile.filterMap)}
-                        className="button-li	ke-anchor"
+                        	className="button-like-anchor"
```
lkiesow, line 171: Indentation still doesn't seem to be correct here

dennis531: There was a mix of spaces and tabs. Now it should be tabs only. I would prefer to use spaces instead of tabs in general.

lkiesow (approved):

PR #819, opened 2024-07-08
Set lower maximum z-index than tooltip index

### LINKED ISSUE #810: Tooltips hidden behind stored filter menu
The stored filter menu overlays the tooltips for some of its elements:

Screenshot from 2024-07-03 23-14-27

dennis531:
Fixes the tooltip from being overlapped by the stored filter menu. The z-indices of the mui components are documented at https://mui.com/material-ui/customization/z-index/.

Close #810

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -160,7 +160,7 @@ const Header = ({
 				<nav className="header-nav nav-dd-container" id="nav-dd-container">
 					{/* Select language */}
 					<div className="nav-dd lang-dd" id="lang-dd" ref={containerLang}>
-						<Tooltip title={t("LANGUAGE")}>
+						<Tooltip active={!displayMenuLang} title={t("LANGUAGE")}>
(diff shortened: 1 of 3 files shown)


lkiesow (changes requested):
Unfortunately, the tooltips now interact weirdly with the topmost menu:

https://github.com/opencast/opencast-admin-interface/assets/1008395/cc0f5c4c-ff64-451d-8b45-e80b215c6a85

Maybe we can close the tooltip when I click or remove the tooltip from the icon entirely if the menu is open? That would also solve the problem that I now have a tooltip on the menu.

dennis531:
> Maybe we can close the tooltip when I click or remove the tooltip from the icon entirely if the menu is open? That would also solve the problem that I now have a tooltip on the menu.

Done, the tooltip is closed when the menu is open.

lkiesow (approved):

Issue #805, opened 2024-07-03
Selecting a filter should open the filter value selector


lkiesow:
Selecting a filter (A) will show the value selection (B) but not automatically open the selection which means an unnecessary click. The value selection dropdown should open automatically when I select a filter.

![Screenshot from 2024-07-03 23-02-51](https://github.com/opencast/opencast-admin-interface/assets/1008395/ed902e19-276d-44da-8f69-e00bed7c1fd6)

PR #827, opened 2024-07-10
Use react select for table filters

dennis531:
Automatically opens the second drop-down list as suggested in #805


Arnei (changes requested):
Works pretty well and makes the code a lot more readable.

Got a problem with the second filter sometimes not closing when clicking outside of it. This allows me for example to open event details with the second filter still open:
Bildschirmfoto vom 2024-07-11 15-05-45

Arnei:
[a9c6f5f](https://github.com/opencast/opencast-admin-interface/pull/827/commits/a9c6f5f8697a357f1e621bc69a609c24616f3ab7) does fix the issue mentioned above, but now the second filter dropdown will not open when switching between options in the first filter dropdown.

dennis531:
> [a9c6f5f](https://github.com/opencast/opencast-admin-interface/pull/827/commits/a9c6f5f8697a357f1e621bc69a609c24616f3ab7) does fix the issue mentioned above, but now the second filter dropdown will not open when switching between options in the first filter dropdown.

This problem should be fixed, but adds more complexity to the dropdown. Also, the first filter dropdown is opened initially.

Arnei (approved):
Does work without problems now. Does make our Dropdown a bit more complex, but that is to be expected when extending a component with new features.

PR #830, opened 2024-07-12
Port next/previous tab hotkeys

Arnei:
The old admin ui featured two hotkeys that would let you move forward and backward between the different steps in a create modal. This ports them over to our ui.

Unfortunately, they are not that useful currently, as switching tabs almost always results in an element being autofocused that will eat our hotkey inputs (<input>, <select>, etc.). Not sure currently how to get hotkeys to work despite that in a general fashion.


lkiesow:
If they usually not work properly due to autofocussing input fields, then I think we shouldn't add them in the first place. Any reason for why you do want to add them if you don't think they are useful? I would tend to close this?

Arnei:
I believe hotkeys for switching between tabs to be useful. *if* the keyboard focus after switching would actually focus something that would allow continued keyboard navigation.

I'm not getting into resolving this problem anytime soon, so I'm fine with you closing this and maybe transforming it into an issue instead (and linking this PR as a failed attempt in the issue).

lkiesow:
Created an issue. Closing this for now. We can always re-open it if we want to.

PR #831, opened 2024-07-12
Fix height of filters

### LINKED ISSUE #803: Margin between filter selector and selected filters
There is some free space between selected filters. That is good. The filter selector should have the same space, but is running more or less into the filter instead. That's not good. There should be the same amount of space between all elements.

Screenshot from 2024-07-03 22-27-12

### LINKED ISSUE #804: Height difference between filter and filter selector
The filter selector and the selected filters should have the same height. 

Screenshot from 2024-07-03 22-26-55

dennis531:
Basically adds flexbox to elements and removes some useless `<span>`.

Close #804
Close #803

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -359,6 +359,7 @@ input[type="text"].small-search {
 }
 
 div.table-filter {
+  display: flex;
   float: right;
   border: 1px solid #c9d0d3;
   border-radius: 4px;
@@ -369,8 +370,6 @@ div.table-filter {
     display: flex;
     border: 0;
     border-radius: 4px;
-    float: left;
-    background: #fff;
     background-color: #fafafa;
     min-height: 23px;
     min-width: 150px;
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -403,7 +402,9 @@ div.table-filter {
     }
 
     .ng-multi-value {
-      align-self: center;
+      display: flex;
```
lkiesow, line 405: I would also suggest adding a slight `gap` here:
```suggestion
      display: flex;
      gap: 4px;
```

src/styles/main.scss
```diff
@@ -369,8 +370,6 @@ div.table-filter {
     display: flex;
```
lkiesow, line 370: The filters now run into each other:
Screenshot from 2024-07-12 18-08-32


I think some `gap` would make this better:
```suggestion
    display: flex;
    gap: 4px;
```

lkiesow (changes requested):

src/styles/main.scss
lkiesow: This is how the filters look with both `gap` applied:

Screenshot from 2024-07-12 18-19-40

dennis531:
@lkiesow Thanks for these suggestions. It looks much better now.

dennis531:
Close #803

PR #835, opened 2024-07-12
Fix recursion in effect when fetching series acls

### LINKED ISSUE #829: [OC 16.1] Infinite loop of printing warning message when trying to add recording
**Describe the bug**

During event creation, the web client/Admin UI creates an infinite loop when warning the user about not being able to change access rules due to another episode processing in the same series. 
As a result event creation cannot be finished while the other event is still processing. In addition, the browser will have a spike in CPU usage until the event creation dialog is closed.

**To Reproduce**

This behaviour can be reproduced on stable.opencast.org:

1. Create a new series names "Test Series"
2. Upload a new event "Test Event 1" into this series and let the processing start
3. Now while "Test Event 1" is still processing, create another event in "Test Series"
4. Proceed in the event creation dialog up until the "Access policy" tab, where the warning should now be spammed indefinitely

**Expected behavior**

The warning should only be shown once.

**Screenshots**

Bildschirmfoto vom 2024-07-11 12-11-51

**User environment:**

 - Device: PC
 - OS: Debian 12
 - Browser Chrome
 - Version 126.0.6478.126

**Additional context**

This behaviour was observed with the new Admin UI, I haven't tested whether this also happens with the old Admin UI.

dennis531:
Fixes the mentionied bug in #829.

Due to the React.stric mode, the ACLs are loaded twice and the notification is displayed twice, see https://react.dev/reference/react/StrictMode:

image

This problem could not be solved by simply putting `dispatch(removeNotificationWizardForm());` at the beginning of the effect because the fetch is asynchronous. Is this a problem and should we fix it?

Close #829

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx
@@ -68,8 +68,7 @@ const NewAccessPage = ({
 		if (initEventAclWithSeriesAcl && formik.values.isPartOf) {
 			dispatch(fetchSeriesDetailsAcls(formik.values.isPartOf));
 		}
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, [formik.values.isPartOf, initEventAclWithSeriesAcl]);
+	}, [formik.values.isPartOf, initEventAclWithSeriesAcl, dispatch]);
 
 	// If we have to use series ACL, overwrite existing rules
 	useEffect(() => {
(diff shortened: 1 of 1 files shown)


Arnei (changes requested):
Reasonable change and works.

While you are at it, could you remove the eslint disabling comment? Looks like it really should not be there.

dennis531:
> While you are at it, could you remove the eslint disabling comment? Looks like it really should not be there.

Done. I needed to add dispatch to the list of dependencies, otherwise the warning would be triggered.

Arnei (approved):

PR #836, opened 2024-07-12
Improve highlighting of active step

lkiesow:
This patch causes active steps in wizards to be better highlighted.

I always found the slight size difference while using the same colors, font, … a bit less than I would want from an indicator.

https://github.com/user-attachments/assets/753a4580-36d5-4f44-b8b0-d7093329d442

*Old and new style in action*


Arnei (approved):
Looks reasonable to me.

PR #840, opened 2024-07-16
Unify header filter icon sizes

### LINKED ISSUE #802: Filter icon size
The filter icon is significantly smaller than the other icons. They should all be the same size.

Screenshot from 2024-07-03 22-28-17

dennis531:
Requires the flex box from #831 to center vertically.

Screenshot when #831 is merged:
image

Close #802


lkiesow (changes requested):
Rebased this onto `main` which now contains #831 for testing.
This doesn't quite fix the problem, although it's an improvement.
Here I just took a screenshot and moved all icons to the right, so that they are next to each other:

Screenshot from 2024-07-16 18-42-17

- The gray of the dropdown icon is a lighter gray than the icons on the right
- The magnification lens icon is still larger than the other icons
- The magnification lens also seems like a bitmap icon and not a vector image
- Not sure if the `×` is also smaller or if that's just the icon

dennis531:
> * The gray of the dropdown icon is a lighter gray than the icons on the right

The dropdown icon is lighter when it is disabled. If active the icons should have the same color.

> Not sure if the × is also smaller or if that's just the icon

The icon itself is smaller as the icon heights are all the same.

> The magnification lens also seems like a bitmap icon and not a vector image

The png image is replaced by the font awesome search icon. Now the search icon prevents the click event of the underlying text input because it overlaps the text input. I could not find a simple solution for this problem.

lkiesow (approved):

PR #841, opened 2024-07-16
Remove div container around select component

### LINKED ISSUE #834: Style of schedule information broken
This is a regression bug.

The fields of the scheduling information for hour and minute in the add events dialog should be next to each other. This was the case until very recently. Not sure which patch broke this:

Screenshot from 2024-07-12 15-55-20

dennis531:
The div around the select component leads to styling issues like mentioned in #834.

Close #834


lkiesow (approved):
That seems to fix the problem

PR #842, opened 2024-07-16
Media type specific player

lkiesow:
This patch changes the type of player for media assets based on their type so we actually have player that work with the specific types.

- if `media.has_video` → `<video …`
- else if `media.has_audio` → `<audio …`
- else hide player section

This basically implements my suggestion from https://github.com/opencast/opencast-admin-interface/pull/814#discussion_r1671210608


Arnei (approved):
Works and seems reasonable to me!

PR #846, opened 2024-07-19
Open player directly

lkiesow:
By default, opencast only publishes the engage publications. That is why most installations only have one publication and do not need a menu to choose the publication from.

This patch makes the admin interface link the player directly if there is just the engage publication. If more or different publications exist, we fall back to the old behavior.


Arnei (approved):
Works and looks reasonable to me.

PR #848, opened 2024-07-21
Properly set theme default configuration

lkiesow:
Instead of properly setting the default value for `admin.themes.enabled` in the admin interface, we temporarily set it in the Opencast configuration. That's not a great workaround and it has a few problems, but I let us backport the admin interface to `r/15.x` to fix the major problems there.

Not that 15 is our legacy release and we already have a two 16.x versions out, it's time to properly set the default value.


gregorydlogan:
Are you going to do a PR fixing it in the mainline repo as well?  From 16.x:

```
# Enable themes in the admin interface.
# Format: boolean
# Default: true
prop.admin.themes.enabled=false
```
Immediate issues:
- Non-commented default
- Default setting does not match listed default
- Listed default does not match UI default

:D

PR #850, opened 2024-07-22
auto select workflow when there is only one

### LINKED ISSUE #849: Auto-select workflow when there is only one to choose from
When there is only one workflow to choose from, this one should be automatically selected.

image

schuettloeffel-elsa:
Closes #849

When there is only one workflow to choose from, this one should be automatically selected.

image


Arnei (approved):
Works and seems reasonable to me.

PR #855, opened 2024-07-25
Fix undefined access error in empty access policies

### LINKED ISSUE #334: Admin interface displays ACL rules from different events
It seems like the new admin interface will sometimes show ACL rules from different events. This can easily lead to accidents where a user overwrites the ACL set with an incorrect set of rules.

https://github.com/opencast/opencast-admin-interface/assets/1008395/85b96839-29c6-4185-a899-fb15b2280df9

dennis531:
Changes:
- Add check if fetched acls contain an `ace` property
- Hopefully improve code readability

Close #334

### DIFF
--- a/src/slices/eventDetailsSlice.ts
+++ b/src/slices/eventDetailsSlice.ts
@@ -843,25 +843,34 @@ export const fetchAccessPolicies = createAppAsyncThunk('eventDetails/fetchAccess
 	);
 	let accessPolicies = await policyData.data;
 
-	let policies: TransformedAcl[] = [];
-	if (!!accessPolicies.episode_access) {
-		const json = JSON.parse(accessPolicies.episode_access.acl).acl.ace;
-		let newPolicies: { [key: string]: TransformedAcl } = {};
-		let policyRoles: string[] = [];
-		for (let i = 0; i < json.length; i++) {
-			const policy: Ace = json[i];
-			if (!newPolicies[policy.role]) {
-				newPolicies[policy.role] = createPolicy(policy.role);
-				policyRoles.push(policy.role);
-			}
-			if (policy.action === "read" || policy.action === "write") {
-				newPolicies[policy.role][policy.action] = policy.allow;
-			} else if (policy.allow === true) { //|| policy.allow === "true") {
-				newPolicies[policy.role].actions.push(policy.action);
-			}
-		}
-		policies = policyRoles.map((role) => newPolicies[role]);
+  let policies: TransformedAcl[] = [];
+
+	if (!accessPolicies.episode_access) {
+		return policies;
 	}
+
+	const json = JSON.parse(accessPolicies.episode_access.acl).acl.ace;
+	if (json === undefined) {
+		return policies;
+	}
+
+	let newPolicies: { [key: string]: TransformedAcl } = {};
+	let policyRoles: string[] = [];
+
+	for (let i = 0; i < json.length; i++) {
+		const policy: Ace = json[i];
+		if (!newPolicies[policy.role]) {
+			newPolicies[policy.role] = createPolicy(policy.role);
+			policyRoles.push(policy.role);
+		}
+		if (policy.action === "read" || policy.action === "write") {
+			newPolicies[policy.role][policy.action] = policy.allow;
+		} else if (policy.allow === true) { //|| policy.allow === "true") {
+			newPolicies[policy.role].actions.push(policy.action);
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


src/slices/eventDetailsSlice.ts
```diff
@@ -855,25 +855,34 @@ export const fetchAccessPolicies = createAppAsyncThunk('eventDetails/fetchAccess
...
-				newPolicies[policy.role][policy.action] = policy.allow;
-			} else if (policy.allow === true) { //|| policy.allow === "true") {
-				newPolicies[policy.role].actions.push(policy.action);
-			}
+
+	if (!accessPolicies.episode_access) {
+		return policies;
+	}
+
+	const json = JSON.parse(accessPolicies.episode_access.acl).acl.ace;
```
lkiesow, line 863: Just to be safe
```suggestion
	const json = JSON.parse(accessPolicies.episode_access.acl).acl?.ace;
```

src/slices/eventDetailsSlice.ts
```diff
@@ -855,25 +855,34 @@ export const fetchAccessPolicies = createAppAsyncThunk('eventDetails/fetchAccess
...
+
+	for (let i = 0; i < json.length; i++) {
+		const policy: Ace = json[i];
+		if (!newPolicies[policy.role]) {
+			newPolicies[policy.role] = createPolicy(policy.role);
+			policyRoles.push(policy.role);
+		}
+		if (policy.action === "read" || policy.action === "write") {
+			newPolicies[policy.role][policy.action] = policy.allow;
+		} else if (policy.allow === true) { //|| policy.allow === "true") {
```
lkiesow, line 879: We know `allow` is a `boolean`, don't we?

```suggestion
		} else if (policy.allow) {
```

dennis531: I prefer a strict type check to prevent false truthy values, for example `"false"`, see https://developer.mozilla.org/en-US/docs/Glossary/Truthy.

lkiesow: Isn't the whole idea of Typescript that we have the strict type check? And `allow` is defined as a `boolean`, isn't it? So, this can never have the value `"false"` can it?

dennis531: The policy variable has the type `Ace` which defines `allow` as `boolean`:

https://github.com/opencast/opencast-admin-interface/blob/ace958b0399538849e955f5896a050c143a42b7c/src/slices/aclDetailsSlice.ts#L10-L14

But yes, it can, because Typescript only does a type check at compile time, see https://learning-notes.mistermicheels.com/javascript/typescript/runtime-type-checking/

Arnei: Although if it can be string, we should adjust our type definition to correctly reflect that (`allow: boolean | string`), or fix our backend to always give us a boolean. Since in this case we have full control over the external data (aka the admin ui backend), introducing a validation library is arguably overkill.

lkiesow: Interesting, I thought TypeScript would include some runtime type checks.

> […] or fix our backend to always give us a boolean […]

I think that's already what the backend does. So I think we can still assume this to be a boolean.

But let's assume we aren't sure about the type here, that would mean the code would convert a value like `null`, `undefined`, `"true"` or `"True"` to `false`. This is different from how Opencast's core would handle this: A non-existing `allow` value is assumed to be `true`. A string would cause an error.

In other words, this isn't really a type check, but just setting a default value of `false` for everything but `true`. That means, what this actually does is just to re-define which values are truthy.

If we want an actual type check, we could e.g. add a [`typeof` type guard](https://learning-notes.mistermicheels.com/javascript/typescript/type-guards/#typeof-type-guards) before line 877. Something like:
```js
if (typeof policy?.allow !== "boolean") {
  console.error("Invalid value of allow in acl");
  continue:
}
```

If we want a behavior identical to Opencast's, we could even do

```js
if (policy.allow === undefined) {
  policy.allow = true
}
if (typeof policy?.allow !== "boolean") {
  console.error("Invalid value of allow in acl");
  continue:
}
```

Or we could just assume that the backend provides correct data :D

In the end, I think this is an interesting thing to argue about, but it has little real word relevance. If you ant to modify this, that's good. If you don't, that's fine as well, and we can merge it like this.

What do you think?

dennis531: I think the type checking is not really necessary if we know that the backend provides the correct data. Nevertheless, we should unify the behavior of unset parameters.

Issue #851, opened 2024-07-23
Hidden steps of new series/event modal aren't always skipped


owi92:
The current logic only checks if the next step is hidden, and if so, skips that step. It doesn't take into account that two or more steps in a row might be hidden.
This can lead to situations where a hidden step (i.e. the tobira step) is not visible in the navigation header but will still be included when creating a new series or event.
<img width="809" alt="Bildschirmfoto 2024-07-23 um 16 16 31" src="https://github.com/user-attachments/assets/b8ca325d-0158-4639-b77b-5cc11b33db4a">

Edit: The above tab opens after completing the "Access policy" step.

PR #856, opened 2024-07-30
Make next/previous buttons a bit more robust

Arnei:
Makes the nextPage/previousPage functions for the create event/series modals a bit more robust, by removing assumptions about how many hidden pages in a row are specified in the steps array.


src/components/events/partials/wizards/NewEventWizard.tsx
```diff
@@ -108,20 +108,24 @@ const NewEventWizard: React.FC<{
...
 		setPageCompleted(updatedPageCompleted);
 
-		if (steps[page + 1].hidden) {
-			setPage(page + 2);
-		} else {
-			setPage(page + 1);
+		let newPage = page;
+		do {
+			newPage = newPage + 1;
+		} while(steps[newPage] && steps[newPage]!.hidden);
```
lkiesow, line 114: This is more out of curiosity and the fact that I didn't know about the null assertion operator `!.` until now. What is the reasoning behind using it here? From what I read, the operator will add a runtime check failing (it would throw an error) that `steps[newPage]` is not `null`. If the assertion fails, I think this would break the UI. That is why you have the check that it can never be `null` as first part of the condition. But then… you don't need the assertion because it is already always guaranteed to pass, isn't it? I'm somewhat struggling to see where it would make sense to use this operator and don't know why it would make sense here.

Arnei: Nah you're right, the `!` operator is a left-over that is not needed here anymore. I've removed it.

The null assertion operator is to be used in cases where you successfully determined that your variable is not `null | undefined`, but typescript fails to infer on that. You can then use `!` to shut typescript up.
But more often than not using `!` points to a problem in your code that you should fix instead of using `!`.

owi92 (approved):
This works, ~~but when the first or last page are hidden, the next/previous button is still visible. I suppose the first page would practically never be hidden, but the last page might. Still, not really directly connected to this PR~~ (which is already an improvement and fixes what it says), so should be ok to merge.
Edit: The last step also won't ever be hidden, so disregard that part.

Btw, this closes https://github.com/opencast/opencast-admin-interface/issues/851. Not sure if it does that automatically upon merge while only stated in my comment, so feel free to add that to your description.

PR #857, opened 2024-07-30
Fix missing `agreedToPolicy` parameter when add adopter

dennis531:
No parameter `agreedToPolicy` is sent to Opencast, which means that the new policy will not be accepted in Opencast, but the old one will be. Thus, the data will not be sent to register.opencast.org. This pull request adds this missing parameter.

Related lines in Opencast:
- https://github.com/opencast/opencast/blob/904bf85b0c8085b7f157ebf38e1fc3ab2252f0c9/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java#L218

- https://github.com/opencast/opencast/blob/904bf85b0c8085b7f157ebf38e1fc3ab2252f0c9/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java#L164-L166

### DIFF
--- a/src/utils/adopterRegistrationUtils.ts
+++ b/src/utils/adopterRegistrationUtils.ts
@@ -29,6 +29,7 @@ export type Registration = {
 	street: string,
 	streetNo: string,
 	email: string,
+	agreedToPolicy: boolean,
 }
 
 // post request for adopter information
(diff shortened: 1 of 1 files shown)


lkiesow:
This seems to fix registration. Tried this and data end up on register.opencast.org

PR #875, opened 2024-08-07
Don't show episode role ids for events

Arnei:
Supposed to accompany https://github.com/opencast/opencast/pull/5056.

Due to unfortunate technical limitations, 5056 (linked above) requires episode role ids to be added to the index for events.  This is nothing users should concern themselves with, so this filters those roles out right after we get them.


Arnei:
This has become unnecessary, therefore closing.

PR #313, opened 2024-03-21
Port Tobira integration

Arnei:
Resolves #311

The old admin ui had an integration for tobira in the series details and new series modal. This ports the integration to our new ui. The user experience could arguably be improved upon, but this should work and be familiar to users who already used the integration in the old admin ui.

PR #878, opened 2024-08-12
Add additional features for Tobira integration

### LINKED ISSUE #311: Tobira Integration
https://github.com/opencast/opencast/pull/4131 added Tobira integration to the old admin UI. This should be forward-ported.

There might also be some future requirements that can be tracked here.

See also:

* https://github.com/elan-ev/tobira/issues/718
* https://github.com/elan-ev/tobira/issues/719

owi92:
This adds most of the additional requirements in #311.
Needs https://github.com/elan-ev/tobira/pull/1225 and https://github.com/opencast/opencast/pull/6091 to work correctly.

Key changes/features:
- The event details modal now has a Tobira tab which provides a direct link and list any pages in Tobira that feature the event: 
    Bildschirmfoto 2024-08-12 um 15 02 32

- The Tobira tab of the series details modal got the addition of an "edit path" button that lets admins edit paths of series that are already mounted in Tobira (with the condition that the page where the series is mounted does not contain any other blocks)
    Bildschirmfoto 2024-08-12 um 15 05 38Bildschirmfoto 2024-08-12 um 15 38 48

- Selectable pages for adding and moving (i.e. updating the path of) series now feature an additional checkbox that shows which page is currently selected (and selectable at all)
    Bildschirmfoto 2024-08-12 um 15 07 09

Notable omissions/deviations:
- @oas777 suggested to also add editing host-page paths of events. While technically possible, we don't allow choosing a host-page for events via the admin UI in the first place, so that would need to be added first. This needs some discussing and is currently out of scope for this PR.
- @dagraf suggested adding a green checkmark when creating new subpages for hosting a series. Clicking that would then confirm the creation and disable editing the title and path. I decided against that since I already added the checkbox as per David's other suggestion, which serves a similar purpose. I feel like adding yet another checkmark to that is unnecessary and may wrongly suggest that the path will already be created when that is clicked.

If these omissions are dealbreakers, we should discuss them in an upcoming meeting.

**Note: This PR is based on #313 which adds the Tobira integration. That should definitely be merged prior to this one!** _(it's only been open for like 4 months..)_
I'll mark this as a draft until #313 is merged.
Closes #311


owi92:
I tried to make the Tobira tabs a little less "matryoschka"-like, as suggested in todays technical Meeting. This does however mean that these tabs won't follow the design examples set by other tabs.

They are still built like russian dolls (I didn't come up with that analogy but it's fitting).
Though I think that should be fixed in another PR.

Some updated screenshots:
Bildschirmfoto 2024-08-13 um 19 18 57
Bildschirmfoto 2024-08-13 um 19 19 20
Bildschirmfoto 2024-08-13 um 19 19 54

Arnei (changes requested):
Found a page crasher:
- Go to the tobira tab in either the series details or when creating a new series.
- Click on the "+ add subpage" button.
- Try to switch to a different page (e.g. from a subpage to the homepage)
- Observe blank page and "Uncaught Error: Maximum update depth exceeded." in the browser console.

Arnei (changes requested):
Bildschirmfoto vom 2024-10-29 15-04-49

This is an event without a series, so "Mount series" does not make sense here? Should it say "Mount event"? Also when clicking on the "Mount series" button and trying to mount, the mount post request seems to fail.

owi92:
> Found a page crasher (...)

Hm I can't reproduce that. Is this happening every time you try that, even with a freshly started Admin UI, or are there some actions done before that might possibly provoke this?

> This is an event without a series, so "Mount series" does not make sense here? Should it say "Mount event"? Also when clicking on the "Mount series" button and trying to mount, the mount post request seems to fail.

Yeah that shouldn't be there. Mounting events via Admin UI is not a thing (yet).

src/slices/shared/tobiraErrors.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// All Tobira fetch endpoints are using the same return codes which are handled in the same way.
+// So this should be fine to reuse. In the case of `503` I decided not to log the complete error,
+// as it isn't really helpful and might not even be considered as one (see TODO comment below).
+export const handleTobiraError = (response: AxiosErrorResponse, dispatch: AppDispatch) => {
+    const data = response.response;
+    console.info(response.message);
+
+    if (data.status === 503) {
+        // TODO: figure out what to do:
```
Arnei, line 30: If I understand this correctly, there is currently nothing we can do about this until Opencast decides to return less ambiguous data? So it's fine to leave this in?

Arnei (changes requested):
Nitpicks below. Looks good generally.

Arnei:
> Hm I can't reproduce that. Is this happening every time you try that, even with a freshly started Admin UI, or are there some actions done before that might possibly provoke this?

Was happening consistently for me.

Arnei:
Quick screencap: [Bildschirmaufzeichnung vom 2024-10-29 17-25-18.webm](https://github.com/user-attachments/assets/65e8ce0d-8545-4337-8886-56beb5610363)

Arnei:
> I have added a commit that might fix the crashing error you reported, but I can't say for sure.

Now crashes with a different error, see below. When creating a news series, this error also occurs immediately upon switching to the tobira tab.

```
Uncaught TypeError: formik.values.breadcrumbs[(formik.values.breadcrumbs.length - 1)] is undefined
    NewTobiraPage NewTobiraPage.tsx:148
    React 15
    nextPage NewSeriesWizard.tsx:107
    onClick NewAccessPage.tsx:364
    React 15
NewTobiraPage.tsx:148:6
```

Arnei:
Latest and original error are fixed. With this, all my change requests are adressed and this is good to go from my side.

Arnei (approved):

PR #886, opened 2024-08-26
Properly type `createPolicy`

JulianKniephoff:


src/utils/resourceUtils.ts
```diff
@@ -588,12 +588,11 @@ export const buildThemeBody = (values) => {
...
 // creates an empty policy with the role from the argument
-// @ts-expect-error TS(7006): Parameter 'role' implicitly has an 'any' type.
-export const createPolicy = (role) => {
+export const createPolicy = (role: string) => {
 	return {
 		role: role,
 		read: false,
 		write: false,
-		actions: [],
+		actions: [] as string[],
```
Arnei, line 596: Looks to me likes this returns `TransformedAcl` from `src/slices/AclDetailsSlice.ts`. If that's the case, would it make more sense to change the return type of `createPolicy` to `TransformedAcl` instead of "manually" typing the array here?

Arnei (changes requested):

src/utils/resourceUtils.ts
JulianKniephoff: Totally fair

Arnei (approved):

PR #888, opened 2024-08-28
Fix checkmarks for multi-valued fields in bulk metadata editing

JulianKniephoff:
The checkmarks represent whether or not the value was touched
in the bulk editing dialog.

For fields that can have multiple values, this was always the case
becasue the arrays backing those fields were compared for identity
instead of equality.


Arnei (approved):
That seems reasonable.

Arnei:
Unrelated, but since you've been looking at this part of the ui, I got a question: Why are the blue checkmarks not read-only? There's not point in clicking on them, or am I missing something?

JulianKniephoff:
Nope, I got nothing. 🤷‍♀️

PR #912, opened 2024-09-12
Persist filters

### LINKED ISSUE #448: Admin UI loose saved filter sets
I like to save filter sets, but with the next login or if i refresh the page they are gone.

Arnei:
Fixes #448.

Saved filters were lost on page reload. Thanks to modernizing redux this is now an easy fix.


snoesberger:
Works fine for me, filter is saved and after a page reload still available in the "Saved filter sets" list.

lkiesow (approved):

PR #914, opened 2024-09-13
Fix new filters not being loaded

### LINKED ISSUE #693: Data seems to be loaded from cache instead of loading from locally running Opencast
When running Opencast (develop) locally and integrating a new event filter the filter is not visible in the Admin-Ui-Interface until I delete my browser cache and cookies.

Arnei:
Fixes #693.

When a new table filter was added to Opencast, it would not show up in the filters dropdown unless the page cache was cleared. This should fix that.

I legit do not understand why the code was the way it was before this commit. If a reviewer could enlighten me why the previous code makes sense that would be great, but personally I see no harm in fetching the respective table filters everytime we load a new table.


Arnei:
Right. At the time I used [this PR](https://github.com/opencast/opencast/pull/6023) to test this. It is now merged, so you would have to recreate it.

There are probably easier ways to test this, for example by modifying the result of the request in `tableFilters/fetchFilters` in tableFilterSlice.ts. I can provide more detailled code changes if you would like.

geichelberger:
The network log is pretty noisy now; I get a lot of `admin-ng/resources/PUBLICATION.CHANNELS.json` requests.

geichelberger (approved):
LGTM

PR #921, opened 2024-09-29
Open event details on the center of the page

### LINKED ISSUE #919: Open Event Details Popup scrolls with page and is displayed at the top of the page (even for event lower down the list)
Open Event Details Popup scrolls with page and is displayed at the top of the page (even for event lower down the list).
The screenshot shows the Event Details after the Open Event Details Icon of Espresso was clicked.  
  
PC, Linux, Firefox 130.0.1 (64-bit),     Opencast 16.x – b0e0131 – built on 2024-09-29 at stableopencast.org:
image  
  
Edit: It also affects other dialogs (i.e. the series dialog and the deletion confirmation dialog).

lkiesow:
This patch updates the position the event details dialog is opened in and make it a fixed position on the display instead opening it at the top of the page. The previous behaviour caused the dialog to open outside the visible area in some cases – like opening the details from the bottom of a long events table – causing a confusing user experience.

This fixes #919

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -130,6 +130,9 @@ a {
 
 .modal {
   top: 30px;
+  position: fixed;
+  max-height: 90vh;
+  overflow: scroll;
   input[readonly] {
     border: none;
   }


marcusm7:
That was fast! :)  
  
Does the patch also affect the other dialogs (series properties, deletion confirmation)?
Will test it next week.

lkiesow:
> Does the patch also affect the other dialogs

Yes, this should change all dialogs

marcusm7:
Tested on  
https://test.admin-interface.opencast.org/921/2024-09-29_18-17-29/  
with  
Mozilla Firefox 131.0  
Chromium 129.0.6668.89  
  
The issue seems to be fixed. 
- The dialog is always visible and does not scroll with the menu.
- The dialog is scrollable when the browser window is smaller than the dialog-popup.
- The behaviour is fixed for all tested dialogs (event details, series details, new event, user details).

lkiesow:
Completely forgot about this. Sorry.
Thanks for testing this.
Merging.

PR #944, opened 2024-10-10
Fix ICLA Check

Arnei:
Copied from https://github.com/opencast/opencast/pull/6238

The tool we use for checking if contributors signed their ICLA is failing to install on CI runs with new Python versions since we just install it system wide. While that may not be good practice on your local machine, this shouldn't matter in the CI since we immediately throw away the machine anyway.


lkiesow:
Makes sense to get this in as soon as possible → just merging this

PR #945, opened 2024-10-10
Fix test containers

lkiesow:
This patch fixes the test containers, which were failing since `git` is now required for building/running the admin interface.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -5,7 +5,7 @@ import viteTsconfigPaths from "vite-tsconfig-paths";
 import preserveDirectives from "rollup-preserve-directives";
 import child from "child_process";
 
-const commitHash = child.execSync("git rev-parse HEAD").toString().trim();
+const commitHash = child.execSync("giat rev-parse HEAD || cat commit || :").toString().trim();
 
 export default defineConfig({
     base: process.env.PUBLIC_URL || "",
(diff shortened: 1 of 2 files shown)


marcusm7:
Tested the docker image with:
docker run --rm -it -p 127.0.0.1:3000:3000 ghcr.io/opencast/opencast-admin-interface:pr-945

The git dependency is found now, but it can't find the git repo, it seems.

```
fatal: not a git repository (or any of the parent directories): .git
failed to load config from /admin-interface/vite.config.ts
error when starting dev server:
Error: Command failed: git rev-parse HEAD
fatal: not a git repository (or any of the parent directories): .git

    at genericNodeError (node:internal/errors:984:15)
...
```

lkiesow:
Should be fixed now

vite.config.ts
```diff
@@ -5,7 +5,7 @@ import viteTsconfigPaths from "vite-tsconfig-paths";
 import preserveDirectives from "rollup-preserve-directives";
 import child from "child_process";
 
-const commitHash = child.execSync("git rev-parse HEAD").toString().trim();
+const commitHash = child.execSync("giat rev-parse HEAD || cat commit || :").toString().trim();
```
mtneug, line 8:
```suggestion
const commitHash = child.execSync("git rev-parse HEAD || cat commit || echo 'unknown'").toString().trim();
```

mtneug (changes requested):
I would have used env variables with default values, but this also works. There is one outstanding typo and I would suggest having some default non-empty string.

JamesUoM:
Though it works it's using the default:

```
/bin/sh: git: not found
cat: can't open 'commit': No such file or directory
```

ziegenberg (approved):

PR #958, opened 2024-10-28
URL-encode `filter` parameter for `/admin-ng`-endpoints

JulianKniephoff:
See https://github.com/opencast/opencast/pull/6111.


JulianKniephoff:
In yesterday's technical meeting I mentioned that there might still be issues with this PR. These should now be addressed, so people can now test, review, and/or merge without worrying about that. :)

PR #961, opened 2024-10-29
Fix "Start Task"

JulianKniephoff:
You currently can't start any tasks in the admin UI. A `git bisect` showed 356a727c3c9e67795b0065a1a706c70a6421e28f by @Arnei  to be the culprit, but I don't really see how.

Anyway, this should fix it.


Arnei (approved):
Thanks for the biscet, I see how I broke this, and this fix thus makes perfect sense. Approved.

PR #979, opened 2024-11-12
Preselect all capture agent inputs when scheduling new event

### LINKED ISSUE #941: Capture agent options should be pre-selected when scheduling a recording
When I schedule a new capture event recording, the Extron SMP options should be pre-selected in the scheduling dialogue:
grafik

Arnei:
Aims to fix #941.

When scheduling a new event, if the selected capture agent has any inputs, the inputs are all preselected per default now. At least one input must remain selected. Input name should be rendered nicer now.

Note that these changes are only for when scheduling a new event (except the input name rendering), because the linked issue only talked about scheduling new events. For now I would leave the discussion on how this should work for existing events to a separate issue/pr.  

For testing this, a capture agent with inputs must be configured in your Opencast. If you do not have a "real" capture agent, a dummy capture agent can still be configured. See the Opencast docs for more info: https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent-input/#setting-input-options

Quick showcase of the PR in action:
[Bildschirmaufzeichnung vom 2024-11-12 11-05-41.webm](https://github.com/user-attachments/assets/f49f9086-f95e-4cd0-8e40-fce43d465e2e)


snoesberger:
I tested this with an Extron SMP (2 inputs, channel A and B) and a pyCA (no input) and it worked as expected for both.

I only have a few minor comments:
- When no input is selected, the next button is greyed out and not clickable, but there is no indication for the user to know what is missing. Can you add some information to indicate that at least one channel needs to be selected? Perhaps a text message like the one that appears when you select an end date that is in the past?
- If the CA has inputs, the inputs field should also have a red star symbol like the other required fields.

Perhaps someone who has CAs from other vendors can have a look and see if this works well for them too?

lkiesow:
I'm wondering what our expectations for capture agents are for this.
Our documentation states that it's up to the agent to interpret whatever value they get.
Which seems to me like no selection would also be a valid option.

In that sense, we should at least test this with all commonly used agents before merging this:

- [ ] Extron SMP
- [ ] Adena Arec
- [ ] Epiphan Perl
- [ ] pyCA
- [ ] Galicaster *←(not sure if anyone is still using that)*

Or was this behavior already present in the old admin interface?

gregorydlogan:
@JamesUoM @NadiaUoM you guys still using Galicaster?

NadiaUoM:
Hi Greg,

Yes, we still use galicaster.

Thanks & Regards,
Nadia Malek.

JamesUoM:
... But we are going to move to pyCA by the next academic year.

James


--
------------------------------------------------------------------------
  James S. Perrin

  Media Technologies Team
  Office One, Kilburn Building,
  The University of Manchester
  Oxford Road, Manchester, M13 9PL

  e: ***@***.***
________________________________

gregorydlogan:
Can you guys take a quick poke and see if this change breaks things for you while you're still using GC?

JamesUoM:
> Can you guys take a quick poke and see if this change breaks things for you while you're still using GC?

Ok, though I doubt we'll still be using GC when we eventually deploy this code ;-)

JamesUoM:
> > Can you guys take a quick poke and see if this change breaks things for you while you're still using GC?
> 
> Ok, though I doubt we'll still be using GC when we eventually deploy this code ;-)

BTW I assume Zaragoza will be using GC

JamesUoM:
This will work fine with GC and matches what we have in the old Admin UI.

NOTE: GC can be configured to "hide" the inputs/tracks from OC and just returns `capture.device.names=defaults`  if a scheduler requests `capture.device.names=defaults` or `capture.device.names=` than all enabled tracks are recorded

gregorydlogan:
This doesn't seem to work for me.  I get the right UI elements, but nothing's selected?
[Screencast from 2025-04-02 16-32-53.webm](https://github.com/user-attachments/assets/4b09b034-7d20-4cdc-87c6-48495820add4)

Arnei:
Might be bugged due to a faulty merge. However, as it is unclear if we even want this feature at all, I'll abstain from working on the code for now.

snoesberger:
I'd still love to have this feature. Btw. the behavior was already like that in the old admin UI.

From an end user's point of view, this improves usability and saves clicks. For Extron devices, the default with no active input is not a valid config because at least one input must be selected to make a usable recording.
And I fear that there will be many support cases because users forget to activate an input and end up with an empty recording.

snoesberger:
The preselection of the inputs doesn't work anymore. I did plan a new event for an Extron SMP, the two inputs Channel A and Channel B were displayed but the checkboxes weren't preselected.

gregorydlogan (changes requested):
Looks like the code underneath changed?

```
/home/greg/opencast/opencast-admin-interface/src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
  627:15  error  'inputDevice' is never reassigned. Use 'const' instead  prefer-const
```

gregorydlogan:
```
/home/greg/opencast/opencast-admin-interface/src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
  626:15  error  'inputDevice' is never reassigned. Use 'const' instead  prefer-const
```

PR #980, opened 2024-11-12
Fix selected filter being lost on tab switch

### LINKED ISSUE #552: Filters do not persist when I switch between Event and Series views.
If I filter something in the Events view, then switch to the Series view, and then switch back to the Events view, my filter for the events is lost. I would expect my filter to persist.

Arnei:
Fixes #552.

When for example setting a filter for the events table, then switching to the series table and then switching back to the events table, the previously set filter will be lost. This change preserves the filter instead.

When testing this, you may need to clear your browser cache for this to work.

Quick showcase:
[Bildschirmaufzeichnung vom 2024-11-12 12-03-42.webm](https://github.com/user-attachments/assets/0bd193b9-707b-431d-9c7a-4fab4d946ac9)


snoesberger:
Thanks for the fix! I have tested it and it works as expected, filters are now persisted when I switch between event and series view.

lkiesow (changes requested):
Unfortunately, this also breaks the quick filter selection in the admin interface:

https://github.com/user-attachments/assets/067da425-f1b0-499f-90b8-118cffe044a3

Arnei:
Thanks for catching that.

I am not able to reproduce the bug you found. Do you have futher information? I also tested against develop.opencast.org.
[Bildschirmaufzeichnung vom 2024-12-02 16-32-22.webm](https://github.com/user-attachments/assets/bb059b55-001e-4418-aacf-e3570fa95759)

marwyg:
I am also not able to reproduce this. I tested the quick filter on a testing server with opencast 16.7. I was also able to use multiple quick filters and the filters from the list. Everything worked es expected.

gregorydlogan:
This doesn't seem to work for me, even after clearing cache and cookies.  Looking at the developer console, I see a query to `http://localhost/admin-ng/series/series.json?limit=10&offset=0&filter=series:ID-wiki-commons&sort=date:DESC` which 400s.  Could it be clearing the filter because it's 400-ing?

Arnei:
That should not be the cause.

Also, just checked and it still works for me. Do you have more details?

owi92:
This is working for me and the code looks reasonable. When switching from the videos table to series, I also get the error Greg mentioned, but that happens on main as well and has been for a while iirc. Could be the cause of the original error, but I'm not sure.

Also not sure how to proceed here to be honest. It's working for Arne, Martin and me, but Lars and Greg have errors... sooo yeah no clue, sorry for this more or less useless comment.

geichelberger:
I found a problem that I was not able to reproduce on develop.opencast.org

https://github.com/user-attachments/assets/dcaa71f6-96b5-449c-9820-c99aa72a2a25

geichelberger (approved):
LGTM, I cannot reproduce the bug @lkiesow mentioned, too.

geichelberger:
@lkiesow @gregorydlogan I would merge it, and if it introduces bugs, we can fix them later. Any concerns?

Issue #69, opened 2022-07-11
Sharing links does not work - Event or workflow details not part of the URL anymore


Rillke:
**Describe the bug**
In the old Angular.js Admin UI when you open an event detail, or workflow detail, the state ist reflected in the URL. This isn't the case in the new Admin UI Picard, therefore sharing links to event details is not possible anymore.

**To Reproduce**
1. Go to https://picard.vm2.elan.codes/admin-ui/index.html
2. Log in (admin / opencast)
3. Click the event details icon.  <img width="29" alt="Screenshot 2022-07-11 at 23 07 31" src="https://user-images.githubusercontent.com/2311611/178358532-e544da26-4733-4a5f-8ee1-47987f5491b1.png">
4. The URL in the browser's address bar is still https://picard.vm2.elan.codes/admin-ui/index.html - therefore you can't share a link to this specific state of the admin UI. This was possible with the  Angular.js Admin UI, for instance: https://stable.opencast.org/admin-ng/index.html#!/events/events?storage=%7B%22pagination%22:%7B%22events%22:%7B%22resume%22:false,%22limit%22:100,%22offset%22:0%7D,%22series%22:%7B%22resume%22:false,%22limit%22:10,%22offset%22:0%7D,%22users%22:%7B%22resume%22:false,%22limit%22:100,%22offset%22:0%7D,%22jobs%22:%7B%22resume%22:false,%22limit%22:100,%22offset%22:0%7D%7D,%22sorter%22:%7B%22events%22:%7B%22title%22:%7B%22name%22:%22title%22,%22priority%22:0,%22order%22:%22ASC%22%7D%7D,%22series%22:%7B%22title%22:%7B%22name%22:%22title%22,%22priority%22:0,%22order%22:%22ASC%22%7D%7D,%22users%22:%7B%22name%22:%7B%22name%22:%22name%22,%22priority%22:0,%22order%22:%22ASC%22%7D%7D,%22jobs%22:%7B%22id%22:%7B%22name%22:%22id%22,%22priority%22:0,%22order%22:%22ASC%22%7D%7D%7D%7D&modal=event-details&tab=metadata&resourceId=ID-3d-print

mwuttke:
I can confirm this error for opencast version 14.4.

[...]

PR #981, opened 2024-11-14
Restore (some) shareable links

JulianKniephoff:
Addresses #69, but see also my remarks in https://github.com/opencast/opencast-admin-interface/issues/69#issuecomment-2473404970; I'm not sure if it closes it because people might want more state in their share links.

TL;DR this puts some of the Redux state in the URL so that you can share "deep links" into the application, as in: provide people with a link to the admin UI that directly opens the event details for a given event.

Right now you can link to:

* The event details modal of an event, and within that, to any tab, and within the workflow tab to any workflow.
* The series details modal of a series, and again to any tab of said modal.

The latter specifically necessitated putting some more state into Redux in the first place. Plus some adjacent refactoring. 🤷‍♀️

Creates links like http://localhost:3000/?persist%3AseriesDetails=%7B%22modal%22%3A%22%7B%5C%22show%5C%22%3Afalse%2C%5C%22page%5C%22%3A0%2C%5C%22series%5C%22%3A%7B%5C%22id%5C%22%3A%5C%22ID-av-portal%5C%22%2C%5C%22title%5C%22%3A%5C%22AV-Portal+Content%5C%22%7D%7D%22%2C%22_persist%22%3A%22%7B%5C%22version%5C%22%3A-1%2C%5C%22rehydrated%5C%22%3Atrue%7D%22%7D&persist%3AeventDetails=%7B%22modal%22%3A%22%7B%5C%22show%5C%22%3Atrue%2C%5C%22page%5C%22%3A0%2C%5C%22event%5C%22%3A%7B%5C%22end_date%5C%22%3A%5C%222024-11-14T01%3A13%3A19Z%5C%22%2C%5C%22agent_id%5C%22%3A%5C%22%5C%22%2C%5C%22displayable_status%5C%22%3A%5C%22EVENTS.EVENTS.STATUS.PROCESSED%5C%22%2C%5C%22needs_cutting%5C%22%3Afalse%2C%5C%22source%5C%22%3A%5C%22ARCHIVE%5C%22%2C%5C%22title%5C%22%3A%5C%22Weitsprung%5C%22%2C%5C%22has_open_comments%5C%22%3Afalse%2C%5C%22has_preview%5C%22%3Atrue%2C%5C%22technical_presenters%5C%22%3A%5B%5D%2C%5C%22has_comments%5C%22%3Afalse%2C%5C%22technical_end%5C%22%3A%5C%222024-11-14T01%3A13%3A19Z%5C%22%2C%5C%22series%5C%22%3A%7B%5C%22id%5C%22%3A%5C%22ID-av-portal%5C%22%2C%5C%22title%5C%22%3A%5C%22AV-Portal+Content%5C%22%7D%2C%5C%22presenters%5C%22%3A%5B%5C%22W%C3%A4lken%2C+Paul%5C%22%5D%2C%5C%22technical_start%5C%22%3A%5C%222024-11-14T01%3A03%3A03Z%5C%22%2C%5C%22location%5C%22%3A%5C%22%5C%22%2C%5C%22managedAcl%5C%22%3A%5C%22public%5C%22%2C%5C%22workflow_state%5C%22%3A%5C%22SUCCEEDED%5C%22%2C%5C%22id%5C%22%3A%5C%22ID-weitsprung%5C%22%2C%5C%22start_date%5C%22%3A%5C%222024-11-14T01%3A03%3A00Z%5C%22%2C%5C%22event_status%5C%22%3A%5C%22EVENTS.EVENTS.STATUS.PROCESSED%5C%22%2C%5C%22publications%5C%22%3A%5B%7B%5C%22name%5C%22%3A%5C%22EVENTS.EVENTS.DETAILS.PUBLICATIONS.ENGAGE%5C%22%2C%5C%22id%5C%22%3A%5C%22engage-player%5C%22%2C%5C%22url%5C%22%3A%5C%22https%3A%2F%2Fdevelop.opencast.org%2Fplay%2FID-weitsprung%5C%22%2C%5C%22enabled%5C%22%3Atrue%2C%5C%22hiding%5C%22%3Afalse%7D%5D%2C%5C%22date%5C%22%3A%5C%222024-11-14T01%3A03%3A00Z%5C%22%2C%5C%22selected%5C%22%3Afalse%7D%2C%5C%22workflowTabHierarchy%5C%22%3A%5C%22entry%5C%22%2C%5C%22assetsTabHierarchy%5C%22%3A%5C%22entry%5C%22%2C%5C%22workflowId%5C%22%3A%5C%22%5C%22%7D%22%2C%22_persist%22%3A%22%7B%5C%22version%5C%22%3A-1%2C%5C%22rehydrated%5C%22%3Atrue%7D%22%7D#, which is not pretty, but oh well, no one is expected to read and/or write these manually.


JulianKniephoff:
That was quick 👀 I wanted to provide an example from the test deployment but that 404s on me. 🤔

Anyway, no, it's pretty bad alright. 👌😂

mtneug:
:D Well, any URL should work if properly escaped. So people need to make sure, but that was already the case for the old UI.

JulianKniephoff:
If this bothers us we can still tweak it a little bit in two ways mainly:

* Look more closely at what info is in the URL. Looking more closely at the example myself I see a few things that don't really need to be in there; they are there because I wanted to change existing code as little as possible.
* `redux-persist` has a notion of "transformations" that you can apply to the serialized state before persisting it. We could compress and/or base64-encode it for example. 🤷‍♀️ This makes them less (read: not at all) editable by the users, but as I said, I don't think anyone is going to want to do that with the current format either.

A completely different approach would preserve editability as well: We could just put all of that stuff in the application router and have nice urls like `/events/ID-some-event/workflows/1`. This was my first instinct, but the spirit of #69 suggested a more general approach, IMO, and frankly I just wanted to see if I can make it work. The router-approach would require very specific changes for every piece of state we want to have in the URL.

JulianKniephoff:
Apart from the conflict and a little bug I just pushed a fix for, I just noticed that metadata for a series doesn't load when you visit a shared link directly.

Converting to draft for now, but will look into it shortly.

JulianKniephoff:
> In addition to the conflicts and a little bug I just pushed a fix for, I just noticed that metadata for a series doesn't load when you visit a shared link directly.

Everything should now be fixed!

Arnei (commented):
More of a thought dump than a review, but here you are.

Long URLs suck. Sure, users can throw an URL shortener at it, but it would be better if they did not have to do that in the first place.

URL now leaks data, e.g. metadata. As you suggested, applying transformations sounds like a reasonable way to solve that. 

Having to lift modal state (state that only interests the modal component) into redux seems bad practice to me. You mentioned another approach would be to use the application-router. What is the issue with that compared with the current redux-persist approach?

Code looks clean, me like. Can also confirm that the whole approach generally works.

There are some parts of the event modal that will not load their information. For example assets subtabs or workflow operation details.

Do you know if, with your current approach, it would be possible to persist a state both in store and in url? For example, if I wanted to change the code persist something about event details in local storage, could I do that?

smarquard:
@JulianKniephoff  Is this PR still active? We're interested in the ability to deep link to a specific event or series via a URL (the popup modal view ideally but events or series list filtered to that event or series would also meet the use case).

JulianKniephoff:
There are currently no plans to continue work on this that I know of. If I remember the community discourse around it correctly, it's also probably not the solution we want for #69. #1074 shows another, more restricted approach, but I don't know about any interest in and or funding for that, either. **If** we are to attack this again, we should probably check in with the community first, about which approach we should pursue. 🤔

PR #983, opened 2024-11-19
Allow series path removal via Tobira tab in series details

owi92:
This modifies the UI from https://github.com/opencast/opencast-admin-interface/pull/878 to include the option to remove a series path in Tobira ("unmounting").
Needs https://github.com/opencast/opencast/pull/6317 to work.

Also fixes a minor bug and introduces icon buttons for the series path editing (see screenshots):
Bildschirmfoto 2024-11-19 um 16 19 53
Bildschirmfoto 2024-11-19 um 16 19 58
Bildschirmfoto 2024-11-19 um 16 20 15
Bildschirmfoto 2024-11-19 um 16 20 20

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
@@ -11,6 +11,8 @@ import EventDetailsTabHierarchyNavigation from "./EventDetailsTabHierarchyNaviga
 import NewTobiraPage, { TobiraFormProps } from "./NewTobiraPage";
 import { fetchSeriesDetailsTobira, setTobiraTabHierarchy, TobiraData, updateSeriesTobiraPath } from "../../../../slices/seriesDetailsSlice";
 import { fetchSeriesDetailsTobiraNew, TobiraPage } from "../../../../slices/seriesSlice";
+import ConfirmModal from "../../../shared/ConfirmModal";
+import { Tooltip } from "../../../shared/Tooltip";
 
 
 export type TobiraTabHierarchy = "main" | "edit-path";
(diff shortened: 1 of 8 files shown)


Arnei (changes requested):
Bildschirmfoto vom 2024-12-10 15-41-34
If I open a series details tab like the above, close it, then open the series details tab of another series, the series details tab of that other series will also look like above, even if it normally wouldn't. This can only be fixed with a page reload.

Furthermore, it seems I cannot update a series path anymore. The web request to do so fails with a 400.
Request: 
http://localhost:3000/admin-ng/series/84e3ddbf-c799-41ac-83b5-22e2e4b23e8f/tobira/path
```
{
	"pathComponents": "[{\"name\":\"uiuiui\",\"pathSegment\":\"uiuiui\"}]",
	"targetPath": "/uiuiui"
}
```

owi92:
> If I open a series details tab like the above, close it, then open the series details tab of another series, the series details tab of that other series will also look like above, even if it normally wouldn't. This can only be fixed with a page reload.

huh dang, once again I can't reproduce that, but I slightly remember running into that in the past.

The other thing should be fixed with my latest push of https://github.com/opencast/opencast/pull/6317. Please don't look at the diff. It's really stupid.

owi92:
Ok, I did just manage to reproduce the first issue.
- First I need to create a series without choosing a path for it in tobira. Apparently, Tobira doesn't like that and fails to create the series in its DB (it used to be able to handle this case, need to investigate)
- Then opening the series details of the newly created series will result in your screenshot above
- The Tobira tab of other events will also be broken, like you noted above

The Tobira thing should be an easy enough fix, but that still leaves the other issue.
 Let me look into it some more.

owi92:
Alright, the error should be fixed now.

Some background information:
When a new series is created without setting a path for Tobira, the series won't be immediatly created in Tobira, but it will eventually get in there when sync happens.
In a dev environment, sync will only happen if you manullay run `cargo run -- worker` in Tobira's backend. In production, this should be always be running, but it might take several minutes, depending on the configured `poll period`.
So as long as Tobira doesn't know about the series, an error is set and the notification from your screenshot is shown. (This has "always" been like that and I don't think it needs to be changed, at least not in this PR)

The issue you noticed when opening another details modal of a series already known to Tobira was happening because the error and notification were not reset when closing a modal.
I don't have a lot of experience with redux, so I am not sure if the `removeNotification` dispatch and error reset are done in the correct places now (there are multiple options), but so far I didn't notice any unwanted side effects with my fix.

Issue #942, opened 2024-10-10
Wrong timezone handling in scheduling conflicts rrule


Nepoh:
When scheduling multiple events, the admin UI sends a request to check for conflicts. It looks like there is a timezone issue within the generated RRULE, as local time is used in `BYHOUR` while UTC is used for the start and end timestamps:
```
POST https://mec.admin.opencast.tum.de/admin-ng/event/new/conflicts
{
    "metadata": "{\"start\":\"2024-10-13T06:10:00.000Z\",\"device\":\"MRI+-+HS-B\",\"duration\":\"6300000\",\"end\":\"2025-01-01T08:55:00.000Z\",\"rrule\":\"FREQ=WEEKLY;BYDAY=TU;BYHOUR=8;BYMINUTE=10\"}"
}
```
The old admin UI on the other hand used UTC consistently:
```
{
    "metadata": "{\"start\":\"2024-10-13T06:10:00Z\",\"device\":\"MRI - HS-B\",\"duration\":\"6300000\",\"end\":\"2025-01-01T08:55:00Z\",\"rrule\":\"FREQ=WEEKLY;BYDAY=TU;BYHOUR=6;BYMINUTE=10\"}"
}
```
As a result, wrong (i.e. non-existing) scheduling conflicts are reported in the UI, preventing event creation. (On the other hand, actual conflicts may remain undetected, but I have not tested this.)

![image-2024-10-9_18-2-5](https://github.com/user-attachments/assets/511ac98e-fea5-4179-94c6-a8d350474b17)

EDIT:

Maybe `getUTCHours` and `getUTCMinutes` should be used here?

https://github.com/opencast/opencast-admin-interface/blob/a3129b961a14ba67b21ccc905e521cdfe001f0be/src/slices/eventSlice.ts#L997

PR #985, opened 2024-11-25
Use UTC consistently

jduehring:
This PR fixes https://github.com/opencast/opencast-admin-interface/issues/942 . 

As proposed in the issue, I changed the calls to `getUTCHours` and `getUTCMinutes` which seems to have solved the issue. 

For testing try to schedule multiple events. Repeat the process and check that the error message only occurs when there really is a conflict i.e. overlapping times for the same capture agent.


Arnei (approved):
Seems reasonable to me.

PR #943, opened 2024-10-10
Display LifeCyclePolicies in UI (read-only)

Arnei:
Adds a new tab to recordings called "LifeCycle Policies". Much like other tabs this tab displays information on its subject in a table format. Unlike i.e. the events tab, the LifeCycle Policies cannot be changed in any way, just be viewed. Editing is supposed to be added at a later date. 

The policy details modal looks rather crude. I hope this can be excused because Editing is supposed to follow soon-ish, which will require to completely rework the modal.

This depends on Opencast PR https://github.com/opencast/opencast/pull/6139 being merged to make any sense. Similarly, if you would like to test this, your admin interface should point to an Opencast with the PR installed.

Open question: Where should the "LifeCycle Policies" tab be displayed?
I've put the tab next to events and series for testing, but it maybe (probably) should go elsewhere? Might even deserve a new main menu item? Give me your opinions!

Screenshots:
Bildschirmfoto vom 2024-10-10 10-30-33
Bildschirmfoto vom 2024-10-10 10-30-52
Bildschirmfoto vom 2024-10-10 10-30-59

PR #1004, opened 2024-12-04
Add Create, Edit, Delete for LifeCycle Policies

Arnei:
Instead of only being able to view lifecycle policies, this commit lets you edit them, create new ones and even delete them. 

Includes #943. Can serve to replace it.

Depends on changes to the backend: https://github.com/opencast/opencast/pull/6139. These changes are currently pointed at develop.

To test this, you'll need an Opencast with backend changes mentioned above.

Demonstration:
[Bildschirmaufzeichnung vom 2024-12-04 10-45-28.webm](https://github.com/user-attachments/assets/ab48f8cb-6d4c-47aa-b3b5-3ee466676c60)


majosch:
@Arnei I just tried to spin up a test container via `podman` as noticed above, but unfortunately it fails to start:

```
~$ podman run --rm -it -p 127.0.0.1:3000:3000 ghcr.io/opencast/opencast-admin-interface:pr-1004

> admin-ui-picard@0.1.0 start
> vite --host 0.0.0.0

/admin-interface/node_modules/rollup/dist/native.js:63
		throw new Error(
		      ^

Error: Cannot find module @rollup/rollup-linux-x64-musl. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (/admin-interface/node_modules/rollup/dist/native.js:63:9)
    at Object.<anonymous> (/admin-interface/node_modules/rollup/dist/native.js:72:76)
    ... 3 lines matching cause stack trace ...
    at Module._load (node:internal/modules/cjs/loader:1104:12)
    at cjsLoader (node:internal/modules/esm/translators:346:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:286:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:234:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:473:24) {
  [cause]: Error: Cannot find module '@rollup/rollup-linux-x64-musl'
  Require stack:
  - /admin-interface/node_modules/rollup/dist/native.js
      at Module._resolveFilename (node:internal/modules/cjs/loader:1225:15)
      at Module._load (node:internal/modules/cjs/loader:1051:27)
      at Module.require (node:internal/modules/cjs/loader:1311:19)
      at require (node:internal/modules/helpers:179:18)
      at requireWithFriendlyError (/admin-interface/node_modules/rollup/dist/native.js:45:10)
      at Object.<anonymous> (/admin-interface/node_modules/rollup/dist/native.js:72:76)
      at Module._compile (node:internal/modules/cjs/loader:1469:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
      at Module.load (node:internal/modules/cjs/loader:1288:32)
      at Module._load (node:internal/modules/cjs/loader:1104:12) {
    code: 'MODULE_NOT_FOUND',
    requireStack: [ '/admin-interface/node_modules/rollup/dist/native.js' ]
  }
}

Node.js v20.18.1
```

JulianKniephoff:
@majosch that's an unrelated problem between npm and our dependencies that we should investigate and fix, but not something we can do anything about as part of this PR, can you maybe try building the project from source?

The container (at least the way you try to start it) is not going to be of much help anyway, since it's going to talk to develop.opencast.org, which doesn't have the life cycle management backend code, yet.

JasminNovotni:
Found a few bugs during testing this:

- Filter `isPartOf`: If the series title contains a `:` it cuts off the rest of the title

- Filter `isPartOf`: The filter seems to be working incorrectly. If I set the filter and select a series with the type `Search` it always accesses all events, even those that aren't part of the series.

- Filter `end_date`: The filter seems to be working incorrectly too. Most of the time, it takes all events despite specifying an exact date and time. And sometimes, for example with `Greater_than` I get a number of events that I can't understand.

- Using `Timing` `repeating`: I always get the error message "The lifeCycle policy could not be saved" when saving.

- Entering invalid filter specifications: If I use specifications for a filter that don't work, there is no error message but a blank page appears. (e.g., filter description + empty value.)

Arnei:
Thanks for testing!

> Filter isPartOf: If the series title contains a : it cuts off the rest of the title

Can you give more details on what exactly is going wrong for you? It looks fine to me

Bildschirmfoto vom 2025-06-17 10-34-16

> Filter isPartOf: The filter seems to be working incorrectly. If I set the filter and select a series with the type Search it always accesses all events, even those that aren't part of the series.

Thanks for catching that. That was a bug in the backend (https://github.com/opencast/opencast/pull/6139.), should be fixed now.

> Filter end_date: The filter seems to be working incorrectly too. Most of the time, it takes all events despite specifying an exact date and time. And sometimes, for example with Greater_than I get a number of events that I can't understand.

Thanks for catching that. Even though the backend offered `end_date` as an option, it was not actually supported for some reason. Should be fixed now.

> Using Timing repeating: I always get the error message "The lifeCycle policy could not be saved" when saving.

Thanks for catching that. Fixed in backend. That message should not appear anymore.

> Entering invalid filter specifications: If I use specifications for a filter that don't work, there is no error message but a blank page appears. (e.g., filter description + empty value.)

Thanks for catching that. You should now be unable to create a policy if you leave any field in a filter blank.

JasminNovotni:
Thanks for fixing :)


> Can you give more details on what exactly is going wrong for you? It looks fine to me

The `:` bug is a bit more complicated than I thought. I've been able to recreate it a few times now.

If a series has both a `.` and a `:` in the title, it cuts off the title before the `:` (not after as I had written first)
grafik

grafik

majosch:
* It would be nice if changing the "Filter" to `created`/`start date`/`end date` would automatically update the "Type"

https://github.com/user-attachments/assets/9bb1f13d-5355-4b80-8637-0ce7d9e723d6

* Setting an "Action Date" with a dedicated "Time" is a bit awkward:

https://github.com/user-attachments/assets/42872078-2c53-4523-9235-27c4fe30dd62

* When not specifying any "Filters", all events are targeted. That's a valid use case - but maybe there should be a warning?

* When using `REPEATING` as "Timing" the "Cron Trigger" doesn't work for me:

https://github.com/user-attachments/assets/8445fad9-c832-4382-ae6b-fd20dba4dc11

* In the "Summary" tab, the value for "Active" is always empty:

Bildschirmfoto vom 2025-11-17 10-33-36

* As reported by @JasminNovotni: If a series name contains `.` and `:`, the value is not displayed properly:

https://github.com/user-attachments/assets/7d5c3888-5ca6-4b5d-808b-f119855d6df1

* When editing an existing policy and clicking "Save", the notification is not human readable but "NOTIFIACTIONS.LIFECYCLEPOLICY_ADDED"

* If a policy has already fired: Is there any scenario that does trigger new tasks, when changing filters/timing/action date? When I try to do so (and enabling the "Active"-boolean), no new tasks are created. I'm absolutely fine with this approach, but maybe it would be more intelligible if the policy is immutable and prevent users from changing filters/timing/action date?

* Is it reasonable to enforce the ACL "public" to policies or are there any use cases you want to have private/authenticated/custom ACLS?

Arnei:
Thanks for testing, will take a look at all those bugs.

As for your other remarks:

> If a policy has already fired: Is there any scenario that does trigger new tasks, when changing filters/timing/action date? When I try to do so (and enabling the "Active"-boolean), no new tasks are created. I'm absolutely fine with this approach, but maybe it would be more intelligible if the policy is immutable and prevent users from changing filters/timing/action date?

If you reenable a lifecycle policy by setting it to active again, it should run again when the timing condition is met. It won't create tasks for entities it has already created tasks for though. So if you have a situation where you reenabled a lifecycle policy, that policy was triggered according to its timing and that policy should have created a task for a new event but did not, then I would consider that a bug.

> Is it reasonable to enforce the ACL "public" to policies or are there any use cases you want to have private/authenticated/custom ACLS?

I don't see what the benefit(s) of forcing a "public" ACL would be, could you elaborate?

majosch:
Regarding ACLs... It feels like I'm missing some point... does it make any difference using different ACLs when creating/updating polices? Especially when these actions are restricted to admins only, who can always access all assets?

Arnei:
If you restrict these actions to admins only, which is entirely reasonable, then what you do with the ACL does not matter because admins get to ignore the ACL anyway. But then there is also no reason to force them "public". If anything I would enforce "private", to make it entirely clear that no one but the admins is supposed to have access.

ACLs for lifecycle policies are not useful if only admins handle them. They are there for a future where we have figured out how to reasonably run lifecycle policies for non-admin users, at least in my opinion.

Arnei:
I hope I managed to address all the other feedback.

I did not address

> Setting an "Action Date" with a dedicated "Time" is a bit awkward

because I could not come up with a good idea on how to improve setting a date here (and this could arguably be considered a more general problem as our date pickers work like this everywhere).

majosch:
I'm not sure if this is caused by the PR itself or if it's just a bug: The event info modal displays a wrong series name, if a new series is created manually?


https://github.com/user-attachments/assets/586c2767-5142-4f38-9282-6403b19262a8

Arnei:
> I'm not sure if this is caused by the PR itself or if it's just a bug: The event info modal displays a wrong series name, if a new series is created manually?
> out.mp4

Thanks for finding that and including a video. However, I am unable to reproduce the behaviour. I tried with this PR, develop and r/19.x. Not really sure what could cause this either :(

PR #1009, opened 2024-12-12
Remember table sorting

Arnei:
Tables can be sorted by columns, ascending or descending. However, when switching to a different tab and back again, the sorting reverts to the default. With this change, the sorting is remembered.

Does not persist the sorting, meaning it still reverts to the default after a page reload. I did not do this because the only to way to get rid of persisted sorting would be to the clear the browsers cache, and I did not want alienate people.

Also sneakily changes the default sorting of the series table in the same way as #1008.

### How to test this

PR can be tested as is, no other dependencies or configuration necessary.

[Bildschirmaufzeichnung vom 2024-12-12 11-44-03.webm](https://github.com/user-attachments/assets/e6366bd7-a733-4324-ad17-b4c3e3beae9a)

### DIFF
--- a/src/selectors/tableSelectors.ts
+++ b/src/selectors/tableSelectors.ts
@@ -14,8 +14,8 @@ export const getPageLimit = (state: RootState) => state.table.pagination.limit;
 export const getPageOffset = (state: RootState) => state.table.pagination.offset;
 export const getNumberDirectAccessiblePages = (state: RootState) => state.table.pagination.directAccessibleNo;
 export const getResourceType = (state: RootState) => state.table.resource;
-export const getTableSorting = (state: RootState) => state.table.sortBy;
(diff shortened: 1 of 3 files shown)


gregorydlogan (changes requested):
This appears to be causing the following exception every time the page is *started* (eg F5, or a fresh navigation).

```
2025-01-16 17:00:51,271 | WARN  | (WebApplicationExceptionMapper:73) - javax.ws.rs.WebApplicationException: HTTP 400 Bad Request                                                                                                
        at org.opencastproject.index.service.util.RestUtils.parseSortQueryParameter(RestUtils.java:233)                                                                                                                         
        at org.opencastproject.adminui.endpoint.AbstractEventEndpoint.getEvents(AbstractEventEndpoint.java:2668).
...
```

geichelberger (approved):
LGTM

PR #1017, opened 2024-12-17
Reduce duplicate code for NavBar components

Arnei:
This patch attempts to reduce duplicate code by moving components that would usually be passed to the NavBar component INTO the NavBar component.

This does come with the trade-off that the NavBar component is now less general. However, as it is the NavBar component was pretty pointless anyway so this should not be an issue.

This is the "NavBar":
Bildschirmfoto vom 2024-12-17 09-42-22

### How to test this

Can be tested as usual, no configuration or Opencast patch necessary. There should be no functional changes, so the main thing to test here is that everything behaves as usual.


src/components/events/Events.tsx
```diff
@@ -120,7 +93,7 @@ const Events = () => {
 		};
 
 		// Fetch events every minute
-		let fetchEventsInterval = setInterval(loadEvents, 5000);
+		let fetchEventsInterval = setInterval(() => loadEvents(dispatch), 5000);
```
owi92, line 96: Not related to your changes, but: The comment says to "fetch events every minute", yet the interval is set at 5s. This is also repeated in the other resources.
Eh but probably just ignore it for now, maybe at some point we might want to update that comment.

owi92 (approved):
As far as I can tell, there are no functional changes, so it behaves as described and expected. Code looks reasonable (Maybe you could factor out some more stuff, but this is already a good improvement and enough for now imo).

PR #1018, opened 2024-12-17
Reduce duplicate code in action cells

Arnei:
Mostly by turning the button that's copy-pasted everywhere into its own component.

These are the "action cells":
Bildschirmfoto vom 2024-12-17 17-38-56

### How to test this

Can be tested as usual, no configuration or Opencast patch necessary. There should be no functional changes, so the main thing to test here is that everything behaves as usual.


owi92 (approved):
lgtm, everything's still working.

Issue #1005, opened 2024-12-07
Issues of the date field in the scheduling tab of the event properties 


marcusm7:
### Steps
1. Create a scheduled event @ stable.opencast.org
2. Click Open Event Details, select Scheduling tab
3. Work with the Start date field.

The currently set date is not set:
![image](https://github.com/user-attachments/assets/12163e81-6fa6-43b9-9f67-27cb89290f20)
Expected: the date set in Step 1 above should be displayed.  
  
### Note:  
- To minimize possible merge conflicts/ extra work, please keep in mind that there is a PR for the replacement of the date picker. The foxus stealing issue can be seen there too (https://github.com/opencast/opencast-admin-interface/pull/986#issuecomment-2523281300) but doesn't seem to be related with the PR. 
- https://github.com/opencast/opencast-admin-interface/issues/1016
  
### My Environment:
```
$ plasmashell --version
plasmashell 6.2.3
$ firefox --full-version 
Mozilla Firefox 133.0 20241121140525 20241121140525
```

Arnei:
Suggestion: If you already feel the need to create titles like "Issue1" and "Issue2", I'd suggest to create separate Github issues for them. That way they can be discussed and worked on individually.

marcusm7:
> Suggestion: If you already feel the need to create titles like "Issue1" and "Issue2", I'd suggest to create separate Github issues for them. That way they can be discussed and worked on individually.

I thought it would be better to combine the issues concerning the very same element in one ticket - as two problematic aspects of the same problem.  I separated the issues as suggested.

PR #1019, opened 2024-12-18
Fix date not rendering in scheduling tab

Arnei:
Fixes https://github.com/opencast/opencast-admin-interface/issues/1005.

The date was not rendering the event details scheduling tab. This patch fixes that.

### How to test this
Can be tested as usual, no configuration or Opencast patch necessary. To test, follow the steps in the linked issue and check if the issue is resolved.


marcusm7:
https://test.admin-interface.opencast.org/1019/2025-01-06_11-02-00/

When I open the page the date picker for the start date is already openened (firefox 134 in wayland). Is this a side effect in the test-mockup or a bug?

image

Arnei:
I can reproduce that behaviour on the main branch, but only if I had that specific date picker open before reloading the page. I cannot reproduce it any other way. 

Still, this is technically a bug. Want to open an issue for it?

PR #1021, opened 2024-12-18
Fix notification display in event metadata tab

### LINKED ISSUE #947: Edit Metadata: no editing disabled hint, while status is running
There should be an editing disabled hint/ reason, while the editing is disabled.

What it looks like at stable.opencast.org (Opencast 16.x – 578aa2e – built on 2024-10-11)
image

Screenshot from a previous version:
image

Arnei:
Fixes #947.

Notifications were not being rendered in the metadata tab of the event details, especially the "Active Workflow"-notification. This patch fixes that.

This patch also fixes:
- The "Active Workflow"-notification being rendered multiple times.
- The "Active Workflow"-notification being rendered even when there is no active workflow anymore.

### How to test this

Can be tested as usual, no configuration or Opencast patch necessary. To test, start a workflow on an event and open the event details. Then click through the tabs a bit.

Bildschirmfoto vom 2024-12-18 15-13-44


src/components/events/partials/modals/EventDetails.tsx
```diff
@@ -80,10 +83,36 @@ const EventDetails = ({
...
 	useEffect(() => {
 		dispatch(removeNotificationWizardForm());
-		dispatch(fetchMetadata(eventId));
-		dispatch(fetchSchedulingInfo(eventId));
-		dispatch(fetchEventStatistics(eventId));
-		dispatch(fetchAssetUploadOptions());
+		dispatch(fetchMetadata(eventId)).then();
+		dispatch(fetchSchedulingInfo(eventId)).then();
+		dispatch(fetchEventStatistics(eventId)).then();
+		dispatch(fetchAssetUploadOptions()).then();
```
owi92, line 89: I don't think you need these `then()`s.

PR #1022, opened 2024-12-19
Fix "TypeError: pages[offset] is undefined"

Arnei:
This patch fixes the above error from the webconsole.

### How to test this

The error message in the web console can only consistently be seen when running the admin locally via `npm start`, so I suggest doing that. Functionality should otherwise remain unchanged, so test that all pages load as usual.


owi92 (approved):
I can't reproduce the error on main, but your changes don't break anything either. Maybe check if it's still needed, or simply merge, should be fine™ 🤷

PR #1024, opened 2024-12-20
Fix request to /staticfiles failing during dev

### LINKED ISSUE #1023: Development: Can't proxy to /staticfiles
This bug only concerns development, when running the admin ui locally and proxying requests to Opencast. This is not an issue coming up in production.

When trying to upload a file for a theme, the related web request against `/staticfiles` returns with a 404. This seems to be the only endpoint with that particular problem.

Arnei:
Fixes #1023.

The created request to /staticfiles were not matching our proxy regex. This fixes that.

### How to test this

This only affects development, not completed builds, so you will have to install this PR yourself and run it locally. You will also need an Opencasts with themes enabled that you can proxy against. Then open the "Create Theme" dialog and check if you can successfully upload a file in any of the file uploads.

### DIFF
--- a/src/components/shared/wizard/FileUpload.tsx
+++ b/src/components/shared/wizard/FileUpload.tsx
@@ -55,7 +55,7 @@ const FileUpload = <T extends RequiredFormProps>({
 		const data = new FormData();
 		data.append("BODY", file, file.name);
 		axios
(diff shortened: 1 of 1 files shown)


owi92 (approved):
This is working 👍

PR #1041, opened 2025-01-03
Fix metadata fields not editable during multi edit

Arnei:
In the modal for editing multiple events at the same time, the fields "presenter(s)" and "contributor(s)" were not editable in some cases. This patch fixes that.

### How to test this

This patch should ideally be tested with a production Opencast. Unfortunately, I do not actually know what the exact Opencast configuration is that causes the error this patch this is supposed to fix.


owi92 (changes requested):
Is there a trick in reproducing the bug? I didn't notice any issues either on main or this branch, but the code looks reasonable.

I did notice however that clicking the `start date` field when editing multiple events crashes the app. This is also the case on main, so you don't necessarily need to fix it here, but it would be fitting I suppose.
So if you don't want to do that, consider this approved.

Arnei:
> Is there a trick in reproducing the bug? 

Likely, but I unfortunately do not know it. I developed this against the Opencast of one of our clients, but was unable to determine which change in their configuration was responsible.

> I did notice however that clicking the start date field when editing multiple events crashes the app. This is also the case on main, so you don't necessarily need to fix it here, but it would be fitting I suppose.

I'd prefer handling that in a separate PR. Would you open an issue for it?

owi92 (approved):

PR #1042, opened 2025-01-06
Migrate SASS

### LINKED ISSUE #1040: Dart SASS deprecation warnings
When starting the app via `npm start` there are a slew of deprecation warnings regarding Sass. We should do something about that.

```
Deprecation Warning: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.

More info: https://sass-lang.com/d/legacy-js-ap
```

Arnei:
Fixes #1040.

Vite informs us that our SASS will be deprecated with the next major SASS version (2.0.0). This would stop us from upgrading to the next major vite version (7.0.0), so this patch aims at getting our SASS undeprecated.

Almost all of the changes in this PR were generated with the npm module "sass-migrator".

### How to test this

This patch can be tested as usual and only includes style changes. Ideally this patch does not affect how the ui behaves, so the goal of testing this PR is simply to check if everything still works as expected.


geichelberger (approved):
LGTM clicked through the UI, and it looked and felt the same.

PR #1044, opened 2025-01-06
Fix string "true" not showing in extended metadata

Arnei:
In the extended metadata tab of the event details, fields with the stirng  value "true" or "false" would not render. The fields value could only by checked by clicking into it.

This fixes that by removing a method that, as far as I can tell, is absolutely pointless.

### How to test this

To properly test this, an Opencast with extended metadata for either events or series is required. You can find more details on how to configure that here: https://docs.opencast.org/develop/admin/#configuration/metadata/#extended-metadata


Arnei:
Thanks to https://github.com/opencast/opencast-admin-interface/pull/1045 this is now obsolete, so I'll close it.

PR #1045, opened 2025-01-07
Reduce code for extended metadata

Arnei:
Some of the code used for rendering extended metadata is very similar to the code used for rendering common metadata. This patch attempts to reduce the resulting code duplication a bit.

Includes a logic change on how we store metadata values in formik for new events/series, by storing them the same way extended metadata is stored.

### How to test this

To properly test this, an Opencast with extended metadata for events and series is required. You can find more details on how to configure that here: https://docs.opencast.org/develop/admin/#configuration/metadata/#extended-metadata


geichelberger:
This fixes a bug; with the new admin-UI, creating a series is impossible if an extended metadata catalog is defined.

src/slices/eventSlice.ts
```diff
@@ -462,7 +461,7 @@ export const postNewEvent = createAppAsyncThunk('events/postNewEvent', async (pa
 	} | undefined = undefined;
 
 	// prepare metadata provided by user
-	let metadataFields = prepareMetadataFieldsForPost(metadataInfo.fields, values);
+	let metadataFields = prepareExtendedMetadataFieldsForPost([metadataInfo], values)[0].fields;
```
geichelberger, line 464: What is the benefit of using the prepareExtendedMetadataFieldsForPost instead of prepareMetadataFieldsForPost? As we cannot get rid of the method as it's used internally, I prefer the old version regarding readability.

src/slices/seriesSlice.ts
```diff
@@ -192,10 +191,10 @@ export const postNewSeries = createAppAsyncThunk('series/postNewSeries', async (
 	const { values, metadataInfo, extendedMetadata } = params
 
 	// prepare metadata provided by user
-	let metadataFields = prepareSeriesMetadataFieldsForPost(
-		metadataInfo.fields,
+	let metadataFields = prepareSeriesExtendedMetadataFieldsForPost(
```
geichelberger, line 194: Same as for events, but can't we eliminate the methods at all and use the one for the events?

My patch (https://github.com/geichelberger/opencast-admin-interface/commit/d656514b024c5bc8c9f76d14cec3dd7df6065df0) for the series creation bug mentioned in the comment was using the methods of the event metadata and it seems to work, or did I miss something?

geichelberger (changes requested):

src/slices/eventSlice.ts
Arnei: Thanks for pointing that out. Iirc I changed this because I planned to combine the two methods (prepareMetadataFieldsForPost and prepareExtendedMetadataFieldsForPost), but then never did. Will change it back.

src/slices/seriesSlice.ts
Arnei: > can't we eliminate the methods and use the one for the events?

I did not check that. At a glance, maybe? The methods for "extended" metadata are certainly the same. Would have to check in more detail if the differences for the "normal" metadata methods are relevant and how they may be condensed.

Arnei: > I realized later that your PR is fixing the bug

Not intended, but great to hear :D

geichelberger: Ah, now I see why you used the method; because you changed the name of the common metadata value fields, they are also using the catalog prefix. That would mean prepareSeriesMetadataFieldsForPost needs a third argument with `metadataInfo.flavor + "_"`, which is error-prone, or change the method signature like this:

```
export const prepareSeriesMetadataFieldsForPost = (
	metadataInfo: MetadataCatalog,
	values: { [key: string]: string[] },
	formikIdPrefix = metadataInfo.flavor + "_",
)
```
I don't know if that's the best solution.

Arnei: For now I've reverted to the less readable call to "prepareSeriesExtendedMetadataFieldsForPost", and made "prepareMetadataFieldsForPost" private to indicate that it is only to be used internally. With the goal of differentiating between metadata and extended metadata as little as possible, this makes the most sense to me, as you now call the same function in both cases. Certainly could be better though (better method signature, combine methods, more intuitive method names...)

Arnei: Thrown in some more changes, please check how you like them.

PR #1046, opened 2025-01-08
Reduce duplicate code for modals

Arnei:
Introduces a generic modal component and replaces existing modal code with the new component. This intention here is to reduce code duplication and to make future generic changes to modals (i.e. focus trapping) easier.

### How to test this

No special dependencies or configuration necessary, can be tested as is. Just click through every modal (i.e. "Event Details") you can find and check if it still properly opens and closes, and still looks like expected.


geichelberger:
I found a bug. You can skip the stepper by clicking on the steps, at least for the events:

[modal-bug.webm](https://github.com/user-attachments/assets/8b5b8f1d-94b9-4fac-ba7d-c5cb99a208c6)

Arnei:
I can reproduce the bug you showcased on the `main` branch, so this PR is probably not guilty of it. I'll create an issue for it.

geichelberger (approved):
LGTM, I clicked through a lot of modals, and they were all working as expected or broken like before :)

PR #1048, opened 2025-01-09
Properly hide Tobira tab

Arnei:
In the event details tabs, the Tobira tab is showing up even if Tobira is not configured. This patch should fix that.

### How to test this

Ideally this is tested by someone with a Tobira setup, so that they can confirm that the tobira tab still correctly shows up if Tobira is indeed configured. Properly setting up Tobira is somewhat involved, so I'll direct all questions about that to the Tobira docs and the main Tobira developers.

### DIFF
--- a/src/components/events/partials/modals/EventDetails.tsx
+++ b/src/components/events/partials/modals/EventDetails.tsx
@@ -164,7 +164,7 @@ const EventDetails = ({
 			accessRole: "ROLE_UI_EVENTS_DETAILS_COMMENTS_VIEW",
 			name: "tobira",
 			page: EventDetailsPage.Tobira,
-			hidden: tobiraError?.message?.includes("503"),
+			hidden: (tobiraError === undefined || tobiraError?.message === undefined || tobiraError?.message?.includes("503")),
 		},
 		{
 			tabNameTranslation: "EVENTS.EVENTS.DETAILS.TABS.STATISTICS",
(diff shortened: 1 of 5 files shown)


src/components/events/partials/modals/EventDetails.tsx
```diff
@@ -164,7 +164,7 @@ const EventDetails = ({
 			accessRole: "ROLE_UI_EVENTS_DETAILS_COMMENTS_VIEW",
 			name: "tobira",
 			page: EventDetailsPage.Tobira,
-			hidden: tobiraError?.message?.includes("503"),
+			hidden: (tobiraError === undefined || tobiraError?.message === undefined || tobiraError?.message?.includes("503")),
```
owi92, line 167: This also hides the Tobira tab when it is configured correctly in Opencast. Maybe you can add 
```
const tobiraStatus = useAppSelector(state => getEventDetailsTobiraStatus(state));
```
in line 103 and then here use
```suggestion
			hidden: ((tobiraError === undefined || tobiraError?.message === undefined) && tobiraStatus === "failed")
				|| !!tobiraError?.message?.includes("503"),
```

and then the same for `SeriesDetails` and `NewSeriesWizard`.
The `!!` part of `!!tobiraError?.message?.includes("503")` should only be necessary in `NewSeriesWizard`.

owi92 (changes requested):
See my comment. But yeah, I'm not surprised that the `503` check alone isn't enough/fails when Tobira is not configured. Somehow I don't know how to _unconfigure_ Tobira, so I can't test this properly 🥲

src/components/events/partials/modals/EventDetails.tsx
Arnei: Thanks for the review. I forgot that we only assume Tobira to be unavaible if we get a 503 response explicitly. The new changes hopefully offer a better solution.

owi92 (approved):
This is working now.

PR #1049, opened 2025-01-09
Remove unused warning during bulk delete

Arnei:
This patch removes a warning in the bulk delete modal that claims the user is not authorized.
Removing this for two reasons:
1. The warning claims to have done an authorization check. That check is never performed, but the warning is always displayed nonetheless.
2. Users currently cannot select events if they don't have write access on them anyway (they don't show up in the events table in the first place), so there is no need to perform such an authorization check in the first place.

### How to test this

Can be tested as usual.


geichelberger (approved):
LGTM

PR #1052, opened 2025-01-10
Restore reset button for edit table view modal

### LINKED ISSUE #452: Reset Button “Edit Table View”
Where is this “reset” button in the “Edit Table View” Window? "The order and selection will be saved automatically. Press "Reset" to restore the default view".

Arnei:
In the edit table view modal, the user can select which columns are displayed for the given table. The modal has a help text that mentions a "Reset" button, but that button is nowhere to be seen. This patch restores the reset button.

Fixes #452.

### How to test this

Can be tested as usual.


geichelberger:
It will result in a cleaner git history if you don't merge the main branch into the feature branch and rebase it instead.

geichelberger (approved):
LGTM

PR #1053, opened 2025-01-10
Remove btm spacer, fix scroll bars in modals

Arnei:
Since we decided that our modals should be scrollable, some of them always have a scrollbar showing, due to an absolutely positioned white box at the very bottom of the modal. I am not sure what the original purpose of this "btm-spacer" component was supposed to be, but it seems superfluous to me and so I am removing it. This should stop the scrollbars from appearing unnecessarily.

Before:
Bildschirmfoto vom 2025-01-10 16-06-22

After:
Bildschirmfoto vom 2025-01-10 16-06-50

### How to test this

Can be tested as usual


owi92 (approved):
Lgtm. Also no idea what that `btm-spacer` was useful for. Might have had a purpose once, but I think removing it is fine.
Scroll behaviour is fixed and using `overflow: auto` instead of `overflow: scroll` makes sense to me.

PR #1056, opened 2025-01-14
Remove unused file

Arnei:
Looks to be  a leftover from modernizing redux.We should have no use for this anymore, so it can go.

### How to test this

Can be tested as usual, check if the asset upload still works.


geichelberger (approved):
LGTM

PR #1059, opened 2025-01-14
Reduce duplicate code for modal content

Arnei:
This code structure appears in like 50 files, so I'm putting it in a component so I don't have to see it again.

While there are many changed lines, 99% of them are indentation changes, so this is hopefully fairly easy to review.

### How to test this

Can be tested as usual.


ziegenberg (changes requested):
Could you separate the whitespace changes into a separate commit?

Arnei:
I can try.

Arnei:
Although the white space changes are directly related to the introduction of the new component, so it seems reasonable to me to have them in the same commit. Can you explain to me how separating them into their own commit helps with `git blame`?

ferishili (approved):
Thanks @Arnei.
This looks good to me! Since the changes primarily involve replacing the modal HTML with a modal component, I’d appreciate it if you could highlight any areas where additional modifications were made. That way, I can focus my testing on those specific parts. 🚀

ferishili (changes requested):
I had to add this here to clarify the issue:
I noticed that in User Dialog → Roles List, the modal is not functioning correctly. After reviewing other PRs, I realized this could be a problem in the test environments. However, to ensure this PR is safe, I want to highlight this here first, as it is related to the changes.

Arnei:
Thanks for the review. Could you give me more information on the issue you are seeing?
- What exactly is not "functioning correctly" and in which way?
- What test environment are you using? Test deployment with mock data, docker or a local version of the PR?

ferishili (approved):
The issue I raised has been resolved, so this PR looks good from my side!

PR #1063, opened 2025-01-15
Reduce duplicate code - button like anchor

Arnei:
We have many buttons using the "button-like-anchor" class. This patch introduces a button component for these buttons, with built-in tooltip and access check to better support common code combinations for those buttons. In and of itself not a particularly impressive component, but the underlying goal here is to push the app more towards using common basic components.

### How to test this

Can be tested as usual.


src/components/shared/Table.tsx
```diff
@@ -157,12 +158,11 @@ const Table = ({
...
-						<button
-              onClick={() => showEditTableViewModal()}
-              className="button-like-anchor"
-            >
-                {t("TABLE_EDIT")}
-            </button>
+						<ButtonLikeAnchor
+							onClick={() => showEditTableViewModal()}
+						>
+								{t("TABLE_EDIT")}
```
ferishili, line 164: Indentation! 😄 
```suggestion
							{t("TABLE_EDIT")}
```

ferishili (commented):
Thanks @Arnei
I have provided some inline suggestions, which could be replaced easily!
However, in some places I could still find the old button-like-anchor like:
- https://github.com/opencast/opencast-admin-interface/blob/c0b6d17f667adf8314c894ffa87dbf7b448490c3/src/components/shared/IconButton.tsx#L31 This could be totally replaced, right?
- https://github.com/opencast/opencast-admin-interface/blob/c0b6d17f667adf8314c894ffa87dbf7b448490c3/src/components/shared/modals/DetailsModal.tsx#L37
- https://github.com/opencast/opencast-admin-interface/blob/c0b6d17f667adf8314c894ffa87dbf7b448490c3/src/components/shared/modals/Modal.tsx#L73

ferishili (changes requested):
Forgot to submit my request changes! 😄

ferishili (approved):
Thanks @Arnei 
LGTM

Issue #1062, opened 2025-01-14
Changing the scheduled hour also changes the date


gregorydlogan:
Oh boy.  I'm going to do this in video form since it's probably easier to get:

[2025-01-14 12-05 (Display, OC Studio).webm](https://github.com/user-attachments/assets/7bfaca81-52c8-484e-9d08-b86b5e3c5924)

This occurs regardless of what start time I set.  Notably this does *not* happen for @lkiesow who is in a not-negative-UTC timezone.

I suspect that [this comment](https://github.com/opencast/opencast-admin-interface/blob/962a9e16710547953acc12bf1464be7d17bffc4c/src/slices/eventSlice.ts#L508) might be relevant but haven't stepped through the code.

Further weirdness: Despite what I said earlier (when it *wasn't* doing this), the end time is also displaying weird behaviour later in the day

[2025-01-14 16-28 (Display, OC Studio).webm](https://github.com/user-attachments/assets/80c7ecb4-93d1-408a-9576-a72ffdf3e7ae)

gregorydlogan:
Trying to schedule things is also very broken, obviously.  [This](https://github.com/opencast/opencast-admin-interface/blob/962a9e16710547953acc12bf1464be7d17bffc4c/src/slices/eventSlice.ts#L907) line has a value of "2025-01-14T00:00:00.000Z" when I try and schedule a recording for that time in UTC-6, so we've at the *very least* lost the fact that we're in UTC-6 (the browser renders it as 1800 on 25-01-13), and then the next few things muck with the hours so...

This reminds me a lot of the pain we went through with the original admin UI.  The USask team exposed a *lot* of timezone related things back then :D

Arnei:
Fixed by  #1069

PR #1069, opened 2025-01-15
Fix scheduling for negative timezones

Arnei:
And hopefully all other timezones. The date for a new scheduled event should now always be correct, no matter what hour/minute is selected.

Aims at fixing #1062.

### How to test this

Can be tested as usual. Try scheduling events with many different dates, hours, minutes, durations and check if it works as intended. Switch timezones if possible.

### DIFF
--- a/src/components/events/partials/wizards/NewEventSummary.tsx
+++ b/src/components/events/partials/wizards/NewEventSummary.tsx
@@ -190,7 +190,7 @@ const NewEventSummary = <T extends RequiredFormProps>({
 												</td>
 												<td>
 													{t("dateFormats.date.short", {
-														date: formik.values.scheduleStartDate,
+														date: renderValidDate(formik.values.scheduleStartDate),
 													})}
 												</td>
 											</tr>
@@ -210,7 +210,7 @@ const NewEventSummary = <T extends RequiredFormProps>({
 													</td>
 													<td>
 														{t("dateFormats.date.short", {
-															date: formik.values.scheduleEndDate,
+															date: renderValidDate(formik.values.scheduleEndDate),
 														})}
 													</td>
 												</tr>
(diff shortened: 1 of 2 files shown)


gregorydlogan:
Ok, **super** nitpicky here...

Single Event Scheduling:
- Scheduling an event strictly in the future works 
- Scheduling an event that would currently in progress gives me 'The event has already ended'.  We may not have a string for "Can't schedule things currently in progress", so this may have to be enough for now.
- Scheduling an event strictly in the past gives me the correct error

Multiple Event Scheduling
- Using the multiple event scheduling, I can events in the past :D
- The modal does not prevent me from scheduling something for Monday and Tuesday, but restricting the date range to 25-01-16 -> 25-01-17.  Upon submitting the scheduling request I get an immediate error saying the event could not be scheduled.  Assuming at least *one* of the days and dates match, the even creation succeeds for those matching combination(s).
- With an initial start time of 11:00, and end of 11:55, if I change the *end* time to 10 I get a duration of 23:55 without an indication if it's changing the end *date*
    - image
    - Changing the end time to 02 gets me a duration of 15 hrs, so I'm guessing it's incrementing the day, but that's not made clear like single event scheduling
    - Changing the start time such that it would cause (or undo) a wrap over a day similarly works correctly duration wise, but does not cause the end time to display the date next to it.
    - TBH, not sure scheduling multiples and going over the end of the day makes any sense whatsoever, but that's a different PR.

Have not tested agents in mismatched timezones (eg: user TZ != core TZ != agent TZ).  Does not block this PR imo.

Summary: Fixes the issue, found a bunch of nits.  Do you want to fix them here, or should I file follow-on issues?

Arnei:
Being able to schedule events in the past at the very leasts sounds bad and sounds like it could have been caused by this PR, so I (or someone else) should take a look at that. Thanks for the review!

Arnei:
Did some quick testing, and it *seems* like the nitpicks you described already existed before this PR? In which case I would leave them for a separate PR, because this PR is not claiming to fix them and so this PR can get merged faster.

gregorydlogan:
I can file the relevant nitpicks as issues so they don't get lost.

PR #1072, opened 2025-01-15
Remove unspecified filter on clear

Arnei:
The clear button clears the table filter field of all selected filters. It will however not remove an unspecified filter, meaning the dropdown menu. This patch makes is so that so that those are cleared as well, making sure the filter area is really clear.

Bildschirmfoto vom 2025-01-16 08-53-36

### How to test this

Can be tested as usual.


geichelberger (approved):
LGTM

PR #1074, opened 2025-01-15
Series links

JulianKniephoff:
This is an alternative to the use case that originally made me implement #981. Instead of persisting parts of the Redux store in the URL, this just reads a series ID from a URL parameter when looking at the series or events tables.

I hope the community likes this approach more. ;D


Arnei:
Taking a brief look at the code changes, I'm assuming you're also not happy with modals and might be interested in this PR here https://github.com/opencast/opencast-admin-interface/pull/1046

JulianKniephoff:
Modals in this code base are literally the devli 🤬

Issue #1065, opened 2025-01-15
Tobira integration: When editing the path of a series a badly named link is shown


dagraf:
When an administrator starts editing the path of a Opencast series that allready got integrated into tobira at the upper left corner a link called "Show all pages" does appear (see screenshot). Clicking on it closes the editing part and the direct link to series and the pages in Tobira that contain this series get shown again.

**Steps to reproduce:**
1. Have a Opencast series.
2. Open the series details of this series and switch to the tab "Tobira".
3. Hit on the "Pen" icon ("Edit path").
4. See the "Show all pages" link at the upper left corner.

**Suggest solution:**
Rename this link to "Back" or "Discard".

<img width="896" alt="Bildschirmfoto 2025-01-15 um 13 41 45" src="https://github.com/user-attachments/assets/07bd030e-8776-454a-8547-0ae7417cb5de" />

owi92:
Would you mind if I renamed this to "Overview"/"Übersicht" instead @dagraf?
"Back" or "Discard" might be a little ambiguous as well.

dagraf:
Maybe "Back" might be ambiguous but I think I would still prefer "Discard" over "Overview". Because:
- I think it is more accurate. "Save" does save my changes and closes the edit part. The link in the upper left corner does just close the edit part without saving all my changes.
- And the other view is in my opinion not a "Overview".

But having said that I can also live with "Overview"/"Übersicht". :)

Issue #1073, opened 2025-01-15
Tobira integration: Improve UX for editing the Tobira page that contains an Opencast series


dagraf:
At the moment, when I hit on the "pen" icon in the "Tobira" tab of a Opencast series that is mounted into a page in Tobira ... 

<img width="887" alt="Bildschirmfoto 2025-01-15 um 17 32 47" src="https://github.com/user-attachments/assets/55752e67-aaf5-4477-8308-f0369aee5b5a" />

... the table "Select a target page" gets shown with the page in Tobira that contains this Series as 'superpage' (where I could add a subpage to):

<img width="887" alt="Bildschirmfoto 2025-01-15 um 17 33 39" src="https://github.com/user-attachments/assets/2ac8c055-54d2-40b0-a400-9f30061eb06a" />

This is confusing. Because for example, when I first insert a series into a page I never see this view. I only see the view of the 'superpage' where I insert the subpage into which i mount my series.

Therefore, I suggest a slightliy altered behavior:
1. When clicking on the "Pen" icon the "Select a target page" gets opened at the position of the page that contains the subpage where my series I am working on is mounted (in screenshot: "Homepage").
2. Additionally, the subpage where my series is momentarily mounted gets indicated with a tick in the tickbox (see mockup).

<img width="623" alt="Bildschirmfoto 2025-01-15 um 17 43 57" src="https://github.com/user-attachments/assets/2de6c8b7-c985-4291-8bf7-7a74f72ec33c" />

owi92:
The suggested behaviour was implemented in #1078

Issue #1071, opened 2025-01-15
Tobira integration: Improve UX for specifying subpage titles when adding new subpages


dagraf:
When I create a new subpage in the Tobira tab I can specify a title. But when I insert my Opencast series into this subpage the title of the series replaces automatically this title I specified. I get informed about this in a message (_"The title of the selected page will be overridden with the title of the series."_) that appears pretty far away from where I can specify the title (see screenshot). This is far from optimal in my opinion since users might easily overlook this message and end up frustrated because their deliberately chosen title gets ignored.

Therefore, I suggest a slightliy altered behavior of the form (see also [screencast](https://tube.switch.ch/videos/nbB6MCJzGG)):
1. When I check the checkbox which indicates that I want to mount my series into this subpage I can no longer specify the title of this subpage.
2. Additionally, the textbox for the title of the subpage not only switches to read only but contains also an explanation like "Title of the mounted series" or contains the actual title of the Series (e.g. "All my best cat videos"). 

<img width="708" alt="Bildschirmfoto 2025-01-15 um 17 09 29" src="https://github.com/user-attachments/assets/b925d7b6-cc07-4982-87d6-3d361fefa027" />

Issue #1067, opened 2025-01-15
Tobira integration: Message "Copied link to clipboard." does not get shown


dagraf:
In the "Tobira" tab: When I hit on the "Copy" icon the message "Copied link to clipboard." should get shown right above this button. At the moment, it does not get shown.

Remark: The message seems to be triggered but not shown in the tab "Tobira". Because when I switch to a different tab (e.g., "Access policy") the message appears.

owi92:
Fixed in #1078

PR #1078, opened 2025-01-17
Fix some Tobira integration issues

owi92:
Fixes https://github.com/opencast/opencast-admin-interface/issues/1065, https://github.com/opencast/opencast-admin-interface/issues/1067, https://github.com/opencast/opencast-admin-interface/issues/1071 and https://github.com/opencast/opencast-admin-interface/issues/1073,
and also another small issue I found while doing this, where `not found` errors weren't being reset and their notifications didn't show up in the right place.

For completeness: I also addressed https://github.com/opencast/opencast-admin-interface/issues/1064 and https://github.com/opencast/opencast-admin-interface/issues/1068. These are related to german translations, which I updated in the crowdin project yesterday.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
@@ -126,8 +126,8 @@ const DetailsTobiraTab = ({ kind, id }: DetailsTobiraTabProps) => {
 			{tabHierarchy === "edit-path" && <EventDetailsTabHierarchyNavigation
 				openSubTab={openSubTab}
 				hierarchyDepth={0}
-				translationKey0={"EVENTS.SERIES.DETAILS.TOBIRA.SHOW_PAGES"}
-				subTabArgument0={"main"}
+				translationKey0="EVENTS.SERIES.DETAILS.TOBIRA.DISCARD"
+				subTabArgument0="main"
 			/>}
 			{tabHierarchy === "main" && <div className="modal-body">
 				{/* Notifications */}
--- a/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
@@ -78,7 +78,7 @@ const DetailsTobiraTab = ({ kind, id }: DetailsTobiraTabProps) => {
 			dispatch(addNotification({
 				type: "info",
 				key: "TOBIRA_COPIED_DIRECT_LINK",
-				duration: 3000,
+				duration: 3,
 				parameter: undefined,
 				context: NOTIFICATION_CONTEXT_TOBIRA,
 			}));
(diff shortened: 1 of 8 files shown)


Arnei (changes requested):
The "Copied link to clipboard." notification never disappears. I think it should.

The notification in the screenshot below also does not go away and thus ends up in places it was never intended to go.
Bildschirmfoto vom 2025-02-04 11-32-01

Maybe there should be an explanation for what the check marks are for? To me as a Tobira layman, they are unintuitive. I also don't understand why I need to enter a path segment before I can click the check mark.
Bildschirmfoto vom 2025-02-04 11-33-35

src/i18n/org/opencastproject/adminui/languages/lang-de_DE.json
```diff
@@ -1139,6 +1139,7 @@
 					"DESCRIPTION": "Hier können Sie auswählen, wo die Serie im Videoportal erscheinen soll. Sie können hier nur eine Seite auswählen und diese muss zudem komplett leer sein (fett markiert).\nNeue Seiten werden nur dann effektiv erstellt, wenn Sie die Serie in diese neue Seite einfügen oder wenn Sie die Serie in eine Unterseite der erstellten Seite einfügen.",
 					"SELECT_PAGE": "Zielseite auswählen",
 					"PAGE_TITLE": "Titel der Unterseite",
+					"TITLE_OF_SERIES": "(Titel der serie)",
```
Arnei, line 1142: Reminder: All changes made in translation files that are NOT in `lang-en_US.json` will be lost. All translations have to be added through crowdin.

owi92: Thanks for the reminder. We also talked about this in the technical meeting, I just forgot to remove this line.
I added the other missing german translations (https://github.com/opencast/opencast-admin-interface/issues/1064) in crowdin.
I will add this one once this PR is merged.

owi92:
> The "Copied link to clipboard." notification never disappears. I think it should.

I just realized something. All tobira notifications were set to `duration: 3000` before I started working on this.
But the duration is multiplied by 1000 in the notification slice.

So yeah, we shouldn't use `duration: 3000` unless we want that notification to stick around for 50 minutes.

owi92:
> Maybe there should be an explanation for what the check marks are for? To me as a Tobira layman, they are unintuitive.

David suggested adding these checkmarks because he said basically the same about having to click the title of a page to select it. I can try to formulate some explanation and add it to the explainer text above the table, I'm not sure were else to put it but I am open for suggestions.

> I also don't understand why I need to enter a path segment before I can click the check mark.

Clicking the check mark selects the path to mount the series in. But you can't mount a series without a path. Maybe adding an explanation, as you suggested, might also help to clarify this behaviour.

Arnei (changes requested):
Notification font size changed I think? It should not be this big I think
Bildschirmfoto vom 2025-02-04 16-21-25

This notification is now disappearing. Maybe it should not until the issue is resolved?
Bildschirmfoto vom 2025-02-04 16-25-19

owi92:
> Notification font size changed I think? It should not be this big I think

I don't think it did, though I agree that it looks kinda large. But I didn't touch that part and dev tools tells me it should be 14px, like the other notifications as well.

> This notification is now disappearing. Maybe it should not until the issue is resolved?

Good point, I changed it accordingly.

Arnei (approved):

Issue #982, opened 2024-11-19
UI should prompt user if you change workflow parameters and leave without saving


lkiesow:
The user interface should prompt user if you change workflow parameters and leave without saving.
It does that in several other places and I would expect the interface to behave similar here.

PR #1081, opened 2025-01-21
Add unsaved changes warning in event details

Arnei:
Adds a confirmation modal that warns about unsaved changes to the following tabs of the event details:
Metadata,
Extended Metadata,
Scheduling (for scheduled events),
Workflows (for scheduled events).
Thereby this fixes https://github.com/opencast/opencast-admin-interface/issues/982.

Includes https://github.com/opencast/opencast-admin-interface/pull/1046, because I was expecting this to be tightly related to modals and did not want to write it twice. The general coding approach would be the same though.

Speaking of approach, I would like feedback on if this makes for a good approach, or if there are better alternatives I've missed. If people agree with this approach it could then be extended to other modals (e.g. series) as well.

Possible future improvements (for another PR?): Currently you are only warned about unsaved changes if you try to close the modal. However, changes are also lost if you switch between tabs (e.g. from Metadata to Assets). Maybe there should be a warning there as well?

### How to test this

Can be tested as usual. Make sure the warning only appears if you actually made changes.


src/components/events/partials/modals/EventDetailsModal.tsx
gregorydlogan, line 1: The build seems unhappy with this.

```
src/components/events/partials/modals/EventDetailsModal.tsx:62:5 - error TS2322: Type 'RefObject<FormikProps<any> | null>' is not assignable to type 'RefObject<FormikProps<any>>'.
  Type 'FormikProps<any> | null' is not assignable to type 'FormikProps<any>'.
    Type 'null' is not assignable to type 'FormikProps<any>'.
      Type 'null' is not assignable to type 'FormikSharedConfig<{}>'.

62     formikRef={formikRef}
       ~~~~~~~~~

  src/components/events/partials/modals/EventDetails.tsx:82:2
    82  formikRef: React.RefObject<FormikProps<any>>
        ~~~~~~~~~
    The expected type comes from property 'formikRef' which is declared here on type 'IntrinsicAttributes & { eventId: string; close?: (() => void) | undefined; policyChanged: boolean; setPolicyChanged: (value: boolean) => void; formikRef: RefObject<FormikProps<any>>; }'


Found 1 error in src/components/events/partials/modals/EventDetailsModal.tsx:62
```

gregorydlogan (changes requested):

src/components/events/partials/modals/EventDetailsModal.tsx
Arnei: [ac48771](https://github.com/opencast/opencast-admin-interface/pull/1081/commits/ac4877185b797cb8a2eccca7bf8d016c3543310d) should fix that.

PR #1043, opened 2025-01-06
Update react-router to 7.1.1

Arnei:
Supersedes #1035.

Updates the dependency "react-router-dom" to "react-router" (package restructuring happened).

PR #1083, opened 2025-01-23
Revert PR #1043 "Update react-router to 7.1.1"

Arnei:
Reverts PR #1043, specifically commit a4097a5f84c8d67d9774a55f4c73247d900e36d2. The PR seems to have caused problems with table filtering, which are fixed by reverting it. The issue in question is that clicking on a series name in the series table will not result in a properly filtered events table anymore. See also https://github.com/opencast/opencast-admin-interface/pull/1043#issuecomment-2607816873

### How to test this

Can be tested as usual. Make sure to run `npm install`.


owi92:
This doesn't seem to be working for me. The filter only gets applied sometimes, and when it does, it still takes a few seconds to display the right events. 

https://github.com/user-attachments/assets/d7e6ddb8-cb01-40e1-893d-dfb008a7f46f

<sub>And just as a side note for anyone else testing this branch, I had to take some extra steps before I could do so:</sub>
<sub>- before running `npm install`, I had to run `rm -rf node_modules package-lock.json`. Without doing that first, the app wouldn't start, because it `Cannot find module @rollup/rollup-darwin-arm64`. This seems to be related to https://github.com/npm/cli/issues/4828.</sub>
<sub>- After doing `rm -rf node_modules package-lock.json` followed by another round of `npm install`, the app still didn't start correctly, informing me in devtools that there is an `Uncaught Error: Could not resolve "@emotion/styled" imported by "@mui/styled-engine". Is it installed?`</sub>
<sub>- so I also had to run `npm install @emotion/styled @emotion/react`</sub>
<sub>- with that, the app will start</sub>

Arnei:
Try as I might, I cannot reproduce your flukes. Is there any additional information to be had (in the webconsole etc.)?

Filters not being applied instantly is arguably a different issue.

owi92:
I just tried again, with the same outcome. There are no failed requests or errors in the console. I am on Node.js v20.18.2, if that is of any importance.
It's probably best if another person tests this, and if they don't have any issues, I think we can ignore mine. Just reverting that commit theoretically shouldn't break anything that wasn't broken before.

Also agree that filters not being applied instantly is another issue, it's just something that becomes apparent when testing these.

Arnei:
Closing this in favor #1111.

PR #1085, opened 2025-01-27
Fix `@mui/styled`

JulianKniephoff:
This PR does two things:

* Get rid of `styled-components`
  We were using it in one place only, and while there are certainly some advantages over dynamic inline styles, we are using those pervasively as well anyway.  Plus, we pull in MUI for some stuff, which by default relies on Emotion for that stuff, so if we want those benefits, I would propose that it's at least worth considering to get them through that.
* Explicitly npm install @emotion/react @emotion/styled
  As per https://mui.com/material-ui/getting-started/installation/#default-installation.
 
  These were installed as (optional?) transitive (peer?) dependencies, but that can of course change randomly.  I've seen this happen while preparing a custom build of this repo.
 
  Note that there is also https://mui.com/material-ui/getting-started/installation/#with-styled-components, which is arguably more fitting to this project, because we were already using `styled-components`, but I opted for this solution because it seemed simpler and better supported, and also because our usage of `styled-components` isn't really justified.  However, see the previous commit for that.


Arnei (approved):
Looks reasonable to me and works.

PR #1087, opened 2025-01-28
Fix notes cannot be edited in notes column

Arnei:
In the events table there is a "Notes" column (disabled by default). If there is a note, it currently cannot be changed (changes are lost on page refresh). This patch should fix that.

### How to test this

Can be tested as usual.
- To enable the notes column, go to the "events" table, then click on "Edit" in the top right corner of the table. There you can add the notes column to the other columns. 
- Write a note on an event. Reload the page.
- Change the note you just wrote. Reload the page.
- Check if the changes persisted or not.


owi92 (approved):
This is working as described.

PR #1088, opened 2025-01-29
Fix wrong notes being displayed in notes column

Arnei:
When filtering events or switching between pages, sometimes notes from the wrong events would be displayed in the notes column. This patch should fix that.

### How to test this

Can be tested as usual.
- To enable the notes column, go to the "events" table, then click on "Edit" in the top right corner of the table. There you can add the notes column to the other columns. 
- Have at least two pages worth of events.
- Write a note on the topmost event on the first page. Write a different note on the topmost event on the second page.
- Reload the page.
- Switch between pages. Observe what is displayed in the note for the topmost event on each page.

### DIFF
--- a/src/components/events/partials/EventsNotesCell.tsx
+++ b/src/components/events/partials/EventsNotesCell.tsx
@@ -65,7 +65,7 @@ const EventsNotesCell = ({
 				</textarea>
 			}
 			{comments.map((comment, key) => (
-				<div className="comment" key={key}>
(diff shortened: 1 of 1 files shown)


owi92 (approved):
I could reproduce the bug on main and can confirm that it is fixed on this branch.

owi92:
I could not reproduce the bug you mentioned in your most recent commit, so I cannot actually confirm whether that is fixed. At least the code suggests it, so I'd say this can be merged.

Issue #1102, opened 2025-02-03
Sorting may change when adding a note in the notes column


Arnei:
To reproduce:
- Add the notes column to the events table
- Have multiple events with the same start date X
- Sort the events table by start date (the default sorting)
- Add a note to the topmost event with start date X.
- After a couple seconds, the event may now be the lowermost event with start date X.

It looks like the comment/events "jumps" to another position. This can confuse users and cause misinputs.

Arnei:
Was fixed by #1103.

PR #1103, opened 2025-02-03
Enforce event ordering

Arnei:
**WILL BREAK THE UI WITHOUT THE REQUIRED BACKEND CHANGES:** https://github.com/opencast/opencast/pull/6461

The events table is sorted by a sort parameter. By default, this is the start date. If there are multiple events with the same start date, the sorting for those is left to the elasticseach index in the backend. Elasticsearch however does not guarantee any ordering. This may result in the order of events changing unexpectedly like in #1102.

This patch aims to solve this by adding a secondary sort parameter to the events query. The primary parameter still takes priority. For the secondary parameter, `uid` was chosen as it is the only field guaranteed to be unique.

More sorting parameters can lead to increased query times. If someone with an Opencast with *many* events could test the impact of these changes on the `events.json` request times, that would be great.

### How to test this

Requires an Opencast with the PR https://github.com/opencast/opencast/pull/6461 installed.
- Sort the events table by different parameters and check if the resulting sorting make sense.
- Compare the response times in the network tab of the developer tools.
- Check if #1102 is still reproducible or not.

### DIFF
--- a/src/slices/eventSlice.ts
+++ b/src/slices/eventSlice.ts
@@ -232,6 +232,13 @@ export const fetchEvents = createAppAsyncThunk('events/fetchEvents', async (_, {
 	const state = getState();
 	let params: ReturnType<typeof getURLParams> & { getComments?: boolean } = getURLParams(state);
 
+	// Add a secondary filter to enforce order of events
+	// (Elasticsearch does not guarantee ordering)
+	params = {
+		...params,
+		sort: params.sort ? params.sort + ",uid:asc" : "uid:asc"
+	}
+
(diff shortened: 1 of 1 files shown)


owi92:
I haven't tested the response times, only the ordering. While the specific issue of #1102 seems to be fixed, I can sill observe some jumping when randomly clicking filters (see screencapture). So in regards to

> Sort the events table by different parameters and check if the resulting sorting make sense.

I wouldn't say that it does. (I tested with https://github.com/opencast/opencast/pull/6461).


https://github.com/user-attachments/assets/70f08d47-8bd9-4a98-85a1-c68320be8808

PR #1104, opened 2025-02-04
Split presenters at ";"

Arnei:
It was brought to my attention that this was a feature in the previous admin ui, so this patch adds it back in.

When adding multiple presenters to an event in the metadata tab, you can now add multiple presenters at the same time by delimiting them with a ";". For example, writing "Pacman; Duck Hunt" would result in both "Pacman" and "Duck Hunt" being added as presenters. Will also work for other metadata fields that work the same way, e.g. contributers.

### How to test this

Can be tested as usual.

### DIFF
--- a/src/components/shared/wizard/RenderMultiField.tsx
+++ b/src/components/shared/wizard/RenderMultiField.tsx
@@ -53,23 +53,26 @@ const RenderMultiField = ({
 		}
 
 		if (newInputValue !== "") {
-			// Flag if only values of collection are allowed or any value
-			if (onlyCollectionValues) {
-				// add input to formik field value if not already added and input in collection of possible values
-				if (
-					!fieldValue.find((e) => e === newInputValue) &&
-					fieldInfo.collection?.find((e) => e.value === newInputValue)
-				) {
-					fieldValue[fieldValue.length] = newInputValue;
-					form.setFieldValue(field.name, fieldValue);
-				}
-			} else {
-				// add input to formik field value if not already added
-				if (!fieldValue.find((e) => e === newInputValue)) {
-					fieldValue[fieldValue.length] = newInputValue;
-					form.setFieldValue(field.name, fieldValue);
-				}
+			const splitArray = newInputValue.split(";").map(item => item.trim()).filter(Boolean);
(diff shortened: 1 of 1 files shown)


wsmirnow:
This patch enable splitting value by an `;` on all metadata fields. It does not read the metadata catalog configuration `property.{field-id}.delimiter` value like documented in https://docs.opencast.org/r/17.x/admin/#configuration/metadata/ The side effect is, you can't use `;` any more in any metadata field in Opencast.

wsmirnow:
The delimiter configuration value will be parsed by Opencast backend and provided by metadata endpoints liek `/admin-ng/event/new/metadata`, `/admin-ng/event/EVENTID/metadata.json`, …. Example:

```json
[
    {
        "flavor": "dublincore\/episode",
        "title": "EVENTS.EVENTS.DETAILS.CATALOG.EPISODE",
        "fields": [
            {
                "readOnly": false,
                "id": "title",
                "label": "EVENTS.EVENTS.DETAILS.METADATA.TITLE",
                "type": "text",
                "value": "",
                "required": true
            },
            {
                "delimiter": ";",
                "readOnly": false,
                "id": "contributor",
                "label": "EVENTS.EVENTS.DETAILS.METADATA.CONTRIBUTORS",
                "type": "mixed_text",
                "value": [],
                "required": false
            }
        ]
    }
]
```

Arnei:
> This patch enable splitting value by an ; on all metadata fields.

I cannot reproduce that. As far as I can tell this really only affects metadata fields of the `mixed_text` type, which are "presenters" and "contributors" in the common event catalog. Can you describe in more details what is happening with the other metadata fields, any errors?

[Bildschirmaufzeichnung vom 2025-02-07 09-16-51.webm](https://github.com/user-attachments/assets/b6e85203-9088-4261-bb18-f979c82721b1)

> It does not read the metadata catalog configuration property.{field-id}.delimiter value like documented in https://docs.opencast.org/r/17.x/admin/#configuration/metadata/

I was not aware that this was configurable, thanks for telling me! I will make sure the new admin ui respects the configuration.

wsmirnow (approved):
Just works. Thank you @Arnei

PR #1105, opened 2025-02-04
Fix error when changing date in bulk edit metadata

### LINKED ISSUE #1099: Bug: Metadata bulk-editing crashes the app when `start date` field is clicked
How to reproduce:
- Select more that one event
- Open the `Edit the metadata of multiple events` dialog via the `Actions` menu
- Click inside the empty field of the `start date` row

This will lead to the app crashing with an `Uncaught RangeError: Invalid time value` error shown in the dev tools console.

Arnei:
Fixes #1099.

Trying to change the start date for two or more events at the same time would cause the app to crash (See #1099). This fixes that.

### How to test this

Can be tested as usual. "Bulk edit metadata" can be found in the "Actions" dropdown in the events table. Make sure to select at least two events via the checkmarks on the left side of the table.


geichelberger (approved):
LGTM, no blank screen anymore :tada:

PR #1111, opened 2025-02-17
Fix series filter not being applied to events

Arnei:
Includes https://github.com/opencast/opencast-admin-interface/pull/980, supersedes  https://github.com/opencast/opencast-admin-interface/pull/1083.

Should achieve the same thing as https://github.com/opencast/opencast-admin-interface/pull/1083, but without reverting to older dependency versions. Meaning clicking on a series name in the series table should bring up the events table with the series filter set. Including https://github.com/opencast/opencast-admin-interface/pull/980 as it helps in achieving this.

### How to test this

Can be tested as usual. Ideally #980 is tested and merged beforehand.


geichelberger (changes requested):
It would be nice to clean up the commit history and eliminate the merge commits.

geichelberger (approved):
LGTM

PR #1114, opened 2025-02-21
Reduce duplicate code for table action dropdown

Arnei:
Creates a new component for table action dropdowns. This should result in more readable code and make it easier to
introduce the component into other tables, should we ever need to.

### How to test this

Can be tested as usual. The delete modal not opening is fixed by https://github.com/opencast/opencast-admin-interface/pull/1115.


ferishili (changes requested):
Thanks @Arnei,
Unfortunately, the delete action from the dropdown does not work properly.
However codewise looks good!

geichelberger (changes requested):

ferishili (approved):
Thanks @Arnei, LGTM

PR #1116, opened 2025-02-21
Fix unique key error in NavBar

Arnei:
NavBar did not correctly key its links. This patch fixes that.

### How to test this

Can be tested as usual. Observe if the unique key error in the webconsole disappears.


geichelberger (approved):
LGTM

PR #1118, opened 2025-02-24
Switch to @opencast/eslint-config-ts-react

### LINKED ISSUE #713: Change ESLint configuration

Arnei:
Fixes #713.

Thanks to @ LukasKalbertodt , our Opencast eslint config for typescript-react is now compatible with eslint 9.
High time to not only update to eslint 9, but also to replace our old eslint config with @opencast/eslint-config-ts-react. The  old config was a default  config provided by CreateReactApp, which is not maintained anymore and we have no other reason to keep it around.

That said, this patch only aims to update dependencies and introduce @opencast/eslint-config-ts-react. It does not aim to fulfill the numerous rules the project is now breaking, which is why most of those rules are disabled. They can be gradually enabeled by future patches, when they won't cause massive merge conflicts for the larger outstanding pull requests like #1006 or #1059.

Update: Enabled rules which we were mostly following everywhere already and which required only changing a few lines.

### How to test this

Check out the code, break some rules and see eslint complaining about it during build (or in your IDE if so configured).


owi92 (approved):
lgtm, though I can't get it to run in my editor, only manually. But that's a `me problem`™.

PR #1121, opened 2025-02-25
Reduce api requests

### LINKED ISSUE #940: Applying a "series" filter takes a long time
When I go to Series and select a series, it takes a long time (>5 sec) for the filter to be applied to the events:

https://github.com/user-attachments/assets/5cb33e31-1b1c-483c-baa0-a8b4f3404f97

Arnei:
Reduce the overall number of api requests we send to the backend when loading a new page/table. In particular, we were sending the same requests twice or more which should not be necessary. This cuts down on the number of requests in an attempt to reduce overall loading times. The guiding principle was that each component should be responsible for fetching its own data.

Should fix #940.

### How to test this

Can be tested as usual. Switch between tables and modify the tables (filters, change page, change columns). Also check the requests in the network tab of your browser. (Note that if you test in development mode, due to React StrictMode being enabled requests will still be send twice).


snoesberger:
I wanted to check the performance with this PR in our test environment, but unfortunately it looks like you need Opencast version 17 for this PR to work (we currently have version 16 installed).

However, my local tests of this PR, connected to https://stable.opencast.org, with various filters, switches between tabs etc. were successful. I also wanted to compare the performance when selecting a series in the series tab (as I did on https://github.com/opencast/opencast-admin-interface/issues/940), but this wasn't possible as the series filter on https://stable.opencast.org doesn't work at all at the moment.

ferishili (changes requested):
Thanks @Arnei
Apart from the single code-related issue/suggestion, I found a few **functional issues** as follows:  

### Sorting issues:
#### Event Table
- **Date** sorting does not work.  
- **Start Time** sorting does not work.  
- **End Time** sorting does not work.  
- **Presenters** sorting does not work.  
- **Status (PublishCell)** throws an error.  

Screenshot 2025-03-13 at 14 13 13

#### Series Table
- **Organizer(s)** sorting returns event 400 error!
#### Servers Table
- **Job completed** sorting returns 400 error!

### HTML Content Issue:
A dev-tool error appears: 
`In HTML, <button> cannot be a descendant of <button>.`
This occurs due to two closing `</button>` elements, specifically in:
https://github.com/opencast/opencast-admin-interface/blob/ffde25bcd82d8aefc485acbdcbb20fb0b9ef6293/src/components/shared/Table.tsx#L307

This happens when selecting the table items num from the dropdown!

### Bulk Checkbox Selector Issue
When the Events page first loads, the bulk select checkboxes are missing.
You need to switch between "Series" and "Events" for them to appear.

Some of the issues mentioned above might not be directly relevant to this PR. However, I am highlighting them here because I encountered them while performing **functional testing** on this PR.

Arnei:
Thanks for the extensive testing.

> [...] sorting does not work

That should be fixed by backend PR https://github.com/opencast/opencast/pull/6501. While the PR is already merged, I suppose it has not quite ended up on develop.opencast.org yet.

> Status (PublishCell) throws an error.

Is hopefully addressed by latest commit [d474710](https://github.com/opencast/opencast-admin-interface/pull/1121/commits/d474710fc78984da91d7f0d36d0958e7d2f4a670). If that is not the case, please described what exactly you did to provoke the error message. Please also describe if anything bad is happening in the ui.

> [...] sorting returns event 400 error!

Good catch! As far as I can tell at a glance, these seem to be backend issues not caused by this PR. I'll open tickets for them.

> HTML Content Issue: [...]

Also unrelated to this PR it seems, I will create a ticket for it.

> Bulk Checkbox Selector Issue

Fixed by #1139 and #1150.

ferishili (changes requested):
I can confirm that everything works fine now.
However, sorting in the **Series Table** and **Servers Table** still returns a **400 error**. I understand that this is more of a backend, but I still wanted to highlight it.

ferishili (approved):
In this case, I would say this PR is **good to go!**
**P.S.:** Apologies for my previous mistaken review!

PR #1122, opened 2025-02-25
Fix hiding tobira tab in series create modal

Arnei:
In the series create modal, we query ´/admin-ng/series/new/tobira/page` to check if Tobira is available. However, if the user does not have the role required for that endpoint (ROLE_UI_SERIES_DETAILS_TOBIRA_EDIT), the endpoint will return an error html page which we are not expecting, causing the ui to crash.
This fixes the problem by checking if the user has the required role to make the request to the endpoint. If the user does not have the role, the tobira tab is then not shown.

### How to test this

Ideally test without but also with a configured Tobira, and with a user who has tobira roles and who does not have tobira roles.


owi92 (approved):
Thank you. This works for the suggested test cases.

I wonder if we want or need to add a similar check to the series details modal, since that also allows editing and removing paths in Tobira (just tested, that is hidden anyway). 
I think that should probably only hide the editing part though, and still show the direct link.

That might get a little tricky and doesn't need to be part of this PR, as that at least isn't causing any crashes.

Issue #1079, opened 2025-01-17
Date Picker: on page load, Start Date Filter, date picker is displayed


marcusm7:
Steps to reproduce/ How to test:
- open admin-ui 
- add filter
- select start-date filter
- reload the page
- the date-picker is displayed

Issue:
- the date-picker dialog for the start date is displayed without user interaction. This shouldn't happen.
![Image](https://github.com/user-attachments/assets/ff9acb08-47aa-4e7d-89d5-7b1ffeaafed6)

Environment:
- KDE 6
- Firefox 134 or Chromium
- Opencast: https://github.com/opencast/opencast-admin-interface/pull/1019

Note:
- Deleting the browser cache + closing + reopening the browser "fixes" the issue *once* 
- The issue is described here https://github.com/opencast/opencast-admin-interface/pull/1019#issuecomment-2597723942

Arnei:
Fixed by #1525

PR #1132, opened 2025-03-01
Remove filter selection if no value is selected

eduardklinger:
When a filter was selected, but no value was chosen, the filter dropdown or datepicker was unexpectedly opening on a page reload.

This fixes this issue including #1079


Arnei (changes requested):
This seems to somehow break the feature where we remember set filters across tab switches (see #980).

Arnei:
Superceded by #1525

PR #1134, opened 2025-03-03
Convert indentation to spaces

geichelberger:
I don't know if it's worth to screwup git blame.


Arnei:
> I don't know if it's worth to screwup git blame.

I think so, was planning to do this myself at some point. Mostly wanted to get other PRs that change many lines, like  #1059, get merged first.

PR #1139, opened 2025-03-11
Restore table checkboxes on load

### LINKED ISSUE #1136: Events table checkboxes have disappeared
Upon initial UI load the checkboxes in the events table are gone (missing from the DOM).  Navigating to the series table, then back to events restores them.

I bisected the changes, and it looks like 26d7417fbbd8f2e0c9dca10716378b63157055bc is the culprit, though I don't know enough typescript to undo the issue.

Arnei:
Fixes #1136.
Table checkboxes were not rendered upon initial load. This patch fixes that.

### How to test this

Can be tested as usual. Tables should again always render the selection checkboxes (if applicable for that table). The reset button in the edit table view modal should also continue to work.

### DIFF
--- a/src/slices/tableSlice.ts
+++ b/src/slices/tableSlice.ts
@@ -70,7 +70,7 @@ export function isSeries(row: Row | Event | Series | Recording | Server | Job | 
 // TODO: Improve row typing. While this somewhat correctly reflects the current state of our code, it is rather annoying to work with.
 export type Row = { selected: boolean } & ( Event | Series | Recording | Server | Job | Service | User | Group | AclResult | ThemeDetailsType )
 
-export type Resource = "events" | "series" | "recordings" | "jobs" | "servers" | "services" | "users" | "groups" | "acls" | "themes"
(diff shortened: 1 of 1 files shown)


NadiaUoM (commented):
Just a small question. Why you have added " | " on line no. 73? Because without adding this still checkbox are getting loaded.

eduardklinger:
> Just a small question. Why you have added " | " on line no. 73? Because without adding this still checkbox are getting loaded.

This added an empty string ("") to the types, because the default value was changed to `""`, which was not included in `type Resource`. The `|`  just connects the empty string with the rest of the string values.

eduardklinger (approved):
The default resource value `"events"` prevented updating the multiSelect value on the first events table load. 

Setting the default value to `""` triggers updating the `multiSelect` value, which renders the checkboxes.

Arnei:
Arguably we should just make sure that `multiSelect` is set correctly on first table load, but this also works and removes the assumption that "events" is the first table we load, so *shrug*.

Thanks for reviewing!

NadiaUoM (approved):

PR #1140, opened 2025-03-11
Fix cancel button in column selection modal

Arnei:
The cancel button in the modal where the user can select which columns should be displayed in a table was broken by 26d7417fbbd8f2e0c9dca10716378b63157055bc.
This patch fixes that.

### How to test this
Can be tested as usual.


NadiaUoM (approved):
Checked the functionality is working as expected now. Tested other functionality as well save, reset, drag and drop most of the permutation combination.

PR #1141, opened 2025-03-11
Fix event/series modal not wide enough

Arnei:
The tabs in the event/series details modals were wrapping around in case there were too many of them, due to the of the container being too small. This fixes that by applying the correct class id for those modals again, which was lost due to multiple PRs.

Arguably the fixed widths here are the true problem and we should look into making our modal css more dynamic, but that's a task for another PR.

Screenshot of the bug:
Unbenannt

### How to test this
Cause as many tabs to appear in the event details as possible by
- Schedule the event
- Configure Extended Metadata
- Configure Tobira


owi92 (approved):
lgtm. Also did a quick test.

PR #1145, opened 2025-03-11
Fix create modal strings

Arnei:
Fixes the translation string for the header of the create modals for themes and users.

### How to test this.
Themes need to be enabled in your Opencast in etc/org.opencastproject.organization-mh_default_org.cfg.


NadiaUoM (approved):
Theme
User

PR #1146, opened 2025-03-11
Fix themes not being fetched

Arnei:
The fetch request for populating the themes table was broken. This patch fixes that, so themes should show up in the table again.

Also somewhat tightens our typing to prevent silly mistakes like this in the future.

### How to test this
Themes need to be enabled in your Opencast in etc/org.opencastproject.organization-mh_default_org.cfg


NadiaUoM (approved):
Theme

PR #1150, opened 2025-03-13
Fix typescript conflict for initial state of table field "resource"

### LINKED ISSUE #1149: Main does not build
Main does not build after latest merges.

```
src/slices/tableSlice.ts:100:2 - error TS2741: Property '""' is missing in type '{ events: string; series: string; recordings: string; jobs: string; servers: string; services: string; users: string; groups: string; acls: string; themes: string; }' but required in type '{ "": string; events: string; series: string; themes: string; recordings: string; jobs: string; servers: string; services: string; users: string; groups: string; acls: string; }'.

100  sortBy: {
     ~~~~~~

  src/slices/tableSlice.ts:84:2
    84  sortBy: { [key in Resource]: string },  // Key is resource, value is actual sorting parameter
        ~~~~~~
    The expected type comes from property 'sortBy' which is declared here on type 'TableState'

src/slices/tableSlice.ts:113:2 - error TS2741: Property '""' is missing in type '{ events: "DESC"; series: "DESC"; recordings: "ASC"; jobs: "ASC"; servers: "ASC"; services: "ASC"; users: "ASC"; groups: "ASC"; acls: "ASC"; themes: "ASC"; }' but required in type '{ "": ReverseOptions; events: ReverseOptions; series: ReverseOptions; themes: ReverseOptions; recordings: ReverseOptions; ... 5 more ...; acls: ReverseOptions; }'.

113  reverse: {
     ~~~~~~~

  src/slices/tableSlice.ts:86:2
    86  reverse: { [key in Resource]: ReverseOptions },  // Key is resource, value is actual sorting parameter
        ~~~~~~~
    The expected type comes from property 'reverse' which is declared here on type 'TableState'


Found 2 errors in the same file, starting at: src/slices/tableSlice.ts
```

Arnei:
Fixes #1149.
Fixes a typescript error that was caused by adding the empty string to a string enumeration. Should not change behaviour.

This patch fixes that removing the empty string from the type and fixing reason the empty string was introduced in the first place in a different manner (also see commit 28aa9821aac65f8acea9fd2555d87125bc969ed3).

This has the downside of forcing us to set "resource" to a concrete value, which is technically wrong.
However, I find this preferable to allowing "resource" to be undefined. It makes our typing tighter and easier to handle. Furthermore, "resource" must be set on table load anyway, so the technically wrong value gets overwritten asap.

### How to test this
Can be tested as usual. Make sure checkboxes appear in the leftmost table column if they should, and not appear if they should not. Also make sure that tables look correctly when refreshing the page (although that might not be the case due to unrelated reasons which should be fixed by #1121).

### DIFF
--- a/src/components/shared/Table.tsx
+++ b/src/components/shared/Table.tsx
@@ -1,6 +1,7 @@
 import React, { JSX, useEffect, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 import {
+	getMultiSelect,
 	getPageOffset,
 	getTable,
 	getTableDirection,
@@ -60,6 +61,7 @@ const Table = ({
 	const rows = useAppSelector(state => getTableRows(state));
 	const sortBy = useAppSelector(state => getTableSorting(state));
 	const reverse = useAppSelector(state => getTableDirection(state));
+	const multiSelect = useAppSelector(state => getMultiSelect(state));
 
 	// Size options for pagination
 	const sizeOptions = [10, 20, 50, 100, 1000];
@@ -179,7 +181,7 @@ const Table = ({
 				<thead>
 					<tr>
 						{/* Only show if multiple selection is possible */}
-						{table.multiSelect ? (
+						{multiSelect ? (
 							<th className="small">
 								{/*Checkbox to select all rows*/}
 								<input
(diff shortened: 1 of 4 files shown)


owi92 (approved):
lgtm, the approach also makes sense imo. Checkboxes are well behaved in my testing.

PR #1153, opened 2025-03-14
Fix being unable to step back in event modal

Arnei:
Fixes a bug occurring in the create event modal. When scheduling an event, going back from the "proccesing" step to the "source" step was not possible if the "asset upload" step was also configured.

Bildschirmfoto_2025-03-14_um_04 48 57

### How to test this
You will need an Opencast with asset upload options configured (see also https://docs.opencast.org/develop/admin/#configuration/admin-ui/asset-upload/).
Or cheat by checking out the code and setting `hidden` in line https://github.com/opencast/opencast-admin-interface/blob/57e21b1a7f3d272df09a68e19f1345055dcab2e5/src/components/events/partials/wizards/NewEventWizard.tsx#L86 to false.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/NewAssetUploadPage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/NewAssetUploadPage.tsx
@@ -33,12 +33,6 @@ const NewAssetUploadPage = <T extends RequiredFormProps>({
 		(asset) => asset.type !== "track"
 	);
 
-	// if user not chose upload in step before, the skip this step
-	if (formik.values.sourceMode !== "UPLOAD") {
-		nextPage(formik.values);
-		return null;
-	}
-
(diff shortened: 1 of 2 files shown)


NadiaUoM:
I try to reproduce this issue with the main branch code but couldn't do it? can you please share the video if possible?

Arnei:
I forgot to mention that in the "source" step you will have to schedule an event. (To be able to schedule an event, a capture agent needs to be configured. You can easily configure a dummy capture agent through the Opencast endpoint /capture-admin/agents/{name}).

Here's a video

[Bildschirmaufzeichnung vom 2025-04-28 09-15-05.webm](https://github.com/user-attachments/assets/2efd7262-38ff-4bbd-95cf-ec431824c11d)

NadiaUoM (approved):

NadiaUoM:
Thank you for the inputs...

PR #1158, opened 2025-03-17
Fix end time date display for scheduling

Arnei:
Fixes display behaviour for the end time date display on the scheduling pages to go back to how it was before.
Namely the end time date is now again displayed in the same row as the end time, and is only actually displayed if the end time is "before" the start time (at which the ui assumes that the end time has to be on the next day).

Bildschirmfoto vom 2025-03-17 11-37-10

### How to test this
Can be tested as usual. Should only affect the scheduling pages in the create new event modal and event details modal.


Piriya-VM (approved):
Works for me. :)

Bildschirmfoto vom 2025-05-13 16-06-17
Bildschirmfoto vom 2025-05-13 16-13-16
No code review!

PR #1159, opened 2025-03-18
Fix layouting errors in a various modals

### LINKED ISSUE #1152: CSS for user details role tab is broken
Should look like

Image

but looks like

Image

I'm assuming #1046 to be the culprit and that this can be resolved similar to #1141, but I have not actually looked into it yet.

Arnei:
Fixes css problems in the theme modals, the user modals and probably various others.

Before:
Bildschirmfoto vom 2025-03-18 08-56-50

After:
Bildschirmfoto vom 2025-03-18 08-56-29

Fixes #1152.

### DIFF
--- a/src/styles/mixins/_mixins-bourbon.scss
+++ b/src/styles/mixins/_mixins-bourbon.scss
@@ -1,7 +1,7 @@
 // Bourbon leftovers
 
 @mixin clearfix {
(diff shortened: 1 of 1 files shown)


NadiaUoM (approved):
Working as expected verify User, Group, Access policy and Theme.

PR #1162, opened 2025-03-18
Limit concurrency for deploy and remove test page

geichelberger:
If the test page repository is processed simultaneously, the fastest pushed wins. To prevent this, parallel processing is restricted. The processing order is not guaranteed; thus, a page could be removed before deployment.

https://github.com/opencast/opencast-admin-interface/actions/runs/13932498328/job/38992836228


gregorydlogan:
Looks right?  Guess we'll see.  The last few looked right too :D

PR #1171, opened 2025-03-31
Fix visual errors and path validation in Tobira tab

### LINKED ISSUE #1170: Tobira tab has some graphical errors
I think these were introduced by recent css changes in the project:

- This "button" used to be an icon:
Image

- The footer with the save button shouldn't be "floating" like this:
Image
It used to look like this:

Image

I know that the user base for these features isn't particularly large (also, Bern and ETH both use a custom build afaik), but it would be nice if we can get this fixed soon, preferably before this is added in their builds. I can take care of this on monday.

owi92:
This restores the previous styling of the "edit path" button as well as the footer containing the "save path" button (for screenshots of the broken state, see the linked issue).

Also fixes a small error with the path validation that I overlooked before.

Resolves #1170

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/NewTobiraPage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/NewTobiraPage.tsx
@@ -364,14 +364,14 @@ const NewTobiraPage = <T extends TobiraFormProps>({
 					</p>
 					{!mode.edit && <p style={{ fontSize: 12 }}>{t("EVENTS.SERIES.NEW.TOBIRA.DIRECT_LINK")}</p>}
 				</>}
-			{/* Render buttons for saving or resetting updated path */}
-			{mode.edit && <SaveEditFooter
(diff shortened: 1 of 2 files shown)


src/components/events/partials/ModalTabsAndPages/NewTobiraPage.tsx
```diff
@@ -365,13 +364,13 @@ const NewTobiraPage = <T extends TobiraFormProps>({
 					{!mode.edit && <p style={{ fontSize: 12 }}>{t("EVENTS.SERIES.NEW.TOBIRA.DIRECT_LINK")}</p>}
 				</>}
 			{/* Render buttons for saving or resetting updated path */}
```
Arnei, line 366: Forgot to move the comment with the rest of the code

Arnei (changes requested):
I apologize, those visual errors are most likely my fault. Changes look reasonable, one nitpick below.

Arnei (approved):

PR #1178, opened 2025-04-08
Show service online status

### LINKED ISSUE #1166: Show when a service is offline
Not a bug per se, but while the error state of services is indicated clearly in the Admin UI, whether the service is online is not visible at all. This can lead to confusion when jobs can't be dispatched because a service is marked as offline, but the services table doesn't show anything abnormal. It would be great if this could be seen at a glance.

Arnei:
**Requires https://github.com/opencast/opencast/pull/6632 to function.**

Fixes #1166.

Adds a column to the services table that displays the online status of a service using traffic lights, much like the respective column in the servers table.
This change should make it easier to reason about why certain jobs may not able to be dispatched.

Notably, an offline service does not necessarily mean that there is an error or even a problem. Therefore this patch does not add a notification about services being offline.

Bildschirmfoto vom 2025-04-08 12-04-43

### How to test this

- Have an Opencast 17 or newer with https://github.com/opencast/opencast/pull/6632 installed. Ideally have a distributed setup as well (no AllInOne).
- Force services to be offline or in maintenance somehow. The latter can be achieved through the servers page in the ui.

### DIFF
--- a/src/configs/tableConfigs/serversTableConfig.ts
+++ b/src/configs/tableConfigs/serversTableConfig.ts
@@ -16,7 +16,6 @@ export const serversTableConfig: TableConfig = {
 			template: "ServersStatusCell",
 			name: "online",
 			label: "SYSTEMS.SERVERS.TABLE.STATUS",
-			sortable: true,
 		},
 		{
 			name: "hostname",
--- a/src/configs/tableConfigs/servicesTableConfig.ts
+++ b/src/configs/tableConfigs/servicesTableConfig.ts
@@ -13,6 +13,11 @@ import { TableConfig } from "./aclsTableConfig";
 export const servicesTableConfig: TableConfig = {
 	columns: [
 		{
+			template: "ServersStatusCell",
+			name: "online",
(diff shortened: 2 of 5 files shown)


gregorydlogan:
For easier testing, even on an all in one set `heartbeat.interval=0` in `org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg`, then use `update oc_host_registration set online=0;` and/or `update oc_host_registration set maintenance=1;` to make hosts be offline and/or in maintenance.

This is *really* nitpicky, but should we center the traffic light?  I feel like it looks odd being that far over, even though everything else is left adjusted.


As is             | |  Centered
:-------------------------:|:-:|:------------------------:
image  | |  image

Arnei:
> This is really nitpicky, but should we center the traffic light? I feel like it looks odd being that far over, even though everything else is left adjusted.

I'm slightly in favor of keeping it left, just because that keeps it the same as the other columns and does not make it look so lost in case the column is wider than min-width. But I also do not care all that much.

wsmirnow (approved):
Works for me

gregorydlogan:
Fine by me

PR #1186, opened 2025-04-29
Improve word wrapping in tables

lkiesow:
Many tables in the admin interface cannot handle events or series with longer texts, either wrapping them at weird places or even breaking the user interface completely.

This patch tries to improve the situation and allow for better wrapping of table content.

Before:
Screenshot from 2025-04-29 21-50-59

After:
Screenshot from 2025-04-29 21-49-52


dennis531 (approved):
Overall, it looks good!

PR #1196, opened 2025-05-07
Uncheck 'Select all' header cell on page change

### LINKED ISSUE #1193: Dismiss selection if table content changes
If you select all items on a page like the event details and the content changes since you select another page, the “select all” checkbox should no longer be selected.

Select all events:
Image

Switch to page 2; the “select all” checkbox is still selected:
Image

klaraloreen:
Fixes: #1193.

Adds a state variable to control the value of the "Select all" header cell on page change.

Screenshot before issue (page just changed):
Screenshot from 2025-05-07 09-24-31

Screenshot after issue (page just changed):
Screenshot from 2025-05-07 09-43-23


lkiesow:
Looks good for pagination. But I found another problem:
- The select all checkbox keeps being selected after starting an action (Select all → Action → Start Task → e.g. run „Republish Metadata“)

ferishili:
Hi @klaraloreen, thanks for your work.
I like the idea of providing the state val and checked props, however this would (I assume) only uncheck the header as written in the title. But we should also cover the deselection of the selected items, so I introduce the mechanism (in a method) here: https://github.com/opencast/opencast-admin-interface/pull/1286
There, I use ref to target the header checkbox too1
So, I would recommend going as follows:
- Wait for #1286 to be merged first & rebase to this PR.
- No using state & checked props! in favor of reusing the method in sync mode!
- Use the deselect method also in place related to pagination!

Thanks & BR

klaraloreen:
Hi @ferishili, thanks for your improvements! I removed the state variable and used the deselect method instead, just like you suggested.

ferishili:
Thanks, @klaraloreen,
it works as expected, however, it would also make sense to force that deselection on the page number dropdown.
What do you think?

ferishili (changes requested):
@klaraloreen, please consider adding `forceDeseletAll()` in this method as well:
https://github.com/opencast/opencast-admin-interface/blob/c0402b87b70ba06513ce45809c34c672eb4e75dd/src/components/shared/Table.tsx#L115

ferishili (approved):
Thanks @klaraloreen

Issue #1195, opened 2025-05-05
“Add User” should be a wizzard


lkiesow:
If you open the “Add User” dialog, you can create the user once you finish the first page. This should be a wizard and the button should lead a user to the second page where you can add roles for the user. I've clicked and see others click on the “Create” button countless times, expecting that it's a “next” button for that wizard.

![Image](https://github.com/user-attachments/assets/a6cc97e5-7dd0-4e87-a6ee-af1a5959e957)

priya-tik:
Hi @lkiesow, I saw this issue and worked on it. The "Add User" dialog now behaves like a wizard, allowing users to proceed to the next step (roles) before final submission.

I don’t have permission to assign the issue to myself, but feel free to do so if needed. Let me know if any adjustments are required.

Thanks!

PR #1197, opened 2025-05-07
Fix: Make Add User dialog a multi-step wizard

priya-tik:
This PR refactors the "Add User" dialog to behave like a multi-step wizard, improving the user experience by splitting user details and role assignment into separate tabs. This change ensures better usability and mirrors the design of similar multi-step forms in the interface.

Note:
Please let me know if any additional changes or improvements are needed!

### DIFF
--- a/src/components/users/partials/wizard/NewUserWizard.tsx
+++ b/src/components/users/partials/wizard/NewUserWizard.tsx
@@ -83,12 +83,25 @@ const NewUserWizard = ({
 
 							{/* Navigation buttons and validation */}
 							<WizardNavigationButtons
-								isLast
-								formik={formik}
-								nextPage={() => formik.handleSubmit()}
-								previousPage={() => close()}
-								cancelTranslationString={"CANCEL"}
-							/>
+							isLast={tab === 1}
+							formik={formik}
+							nextPage={() => {
+							if (tab < 1) {
+							setTab(tab + 1);
+							} else {
+								formik.handleSubmit();
+								}
+						}}
+							previousPage={() => {
+							if (tab > 0) {
+							setTab(tab - 1);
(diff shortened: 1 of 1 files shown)


package-lock.json
JamesUoM, line 1: Don't include this unless you've changed package.json

JamesUoM (changes requested):
Though the Next button functions as expected it doesn't change the look& feel to that of the Wizard Model i.e steps instead of tabs. 
image

priya-tik:
i updated my package.json

package-lock.json
priya-tik: i updated my package.json

gregorydlogan:
This also doesn't *look* like a multi-step wizard.  Compare creating an event to creating a user:

* It's not obvious that there are two steps right away cf the new event modal's five dots right at the top
* It's not obvious that modifying the roles is pretty much required *and* that it must happen after username selection etc
* It's possible to go straight to the roles without inputting the required parts

Arnei (changes requested):
This changes exactly what the text in #1195 requests and as such *technically* solves the issue. But as others mentioned, this is not enough to solve the *spirit* of the issue which requires the "Add User" dialog to behave and look like all the other "Add" and "Create" dialogs.

I would argue that solving the spirit of the issue is what we really want here. Just technically solving the issue does not lead to the improved user experience and usability we want, but just causes different kinds of confusion for our users. Therefore I would not merge this PR as it is right now.

If you would like to make the "Add User" dialog behave and feel like other "Add dialogs", I would suggest looking at the "Add group" dialog and going from there. Others have already given pointers to what the key differences are.

Arnei:

PR #1220, opened 2025-05-13
Add tooltip text to the close button of Modals.

### LINKED ISSUE #1218: Close button of Modals must have discernible text
Screen reader users are not able to discern the purpose of elements with role="link", role="button", or role="menuitem" that do not have an accessible name. All modals using `src/components/shared/modals/Modal.tsx` do not have such a text on the close button.

Image

luniki:
Fixes #1218.


NadiaUoM (changes requested):
Instead of hard coding the CLOSE text directly can you please read it from lang-en_US.json file. If you do so internalization can be possible if needed....

lkiesow:
> Instead of hard coding the CLOSE text directly can you please read it from lang-en_US.json file. If you do so internalization can be possible if needed....

Internationalization already works. It's not actually hard-coded. That's the translation string. So, I think we can ignore that complaint.

However, looking at the placement of the tooltip, that doesn't seem right. The placement is far too low:

Screenshot from 2025-05-13 14-40-34

PR #1234, opened 2025-05-13
No sorting for ROLES header in users page

### LINKED ISSUE #1210: Users: Sorting by roles doesn't work althought the option is available
The user table allows to sort by roles. When clicking the header cell multiple times, the data's order does not change. It doesn't matter if the sorting direction is ASC or DESC:

Image

A request is sent but the order stays the same.
Is sorting by roles even a useful or realistic use case?

ferishili:
This PR fixes #1210,

As to simple disable the sorting for the Roles column in users page.
Since it uses redux storage with might happen the cookies data need to be remove for testing!


luniki (approved):
Code is fine and it works as expected.

PR #1240, opened 2025-05-13
Sorting capture devices by last-updated-timestamp isn't possible

### LINKED ISSUE #1209: Locations: Sorting by 'Last updated' not possible although the option is available
The locations table allows to sort by 'last updated'. When clicking the header cell multiple times, the data's order does not change. It doesn't matter if the sorting direction is ASC or DESC:

Image

A request is sent but the order stays the same:

Image

wsmirnow:
Let's disable the sorting table by this column.

Fixes #1209

### DIFF
--- a/src/configs/tableConfigs/recordingsTableConfig.ts
+++ b/src/configs/tableConfigs/recordingsTableConfig.ts
@@ -29,7 +29,7 @@ export const recordingsTableConfig: TableConfig = {
 			template: "RecordingsUpdateCell",
 			name: "update",
 			label: "RECORDINGS.RECORDINGS.TABLE.UPDATED",
-			sortable: true,
+			sortable: false,
 		},
 		{
 			template: "RecordingsActionCell",
--- a/src/configs/tableConfigs/recordingsTableConfig.ts
+++ b/src/configs/tableConfigs/recordingsTableConfig.ts
@@ -27,9 +27,9 @@ export const recordingsTableConfig: TableConfig = {
 		},
 		{
 			template: "RecordingsUpdateCell",
-			name: "update",
+			name: "updated",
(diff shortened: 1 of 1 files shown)


stefanosgeo:
I don't think this is a good idea. In some cases, sorting the CAs by their 'last updated' timestamp makes more sense—because although the status might show 'online', the CA could actually be offline, and the status hasn't been updated yet.

Screenshot_20250513_161623

Arnei:
I'm inclined to agree with Stefanos. We have other date columns that are sortable, I see no reason to make this one in particular unsortable.

gregorydlogan:
@wsmirnow see my comment on #1209: rest call is sorting on `update`, needs to be `updated`.

wsmirnow:
> @wsmirnow see my comment on #1209: rest call is sorting on `update`, needs to be `updated`.

Thank you @gregorydlogan, I've updated this PR accordingly. The sorting by update**d** collumn is enabled again and fixed by using same naming as OC backend.

wsmirnow:
image

stefanosgeo:
@wsmirnow  I can review this...

stefanosgeo:
I proxied the PR with our dev cluster,  but I’m unable to sort the CAs. Did I do something wrong?

[2025-05-15 14-27 (Display, OC Studio).webm](https://github.com/user-attachments/assets/20f1a38e-6775-473d-ab10-fcab82795749)

wsmirnow:
> I proxied the PR with our dev cluster, but I’m unable to sort the CAs. Did I do something wrong?
> 
>  2025-05-15.14-27.Display.OC.Studio.webm

Please remove cookies and try again.

stefanosgeo (approved):
Sorting per update column is now working, and code change looks good

PR #1245, opened 2025-05-13
Remove console.info statements from postNewGroup and deleteGroup thunks

### LINKED ISSUE #1244: Unwanted console info in group deletion/creation
When deleting and/or creating a group in admin UI, there is an unwanted response info printed in the console which has no use and might compromise the security:

Image

ferishili:
This fixes #1244,

It helps maintain a cleaner codebase by avoiding the direct exposure of request and response details in the console, which is also beneficial from a security perspective.


ferishili:
Thanks @stefanosgeo 

[UPDATE]: I found another one related to this PR, and will update this PR as soon as I am done with an ongoing dev.

stefanosgeo (approved):

PR #1248, opened 2025-05-13
Update npm and yarn/mui/x date pickers 8.1.0

JamesUoM:
Small additional code change for dependabot PR https://github.com/opencast/opencast-admin-interface/pull/1189


Arnei (approved):
Thanks for doing the thankless tasks.

Hopefully we can get rid of this dependency alltogether soon, but until then it's good to have it up-to-date.

PR #1251, opened 2025-05-13
Display only directly assigned roles in user modal

### LINKED ISSUE #1239: In the user modal, the "Roles" and "Effective Roles" tabs display the same information.
The "Roles" tab in the user modal should only display the roles directly assigned to the user. Instead also roles inherited through other groups are shown (like in the "Effective Roles" tab):

Image

Image

snoesberger:
fix #1239

This fix ensures that only directly assigned roles are displayed in the roles tab of the user modal.

### DIFF
--- a/src/components/shared/wizard/SelectContainer.tsx
+++ b/src/components/shared/wizard/SelectContainer.tsx
@@ -252,7 +252,8 @@ const SelectContainer = ({
 							onChange={(e) => handleChangeRemove(e)}
 							value={markedForRemoval}
 						>
-{/* @ts-expect-error TS(7006): Parameter 'item' implicitly has an 'any' type. */}
+							{/* Show assigned users or directly assigned roles*/}
+							{/* @ts-expect-error TS(7006): Parameter 'item' implicitly has an 'any' type. */}
 							{selectedItems.map((item, key) => (
 								<option key={key} value={item.name}>
 									{item.name}
--- a/src/components/shared/wizard/SelectContainer.tsx
+++ b/src/components/shared/wizard/SelectContainer.tsx
@@ -252,7 +252,6 @@ const SelectContainer = ({
 							onChange={(e) => handleChangeRemove(e)}
 							value={markedForRemoval}
 						>
-							{/* Show assigned users or directly assigned roles*/}
 							{/* @ts-expect-error TS(7006): Parameter 'item' implicitly has an 'any' type. */}
 							{selectedItems.map((item, key) => (
 								<option key={key} value={item.name}>
--- a/src/components/users/partials/modal/UserDetails.tsx
+++ b/src/components/users/partials/modal/UserDetails.tsx
@@ -10,6 +10,8 @@ import { useAppDispatch, useAppSelector } from "../../../../store";
 import { UpdateUser, updateUserDetails } from "../../../../slices/userDetailsSlice";
 import WizardNavigationButtons from "../../../shared/wizard/WizardNavigationButtons";
 import { ParseKeys } from "i18next";
+import { UserRole } from "../../../../slices/userSlice";
+import { SerializedError } from "@reduxjs/toolkit";
 
 /**
  * This component manages the pages of the user details
@@ -24,7 +26,7 @@ const UserDetails: React.FC<{
 	const [page, setPage] = useState(0);
 
 	const userDetails = useAppSelector(state => getUserDetails(state));
-	const assignedRoles = userDetails.roles.filter(role => role.type === "GROUP" || role.type === "INTERNAL")
+	const assignedRoles = userDetails.roles.filter(role => role.type === "GROUP" || role.type === "INTERNAL");
 
 	const initialValues = {
 		...userDetails,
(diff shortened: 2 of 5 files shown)


Arnei (changes requested):
Works, but I would like to see the filtering happen outside the SelectContainer component. Imo it is the task of the calling components to provide the correct data. 

Did you test this with users with many roles (upwards of a thousand)? Is it performant?

snoesberger:
Thanks for the review @Arnei. I did address your suggestions and did move the filtering to the UserDetails component. Code is ready for an other review.

src/components/shared/wizard/SelectContainer.tsx
```diff
@@ -252,7 +252,8 @@ const SelectContainer = ({
 							onChange={(e) => handleChangeRemove(e)}
 							value={markedForRemoval}
 						>
-{/* @ts-expect-error TS(7006): Parameter 'item' implicitly has an 'any' type. */}
+							{/* Show assigned users or directly assigned roles*/}
```
Arnei, line 255: Leftover comment

src/components/users/partials/modal/UserDetails.tsx
```diff
@@ -24,9 +24,11 @@ const UserDetails: React.FC<{
 	const [page, setPage] = useState(0);
 
 	const userDetails = useAppSelector(state => getUserDetails(state));
+	const assignedRoles = userDetails.roles.filter(role => role.type === "GROUP" || role.type === "INTERNAL")
```
Arnei, line 27: Of type `GROUP` too? Do those not count as "effective roles" (I genuinly don't know)?

src/components/users/partials/modal/UserDetails.tsx
```diff
@@ -24,9 +24,11 @@ const UserDetails: React.FC<{
...
 
 	const initialValues = {
 		...userDetails,
+		assignedRoles,
```
Arnei, line 31: If you rename `const assignedRoles` to `const roles`, you can also write roles here. What that does is your `roles` overwrites the roles from `userDetails`, which means you don't have to change the formikField string in UserRolesTab.tsx

src/components/users/partials/modal/UserDetails.tsx
```diff
@@ -58,8 +60,16 @@ const UserDetails: React.FC<{
 		setPage(tabNr);
 	};
 
-	const handleSubmit = (values: UpdateUser) => {
-		dispatch(updateUserDetails({values: values, username: userDetails.username}));
+	const handleSubmit = (values: any) => {
```
Arnei, line 63: !`any`! Please specify a type.

Arnei (changes requested):
If I click on "Add user" and click on the "Roles" tab in the "Create new user" modal, the ui breaks (white screen).

src/components/users/partials/modal/UserDetails.tsx
snoesberger: Yes, we need also the "GROUP" type roles. They can be directly assigned to users (and in the old admin UI they were also displayed)

src/components/users/partials/modal/UserDetails.tsx
snoesberger: Unfortunately the roles attribute is also used for the "effective Roles" tab. If we overwrite them, we only see the directly assigned roles in the "effective Roles" tab.

Arnei: Ah right, we now need to pass different information to those tabs. You're absolutely right, disregard my comment.

src/components/shared/wizard/SelectContainer.tsx
snoesberger: Removed in my most recent commits

snoesberger:
> If I click on "Add user" and click on the "Roles" tab in the "Create new user" modal, the ui breaks (white screen).

Since the new user dialog also uses the SelectContainer, I had to add the assignedRoles attribute to the NewUserWizard as well. Now, creating a new user and assigning roles simultaneously is working again.

src/components/users/partials/modal/UserDetails.tsx
```diff
@@ -60,7 +62,19 @@ const UserDetails: React.FC<{
...
+		provider?: string,
+		roles?: UserRole[],
+		name?: string,
+		username: string,
+		email?: string,
+		manageable?: boolean,
+		assignedRoles?: UserRole[],
+		password?: string,
+		passwordConfirmation?: string,
+	}) => {
```
Arnei, line 77: You are specifying way more properties than `handleSubmit` is using. You only need to specify the properties that are actually used (in this case `email`, `name`, `username`, `assignedRoles`, `password`). Specifying more properties may confuse future developers that try figure out what is going on.

Same in NewUserWizard.

Arnei (changes requested):

Arnei:
Everything else is looking fine now.

src/components/users/partials/modal/UserDetails.tsx
snoesberger: I was not sure about that. I took all the attributes that are part of the initialValues. I thought it is possible to get all these attributes when submitting.
How can I find out which attributes I get from onSubmit then?

Arnei: It is not important which values `handleSubmit` *could* get from `onSubmit`, it is important which values it actually makes use of. As a general rule, a function should not define taking a value if that value is never used.

>  How can I find out which attributes I get from onSubmit then?

You cannot, not really, because Formik is not that keen on type safety. I find it to be good practice to define a complete initial state, so that you can reasonably assume to get an object with the same values in onSubmit.

snoesberger: Thanks for the explanations! In my last commit, I reduced the attributes to only the necessary ones.

Arnei (approved):

PR #1252, opened 2025-05-13
Fix the disappearing filter when the bell icon is clicked

### LINKED ISSUE #876: Error notification filter disappears
The `Actions available:Yes' filter disappears immediately after the services are loaded.

[Screencast from 08.08.2024 14:21:43.webm](https://github.com/user-attachments/assets/9e637ebc-2d48-47e6-b41e-337da394574d)

marwyg:
Fixes #876 

When clicking on the "Bell Icon" -> Backend Services, there should be a filter activated. When switching to the services page **the first time**, the filter will not be set / will disappear instantly. This PR should fix this, so the filter will be set correctly, especially when opening the page the first time.


luniki (approved):
Works as expected. Code is fine.

PR #1254, opened 2025-05-13
Changing themes of series fixed

### LINKED ISSUE #1205: Themes can't be changed for series
See https://github.com/opencast/opencast/pull/6640#pullrequestreview-2787013625: 

Setting the theme when creating a new series works, but then changing the theme for that series afterwards doesn't work - the themes are correctly listed by their name in the drop down, but once selected it only shows the id, and the changes cannot be saved. 

Looks like a key and value mix-up of the listprovider data in the admin ui that leads to an incorrect update request to the back-end that then gets rejected.

wsmirnow:
In the series dialog the themes selection is fixed.

fixes #1205

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/SeriesDetailsThemeTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/SeriesDetailsThemeTab.tsx
@@ -11,6 +11,8 @@ import { updateSeriesTheme } from "../../../../slices/seriesDetailsSlice";
 import WizardNavigationButtons from "../../../shared/wizard/WizardNavigationButtons";
 import ModalContentTable from "../../../shared/modals/ModalContentTable";
 
+type SeriesTheme = { id: string; value: string; };
+
 /**
  * This component renders the tab for editing the theme of a certain series
  */
@@ -19,14 +21,8 @@ const SeriesDetailsThemeTab = ({
 	themeNames,
 	seriesId,
 }: {
-	theme: {
-		id: string;
-		value: string;
-	} | null,
-	themeNames: {
-		id: string;
-		value: string;
-	}[],
+	theme: SeriesTheme | null,
+	themeNames: SeriesTheme[],
 	seriesId: string
 }) => {
 	const { t } = useTranslation();
@@ -34,11 +30,11 @@ const SeriesDetailsThemeTab = ({
 
 	const user = useAppSelector(state => getUserInformation(state));
 
-	const handleSubmit = (values: { theme: {id: string, value: string } | null }) => {
+	const handleSubmit = (values: { theme: SeriesTheme | null }) => {
 		dispatch(updateSeriesTheme({id: seriesId, values: values}));
 	};
 
-	const checkValidity = (formik: FormikProps<{theme: {id: string, value: string } | null }>) => {
+	const checkValidity = (formik: FormikProps<{theme: SeriesTheme | null }>) => {
 		if (formik.dirty && formik.isValid) {
 			// check if user provided values differ from initial ones
 			return !_.isEqual(formik.values, formik.initialValues);
--- a/src/components/events/partials/ModalTabsAndPages/SeriesDetailsThemeTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/SeriesDetailsThemeTab.tsx
@@ -19,7 +19,10 @@ const SeriesDetailsThemeTab = ({
 	themeNames,
 	seriesId,
 }: {
-	theme: string,
+	theme: {
+		id: string;
+		value: string;
+	} | null,
 	themeNames: {
 		id: string;
 		value: string;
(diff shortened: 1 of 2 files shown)


src/slices/seriesDetailsSlice.ts
```diff
@@ -298,17 +298,16 @@ export const updateSeriesAccess = createAppAsyncThunk('seriesDetails/updateSerie
...
+	let themeId = values.theme?.id;
 
-	let themeId = themeNames.find((theme) => theme.value === values.theme)?.id;
-
-	if (!values.theme) {
+	if (!themeId || themeId === '') {
         axios
             .delete(`/admin-ng/series/${id}/theme`)
             .then((response) => {
+				dispatch(setSeriesDetailsTheme(values.theme));
```
Arnei, line 310: Indentation

src/components/events/partials/ModalTabsAndPages/SeriesDetailsThemeTab.tsx
```diff
@@ -19,7 +19,10 @@ const SeriesDetailsThemeTab = ({
 	themeNames,
 	seriesId,
 }: {
-	theme: string,
+	theme: {
+		id: string;
+		value: string;
+	} | null,
```
Arnei, line 25: Consider turning this into a type, e.g.

```
type SeriesTheme: {
		id: string;
		value: string;
	} | null;
```

This helps avoid copy & paste mistakes, and also makes it more clear to other developers that info for the theme of a series is expected here, not some arbitrary key-value pair.

src/slices/seriesDetailsSlice.ts
```diff
@@ -165,12 +165,12 @@ export const fetchSeriesDetailsTheme = createAppAsyncThunk('seriesDetails/fetchS
 	const res = await axios.get(`/admin-ng/series/${id}/theme.json`);
 	const themeResponse = res.data;
 
-	let seriesTheme = "";
+	let seriesTheme = null;
```
Arnei, line 168: Consider typing this variable (typescript inferences `any`).

Also ` = null` is not actually necessary I think.

Arnei (changes requested):
Works and code looks generally good to me. I have some typescript considerations, tell me what you think about them.

src/slices/seriesDetailsSlice.ts
wsmirnow: I'm not sure, this variable may never be set. Isn't a better way to initialize it with null?

Arnei: That is not necessary in Javascript, as Javascript will make sure to set not-initialized variables to `undefined` .

wsmirnow: Without initialization the build process end up with an error:

```
$ npm run build

> admin-ui-picard@0.1.0 build
> tsc && eslint . --max-warnings=0 && vite build

src/slices/seriesDetailsSlice.ts:567:5 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(actionCreator: AsyncThunkFulfilledActionCreator<{ id: string; value: string; } | undefined, string, { state: { tableFilters: TableFilterState; tableFilterProfiles: TableFilterProfilesState & PersistPartial; ... 22 more ...; statistics: StatisticsState; } & PersistPartial; ... 6 more ...; rejectedMeta?: unknown; }>, reducer: CaseReducer<...>): ActionReducerMapBuilder<...>', gave the following error.
    Argument of type '(state: WritableDraft<SeriesDetailsState>, action: { payload: { id: string; value: string; } | null; type: string; }) => void' is not assignable to parameter of type 'CaseReducer<SeriesDetailsState, PayloadAction<{ id: string; value: string; } | undefined, string, { arg: string; requestId: string; requestStatus: "fulfilled"; }, never>>'.
      Types of parameters 'action' and 'action' are incompatible.
        Type 'PayloadAction<{ id: string; value: string; } | undefined, string, { arg: string; requestId: string; requestStatus: "fulfilled"; }, never>' is not assignable to type '{ payload: { id: string; value: string; } | null; type: string; }'.
          Types of property 'payload' are incompatible.
            Type '{ id: string; value: string; } | undefined' is not assignable to type '{ id: string; value: string; } | null'.
              Type 'undefined' is not assignable to type '{ id: string; value: string; } | null'.
  Overload 2 of 2, '(type: string, reducer: CaseReducer<SeriesDetailsState, { payload: { id: string; value: string; } | null; type: string; }>): ActionReducerMapBuilder<SeriesDetailsState>', gave the following error.
    Argument of type 'AsyncThunkFulfilledActionCreator<{ id: string; value: string; } | undefined, string, { state: { tableFilters: TableFilterState; tableFilterProfiles: TableFilterProfilesState & PersistPartial; ... 22 more ...; statistics: StatisticsState; } & PersistPartial; ... 6 more ...; rejectedMeta?: unknown; }>' is not assignable to parameter of type 'string'.

567    .addCase(fetchSeriesDetailsTheme.fulfilled, (state, action: PayloadAction<
        ~~~~~~~

Found 1 error in src/slices/seriesDetailsSlice.ts:567
```

I will leave it initialized

Arnei: Aah, because the reducer expects `null` instead of `undefined`. Fair enough, did not think about that.

I'd still suggest typing the variable though to avoid the `any` inference, for example:
```
let seriesTheme: SeriesDetailsState["theme"] | null = null;
```

wsmirnow: Make sens. Is done.

Arnei (approved):

Issue #1201, opened 2025-05-08
Selected roles are not displayed in group details


snoesberger:
When you open a group and go to the "Roles" tab, the selected roles are not displayed (even though roles are actually assigned to the group):

![Image](https://github.com/user-attachments/assets/46a51dfd-9210-49bc-8b61-f012845e254e)

priya-tik:
Hello all,  i resolved this bug

Arnei:
This has been fixed.

PR #1256, opened 2025-05-13
Updating groups now preserves previously selected roles

### LINKED ISSUE #1237: As soon as a user is assigned to a group, the group loses its roles.
To reproduce the problem:
1. Create a group and assign roles.
2. Open the newly created group and switch to the Users tab. Add a user.
3. Save the changes.
4. Check the group in the Admin UI or via the API to see that the roles added in step 1 are no longer assigned.

ferishili:
This PR fixes #1237,
### Description
Updating groups removes the already assigned Roles

### Problem
Because of using Formik with `const initialValues` and consuming its values later on in the SelectContainer component automatically, we should not use the `useState` to set values to roles as in `roleNames` => **asynchronously** and then directly use it in a const **synchronously**

### Solution
The changes just make sure the `initialValues` gets the mapped roles properly!

### DIFF
--- a/src/components/users/partials/modal/GroupDetails.tsx
+++ b/src/components/users/partials/modal/GroupDetails.tsx
@@ -24,26 +24,15 @@ const GroupDetails: React.FC<{
 	const groupDetails = useAppSelector(state => getGroupDetails(state));
 
 	const [page, setPage] = useState(0);
-		// transform roles for use in SelectContainer
 
-	const [roleNames, setRoleNames] = useState<{ name: string }[]>([]);
-
-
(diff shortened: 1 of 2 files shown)


ferishili:
This might also solve issue #1201

luniki:
This works now as expected.

Roles tab after adding a user:

image

Users tab:

image

Functionality reviewed

ferishili:
The removal of already selected users (right) from the users list (left) is fixed in PR #1257

luniki (approved):
Code is fine.

PR #1257, opened 2025-05-14
Make sure selected users are not listed in the users lists

### LINKED ISSUE #1243: Groups: Assigned users are listed twice in group details
I created a new group and assigned two users. After opening the corresponding group details again and switching to the users tab, the assigned users are displayed in the list on the left as well as in the _Selected users_ list:

Image

I expected all assigned users to be removed from the left list.

ferishili:
This PR fixes #1243,

which removes the selected users (right list) from the users list (left list) in SelectContainer,
also introducing array compare for the remove method instead of single comparison.


luniki:
Works as expected. Functionality reviewed. 

image

luniki (approved):
Code reviewed!

PR #1258, opened 2025-05-14
Filtering from events cells in pages other than first page now works!

### LINKED ISSUE #1229: Filtering for specific properties works on first page only
Steps to reproduce:

1. Open Events table and stay on first page.
2. Click on a presenter or a series name to filter for that property - it works.
3. Switch to the second page.
4. Again, click on a presenter or a series name - an empty result set is displayed although entries with the selected presenter or series name exist.

https://github.com/user-attachments/assets/6d6b1c16-c6ee-4ab5-829e-57534c52ee78

ferishili:
This PR fixes #1229,

Applying a debouncing mechanism when the filtering should be applied in order to fetch the events and load them into the table. Using debounce helps reduce unwanted redundent calls.
In addition, we use useState and useEffect in order to use the advantage of async and watcher properties!

The main problem also was the fact that by applying filters no matter which page the table is set to, it has to be set to first page, so here we use `goToPage(0)`

### TESTING
Please make sure to test all event cells that are filterable, by clicking on the values! don't hesitate to ask if you have any question. Thanks in advance!

### DIFF
--- a/src/components/events/partials/EventsDateCell.tsx
+++ b/src/components/events/partials/EventsDateCell.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { useTranslation } from "react-i18next";
 import { editFilterValue } from "../../../slices/tableFilterSlice";
 import { loadEventsIntoTable } from "../../../thunks/tableThunks";
@@ -8,6 +8,9 @@ import { fetchEvents } from "../../../slices/eventSlice";
 import { renderValidDate } from "../../../utils/dateUtils";
 import { Event } from "../../../slices/eventSlice";
 import { IconButton } from "../../shared/IconButton";
+import {
+	goToPage,
+} from "../../../thunks/tableThunks";
 
 /**
  * This component renders the start date cells of events in the table view
@@ -17,6 +20,8 @@ const EventsDateCell = ({
 }: {
 	row: Event
 }) => {
+	// Using itemValue with useState in order to use as flag in useEffect as watcher!
+	const [itemValue, setItemValue] = useState('');
 	const { t } = useTranslation();
 	const dispatch = useAppDispatch();
 
@@ -24,6 +29,7 @@ const EventsDateCell = ({
 
 	// Filter with value of current cell
 	const addFilter = async (date: string) => {
+		let mustApplyChanges = false;
 		let filter = filterMap.find(({ name }) => name === "startDate");
 		if (!!filter) {
 			let startDate = new Date(date);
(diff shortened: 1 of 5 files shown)


wsmirnow (approved):
Works as expected. My knowledge to the code is limited. Maybe someone else can look into the code, too. For me it's fine.

Good job, thank you @ferishili

Arnei (commented):
Could you explain in more detail what the debounce is supposed to achieve? What requests are you debouncing? My understanding of the code is that all the debounce does is delaying code execution by 500 ms for no particular reason.

Arnei:
As an aside, this PR also nicely illustrates that we have code duplication in our table cells which we should get rid of at some point :D. Not necessarily in this PR though, no worries.

ferishili:
You are 100% right about the code duplication! And I can bring this to the /Util and reuse them!
About debouncing, the first advantage is to make sure an unwanted (unintentional) double click does not clutter the processing! And secondly, it would give a nice waiting touch in order for React to do its thing and make the changes more feelable to the users (IMHO)!

ferishili:
If you are convinced with having the debounce and my proposed way! I can make it reusable

Arnei:
> the first advantage is to make sure an unwanted (unintentional) double click does not clutter the processing!

I did not think about that users double- or even multi-clicking would trigger multiple backend requests. I agree that ideally we want to prevent this from happening. I feel like preventing double clicks should be easier to solve though. I worry that have this much code just for debouncing makes the component as a whole unnecessarily harder to understand. I don't know of a better solution off the top my head though.

> waiting touch in order for React to do its thing

I don't think this really achieves that. If understand the code correctly, we first wait 500ms and only then we fetch events and load them into the table. So React is only doing its thing after waiting for 500ms for no reason. Am I getting this wrong?

> make the changes more feelable to the users (IMHO)

Yeah, that is very much a matter of design opinions :D. Personally, I don't like introducing artificial lag when avoidable, due to it feeling more responsive. But I'm likely also way more sensitive about lag then the average user ^^.

Suggestion: All this talk about debouncing is only tangentially related to the issue this PR is supposed to solve. I think it would be better to factor debouncing out of this PR and continue the discussion about debouncing in another PR or issue.

ferishili:
In this case, and for the favor of keeping the code base simple, we should close this and  go for the PR: #1266 from Carolina, which is a simpler version of this and resolves the main problem by adding `setPage` to 0 when cell filtering!

marwyg:
As mentioned above, I will close this PR and reopen Caros PR.

PR #1274, opened 2025-05-14
Broadcasting logout event in order to redirect other tabs to login page

### LINKED ISSUE #1194: Go to login form if user is logged out
- Login to develop.opencast.org
- Stay in the admin interface
- Open develop.opencast.org in a second tab
- In the second tab, log out
- Go back to the first tab

It seems like you are still logged in. You are just missing some data and the admin interface shows some errors:
Image

The interface should detect that the user is no longer logged in and should redirect the user to the login form.

ferishili:
This PR fixes #1194,

It introduces a broadcasting mechanism (to contain further implementation for other places as well), focusing on logout event.

On the App level we first subscribe to the broadcasting event, and on the Header => MenuUser on logout we broadcast the actual logout!

The actual tab is going to logout which redirects to `j_spring_security_logout`, but the other opened tabs will be redirected to login page with a 750 ms wait!


rlucke (approved):
looks fine to me and works as well

PR #1204, opened 2025-05-08
Fix group cancel button

priya-tik:
This PR adds missing cancel functionality to the Group Details wizard modal. Previously, the cancel button was non-functional due to the absence of a callback handler. This update ensures a smooth user experience by allowing users to exit the modal properly.

PR #1277, opened 2025-05-14
make cancel buttons work for users/groups/access

### LINKED ISSUE #1182: Cancel button not working while editing User/Groups/Access policies tab
When trying to edit user, groups or access policies details. The cancel button on the wizard in not working. 

Refer below screenshot:

Image

ppettit:
Fixes #1182 

Just makes the cancel button close the modal which is the behaviour of the old admin UI

### DIFF
--- a/src/components/users/partials/modal/GroupDetails.tsx
+++ b/src/components/users/partials/modal/GroupDetails.tsx
@@ -98,6 +98,7 @@ const GroupDetails: React.FC<{
 						{/* Navigation buttons and validation */}
 						<WizardNavigationButtons
 							formik={formik}
+							previousPage={() => close()}
 							createTranslationString="SUBMIT"
 							cancelTranslationString="CANCEL"
 							isLast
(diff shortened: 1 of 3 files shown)


Arnei:
Supercedes #1204 I suppose?

src/components/users/partials/modal/GroupDetails.tsx
```diff
@@ -98,6 +98,7 @@ const GroupDetails: React.FC<{
 						{/* Navigation buttons and validation */}
 						<WizardNavigationButtons
 							formik={formik}
+							previousPage={() => close()}
```
Arnei, line 101: Can be simplified to (in all three instances):

```suggestion
							previousPage={close}
```

Arnei (changes requested):
Works and makes sense to me. Small suggestion below

src/components/users/partials/modal/GroupDetails.tsx
ppettit: done, i had copied and pasted from elsewhere in the code :wink:

ppettit:
> Supercedes #1204 I suppose?

yes, sorry that as not linked to the issue so i didn't find it. seems like a lot of unrelated changes in that PR too

Arnei (approved):

PR #1278, opened 2025-05-14
Override events ACLs by separate button in Series Details Update

### LINKED ISSUE #1263: Changing series ACL may corrupt data
The old admin interface allowed users to set a new series ACL and then push the ACL to events of the series which would overwrite existing event ACLs. This made it very clear what happens:

Image

The new interface lacks the button and seems to automatically overwrite events, unknown to users. This is not only confusing to users, but may actually cause data corruption, since existing ACLs may be unknowingly be overwritten.

Long story short:
- Bring back the button
- Stop just overwriting data

ferishili:
This PR fixes #1263,
this adds override button functionality to access policy components for series,
The problem was, the override parameter was always "true".
It is now possible to make sure that the overriding of acls onto the events happens intentionally!

In additon, the `SaveEditFooter` component is now tuned to have additional button on demand!

### DIFF
--- a/src/slices/eventDetailsSlice.ts
+++ b/src/slices/eventDetailsSlice.ts
@@ -1708,7 +1708,7 @@ export const saveAccessPolicies = createAppAsyncThunk('eventDetails/saveAccessPo
 	params: {
 		id: Event["id"],
 		policies: { acl: { ace: Ace[] }},
-		override: boolean, // Although there is no use of parameter here, but in order to keep it sync with series and make the mechanism reuseable, we should keep it!
+		override?: boolean,
 	}, { dispatch }) => {
 	const { id, policies } = params;
 	const headers = getHttpHeaders();
--- a/src/slices/eventDetailsSlice.ts
+++ b/src/slices/eventDetailsSlice.ts
@@ -1704,16 +1704,17 @@ export const updateAssets = createAppAsyncThunk('eventDetails/updateAssets', asy
 		});
 });
 
-export const saveAccessPolicies = createAppAsyncThunk('eventDetails/saveAccessPolicies', async (params: {
-	id: Event["id"],
-	policies: { acl: { ace: Ace[] } }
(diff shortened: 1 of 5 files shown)


src/slices/eventDetailsSlice.ts
```diff
@@ -1704,16 +1704,17 @@ export const updateAssets = createAppAsyncThunk('eventDetails/updateAssets', asy
...
-export const saveAccessPolicies = createAppAsyncThunk('eventDetails/saveAccessPolicies', async (params: {
-	id: Event["id"],
-	policies: { acl: { ace: Ace[] } }
-}, { dispatch }) => {
+export const saveAccessPolicies = createAppAsyncThunk('eventDetails/saveAccessPolicies', async (
+	params: {
+		id: Event["id"],
+		policies: { acl: { ace: Ace[] }},
+		override: boolean, // Although there is no use of parameter here, but in order to keep it sync with series and make the mechanism reuseable, we should keep it!
+	}, { dispatch }) => {
```
Arnei, line 1712: I get what you're saying, but I'm really not a fan of forcing functions to accept parameters they have no use for. I don't think this scales well. I would prefer if you could find another solution. Maybe making `override` optional would be a valid approach?

Arnei (changes requested):
Works as advertised and code make sense to me. Really only have one complaint, see below.

src/slices/eventDetailsSlice.ts
Arnei: Now the argument is still there but optional. The reason i suggested to make `override` optional in `ResourceDetailsAccessPolicyTab` was so that it could be removed here entirely :D.

ferishili: you are right, coming asap

snoesberger:
I opened https://github.com/opencast/opencast-admin-interface/issues/1290 because some of the behavior differs from the old admin UI and the documentation.

Issue #1179, opened 2025-04-15
Minor issue: Notification label when uploading multiple events


marcusm7:
When two or more events are uploaded at once, the notification area does not show multiple vertically stacked notifications but instead this happens:
![Image](https://github.com/user-attachments/assets/e5eb4288-f2cd-4fd9-b59a-30bd0e1d1fc3)  
(please allow animated gifs in your browser)
  
Maybe add the event name/ UID and stack the notifications vertically?  
  
Opencast 17.2

PR #1281, opened 2025-05-14
Fix multiple upload notifications overlapping when uploading multiple videos

jduehring:
This PR should fix https://github.com/opencast/opencast-admin-interface/issues/1179 . 

Instead of hardcoding and using the same notification id for every upload, we now generate a new one for each seperate upload so that only the correct notification gets overwritten. 

Additionally, I’ve added the event title to the upload notification, allowing users to more easily associate the upload notifications with their respective events.


michiHB:
For me it worked fine and solved the problem documented in the issue.

Arnei (approved):
Works and code makes sense to me given the circumstances.

There's a small issue with multiple notifications blocking the ui, made worse by the fact that the notifications can't be dismissed (they dissappear but pop right back up). Not sure if we want to consider this a blocker though.

Bildschirmfoto vom 2025-05-14 15-15-26

Oh, and if we have the name of the event in the upload notification now, it would be nice to also have it in the upload-done notification. But that can very well be left for a different PR.

PR #1286, opened 2025-05-14
Deselect table rows when sorting

### LINKED ISSUE #1250: Selection should be dropped if you change the table filter
- Go to events
- Select all/some events
- Change the sort option (sort by name or date; make sure the table changes)
- Your selection is essentially broken

If you change the sorting of events, we should just drop the selection.

ferishili:
This PR fixes #1250,

It applies to every table list that has selectable feature and sorting together!


NadiaUoM:
If you select all items on a page like the event details and you navigate to next ,the “select all” checkbox still checked. See below:

image

Arnei (approved):
Does what it says and makes sense to me.

PR #1295, opened 2025-05-19
Series and events permissions override enhancement

### LINKED ISSUE #1290: Button text for series ACL changes is unclear and 2nd save button can't be deactivated
The text on the save buttons in the ACL changes series modal is unclear to me.

Image

After looking at https://github.com/opencast/opencast-admin-interface/pull/1278, which introduced the second button, I realized that one of the two buttons will override the ACL of all events in that series. The other button will only modify the series ACL.

The end user won't know the difference between the two buttons, they need a better explanation. e.g. "Save series permission" and "Save series and override Event permissions".

The second problem is that the behavior is not as described in the documentation (https://docs.opencast.org/r/17.x/admin/#configuration/acl/#updating-series-permissions) and is different from the old admin UI.
Depending on the configured merge mode, overriding the event ACL may not make sense (e.g., with the "actions" merge mode). At our institution we have such a setting and would expect to still be able to use the parameter `prop.admin.series.acl.event.update.mode=never`, as described in the documentation, to deactivate the button that overrides all event ACLs.

ferishili:
This PR fixes #1290,

It introduces the optional offering of the button for the "overwrite of the events" when saving the series acl based on the config: `admin.series.acl.event.update.mode` as to being `optional` or empty to offer both buttons!
The button label and hint also got slightly more meaningful texts.

**NOTE:** At the time of creating this PR there is a commit/PR missing from `develop` (default) which makes the user and system perm separation, hence the permission dialogs are broken! So, please use `stable.opencast.org` in order to test!


Arnei (changes requested):
The configuration key `admin.series.acl.event.update.mode` seems to be missing in the backend. At least I could not find it. If it should exist, could you add it?

src/components/events/partials/ModalTabsAndPages/SeriesDetailsAccessTab.tsx
```diff
@@ -28,6 +31,10 @@ const SeriesDetailsAccessTab = ({
 	const policies = useAppSelector(state => getSeriesDetailsAcl(state));
 	const policyTemplateId = useAppSelector(state => getPolicyTemplateId(state));
 
+	const orgProperties = useAppSelector(state => getOrgProperties(state));
+
+	const overrideEnabled = (orgProperties['admin.series.acl.event.update.mode'] || 'optional').toLowerCase() === 'optional';
```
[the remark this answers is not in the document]
ferishili: You are right!
in this place to check and offer the button if it is optional is correct.
what is missing is to make sure that never & always are also in the play!

Arnei:
> > The configuration key `admin.series.acl.event.update.mode` seems to be missing in the backend. At least I could not find it. If it should exist, could you add it?
> 
> It is under `etc/org.opencastproject.organization-mh_default_org.cfg` but by default commented out and set to optional!

Sorry for troubling you, but could you point out the line in the file? I'm really not seeing it in https://github.com/opencast/opencast/blob/r/17.x/etc/org.opencastproject.organization-mh_default_org.cfg

Arnei:
> > Sorry for troubling you, but could you point out the line in the file? I'm really not seeing it in https://github.com/opencast/opencast/blob/r/17.x/etc/org.opencastproject.organization-mh_default_org.cfg
> 
> No worries, it seems that it is missing in 17.x but it is there in 16.x!
> 
> https://github.com/opencast/opencast/blob/0fea543a1a8ada9e9017758e66bcbdf2513363c4/etc/org.opencastproject.organization-mh_default_org.cfg#L341C1-L342C1
> 
> However, the documentation for 17.x says it is still effectively there: https://docs.opencast.org/r/17.x/admin/#configuration/acl/#updating-series-permissions
> 
> [UPDATE]: @Arnei, I found [the patch](https://github.com/opencast/opencast/commit/57cac158a36ac0efe8c9b7380498dcc9effffed6#diff-854297734cef7879c036035654a6388a42556536ef0eb1e92104c89d7381c9f7) that removed the settings, it appears to me that this setting is removed considering it is related to old admin-ui which shouldn't! PS: fun fact! Both removing commit and requesting to add it back came from Lars 😆

:D Then would you do Lars the favor of adding it back to 17.x, just like it was in 16.x?

snoesberger:
Thanks for this PR @ferishili.

I did test this PR with our test environment where we have merge mode "actions" configured.

Setting `prop.admin.series.acl.event.update.mode=never` now hides the "Save series and overwrite event permissions" button and changing series permissions doesn't override the event ACLs.

If I configure `prop.admin.series.acl.event.update.mode=optional` and use the "Save series and overwrite event permissions" button, the existing event ACL is removed and replaced by the new series ACL. The "Save series permission" button only changes the series ACL. This works as expected

However, the behaviour of `prop.admin.series.acl.event.update.mode=always` does not appear to be as described in the documentation.
> - always:   When modifying series permissions, automatically remove all permission rules specific to single episodes
>            belonging to the series. This enforces every episode has the rules of the series in effect as soon as they
>            are changed.

In our test environment, "always" and "never" behave exactly the same. As I understand it, when changing the series ACLs, the event ACLs should be removed and replaced with the new series ACLs. This is equivalent to the "Save series and overwrite event permissions" button with the `prop.admin.series.acl.event.update.mode=optional` configuration.

Arnei:
Rebased this onto r/17.x, as this is "fixing" a feature introduced with r/17.x (which was main at the time).

gregorydlogan:
I'm seeing wonky behaviour locally which I can't reproduce with legacy.opencast.org:


https://github.com/user-attachments/assets/fd7224b4-09d5-4db6-8cca-0a3602e5d4b0

This happens every time you modify the permissions for an *unsaved* role.  So at the end of the video, if I went back and modified ROLE_ADMIN's permissions everything would work fine.  If, however, I added a different role then it would reset the new role's dropdown to "Select or create a role" every time I change the permission - either adding, or removing it.

gregorydlogan:
Given that 17.x is pretty much EOL at this point, should this get either punted forward to 18, or refiled when it's ready?

src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1225,12 +1225,14 @@
 						"ADDITIONAL_ACTIONS": "Additional Actions",
 						"ACTION": "Actions",
 						"NEW": "New policy",
+						"DETAILS": "Details",
+						"REPLACE_EVENT_ACLS": "Save series and overwrite event permissions",
+						"REPLACE_EVENT_ACLS_HINT": "Save the series permissions and overwrite the permissions for all events in this series.",
+						"SAVE_SERIES_ACL_ONLY": "Save series permission",
```
mtneug, line 1231: Is adding new keys allowed in release branches (docs say ["modify existing translation keys"](https://docs.opencast.org/develop/developer/#participate/development-process/#acceptance-criteria-for-patches-in-different-versions)).

ferishili: This is not yet final!
I only rebased to 18.x and fixed the conflicts!

ferishili: Regarding this, I would guess this would be the answer: https://github.com/opencast/opencast-admin-interface/pull/1295#issuecomment-3112553370

However, I would ask your feedback on this, @Arnei

ferishili:
> I'm seeing wonky behaviour locally which I can't reproduce with legacy.opencast.org

I went deep and tried to fix this in this PR, even though it's not directly related to the original issue. It should be handled in a separate issue, since it will likely take more time than I initially expected.

The root cause is the use of filtered policies combined with `.map`, which breaks the reactivity of the Formik values. As a result, the `policies` value for each element is no longer reactive and ends up applying its own isolated state to the object.

Arnei (commented):
This seems to work fine and code looks fine too.

I agree with Farbod that the issue Greg found is unrelated to this pull request and as such should be solved separately.

Now all we need to decide on is which branch this should go into.

KatrinIhler:
Discussed this in the Dev Meeting, we would be okay with making an exception this time, but final decision is up to @jduehring and @dennis531

PR #1297, opened 2025-05-19
Make textFilter persist across tabs

Arnei:
Same as #980, but for the textfilter, aka the search field.

When typing something into the search bar, then switching to a different tab, then switching back to the original tab, the search bar is empty. With this patch, the search bar will instead remember what was originally written in it.

This changes the data structure of data persisted in local storage. As such, the app will crash for users with now outdated local storage.

### How to test this

Clear your browser cache. Write something in the search field. Switch to another tab (e.g. users). Switch back to original tab.


snoesberger:
This PR doesn't work for me, it breaks the entire admin UI. The following error message appears in the browser console:
```
tableFilterSelectors.ts:16 Uncaught TypeError: state.tableFilters.textFilter.find is not a function at getTextFilter (tableFilterSelectors.ts:16:51) at TableFilters.tsx:53:45
getTextFilter @ tableFilterSelectors.ts:16
(anonymous) @ TableFilters.tsx:53
TableFilters @ TableFilters.tsx:53
```

Arnei:
Try clearing your browser cache, it should run fine then.

snoesberger:
> Try clearing your browser cache, it should run fine then.

Sorry, you are right... 🙈 After clearing the browser cache it also works for me.

PR #1298, opened 2025-05-20
Add "Metadata saved" notification

Arnei:
Adds a short notification whenever the user clicks on "Save" in the metadata tab.

Bildschirmfoto vom 2025-05-20 11-52-55

### How to test this

Can be tested as usual. Try saving some metadata!


NadiaUoM (approved):

src/components/events/partials/ModalTabsAndPages/DetailsMetadataTab.tsx
```diff
@@ -52,6 +54,13 @@ const DetailsMetadataTab = ({
 
 	const handleSubmit = (values: { [key: string]: any }, catalog: MetadataCatalog) => {
 		dispatch(updateResource({id: resourceId, values, catalog}));
+		dispatch(addNotification({
+			type: "info",
+			key: "METADATA_SAVED",
+			duration: 3,
+			parameter: undefined,
+			context: NOTIFICATION_CONTEXT,
+		}));
```
JulianKniephoff: Should this not wait for the above action to complete? Also what if it fails? 🤔

JulianKniephoff (changes requested):

src/components/events/partials/ModalTabsAndPages/DetailsMetadataTab.tsx
Arnei: How dare you not letting me get away with my half-assed implementation >.> Should properly respect action states now.

src/components/events/partials/ModalTabsAndPages/DetailsMetadataTab.tsx
```diff
@@ -52,6 +54,13 @@ const DetailsMetadataTab = ({
...
 	};
```
JulianKniephoff: Wait, shouldn't something like this suffice? 👀

```suggestion
	const handleSubmit = async (values: { [key: string]: any }, catalog: MetadataCatalog) => {
		try {
			await dispatch(updateResource({id: resourceId, values, catalog}));
			dispatch(addNotification({
				type: "info",
				key: "METADATA_SAVED",
				duration: 3,
				parameter: undefined,
				context: NOTIFICATION_CONTEXT,
			}));
		} catch (_) {
			dispatch(addNotification({
				type: "warning",
				key: "METADATA_NOT_SAVED",
				duration: 3,
				parameter: undefined,
				context: NOTIFICATION_CONTEXT,
			}));
		}
	};
```

Arnei: ...fair enough. My intuition told me to not do this, but now that I'm thinking about it more I'm struggling to come up with any good reasons on why this would be a bad approach.

src/components/events/partials/ModalTabsAndPages/DetailsMetadataTab.tsx
```diff
@@ -50,8 +52,27 @@ const DetailsMetadataTab = ({
...
+			})
+			.catch(() => {
+				dispatch(addNotification({
+					type: "warning",
+					key: "METADATA_NOT_SAVED",
+					duration: 3,
+					parameter: undefined,
+					context: NOTIFICATION_CONTEXT,
+				}));
+			})
```
JulianKniephoff: Interesting! I would prefer

```suggestion
	const handleSubmit = async (values: { [key: string]: any }, catalog: MetadataCatalog) => {
		try {
			await dispatch(updateResource({ id: resourceId, values, catalog })).unwrap();
			
			dispatch(addNotification({
				type: "info",
				key: "METADATA_SAVED",
				duration: 3,
				parameter: undefined,
				context: NOTIFICATION_CONTEXT,
			}));
		} catch {
			dispatch(addNotification({
				type: "warning",
				key: "METADATA_NOT_SAVED",
				duration: 3,
				parameter: undefined,
				context: NOTIFICATION_CONTEXT,
			}));
		}
	};
```

But I'd we willing to let that slide as a matter of taste for teaching me about the necessity of `unwrap` here. :D

**However**, if you don't `await`, you don't need `async`, either.

src/components/events/partials/ModalTabsAndPages/DetailsMetadataTab.tsx
```diff
@@ -50,8 +52,27 @@ const DetailsMetadataTab = ({
...
-		dispatch(updateResource({id: resourceId, values, catalog}));
+	const handleSubmit = async (values: { [key: string]: any }, catalog: MetadataCatalog) => {
+		dispatch(updateResource({id: resourceId, values, catalog}))
+			.unwrap()
+			.then(() => {
+				dispatch(addNotification({
+					type: "info",
+					key: "METADATA_SAVED",
+					duration: 3,
+					parameter: undefined,
```
JulianKniephoff: While we're at it: `key: undefined` is always* redundant.

src/components/events/partials/ModalTabsAndPages/DetailsMetadataTab.tsx
```diff
@@ -50,8 +52,27 @@ const DetailsMetadataTab = ({
...
+					context: NOTIFICATION_CONTEXT,
+				}));
+			})
+			.catch(() => {
+				dispatch(addNotification({
+					type: "warning",
+					key: "METADATA_NOT_SAVED",
+					duration: 3,
+					parameter: undefined,
```
JulianKniephoff: See above.

JulianKniephoff (changes requested):

JulianKniephoff (approved):
LGTM now 😊

PR #1299, opened 2025-05-20
Persist table sort

Arnei:
Tables can be sorted by columns. This is already preserved across tabs (i.e. when switching between "events" and "series" tables), but not page reloads. This adds table sorting to local storage so that sorting is preserved across page reloads.

### How to test this

Can be tested as usual. You may need to clear your browser cache.


NadiaUoM:
I have a question. For testing purposes, I deleted my browser history and performed a certain column sorting, which worked fine. However, if I delete my browser history again and refresh the admin UI, what is  expected behavior? Based on my understanding, the column sorting should revert to the original order, rather than the sorted arrangement I set earlier.

Arnei:
If by history you mean cache (website and cookie data) then yes, that is correct. Sorting is saved in local storage, so if the local storage is deleted, the app should revert to default behaviour.

NadiaUoM:
Thanks Arnei for the clarification....

NadiaUoM (approved):

PR #1310, opened 2025-05-26
Event details comments dropdown fix

priya-tik:
The dropdown in the Event Details Comments section was opening downward by default.
image


gregorydlogan (changes requested):
Hey @priya-tik, can you remove the `.idea` directory?  We don't include environment files like that

priya-tik:
> Hey @priya-tik, can you remove the `.idea` directory? We don't include environment files like that

Thanks for pointing that out — I’ve removed the .idea directory now. I didn’t notice it had been committed earlier. Should be all clean now.

src/components/shared/DropDown.tsx
```diff
@@ -51,6 +51,7 @@ const DropDown = <T, >({
 	disabled?: boolean,
 	menuIsOpen?: boolean,
 	handleMenuIsOpen?: (open: boolean) => void,
+	menuPlacement?: 'auto' | 'top' | 'bottom',
```
Arnei, line 55: Why add this as a prop when we're hardcoding the value later anyway?

src/components/shared/DropDown.tsx
```diff
@@ -121,6 +122,7 @@ const DropDown = <T, >({
 
 
   let commonProps: Props = {
+	  	menuPlacement: 'auto',
```
Arnei: Looks like there are some accidental whitespaces here.

Arnei (changes requested):

src/components/shared/DropDown.tsx
priya-tik: I’ve updated the component to accept menuPlacement as a prop and use it dynamically via destructuring. It now defaults to 'auto' if no value is passed, so it's no longer hardcoded.

src/components/shared/DropDown.tsx
Arnei: The whitespaces are still there.

PR #1312, opened 2025-05-26
Enable lints

JulianKniephoff:
We currently disable some of the lints that our chosen ESLint config has in it because, well, the code currently just doesn't follow that style. It should, though, and this is my first push towards making it so.

Starting with the simple rules, this PR has two commits:

- The first just enables a few more rules
- The second is literally what you get when you run the command in that commit's message, i.e.
  ```
  npx eslint . --fix
  ```

Please don't be intimidated by the size of this diff. You literally only need to check `eslint.config.js`. If you want to verify the rest, just run the above command on the first commit. I promise I didn't sneak in a backdoor. 😉

Also this will probably cause a lot of conflicts with other PRs. I'm prepared to keep this updated if people want to wait for something major, given that it is trivial to update, but that also means that we shouldn't wait an eternity to merge it since other people can update their PRs equally trivially.


Arnei:
@gregorydlogan You said something about switching tabs for spaces when rewriting history. Would this also qualify, because this is also automatically generated style changes?

gregorydlogan:
I probably can, but I'm not sure I should. While this is 'just' style changes, it changes what the functional content of the commit is. 

For example, a commit authored with a given style 2 years ago would now get *current* style applied to it. The only reason I'm planning on applying the whitespace changes are that they *do not* affect a given commit's buildability.

Arnei:
I see, thanks for the explanation.

Arnei (approved):

PR #1313, opened 2025-05-27
Add unsaved changes warning in series details

Arnei:
Adds a confirmation modal that warns about unsaved changes to the metadata and extended metadata tabs of the series details.

Based off of commit cd6b4e36e221751116b5ec764643d9c89ce14984.

### How to test this

Open the series details for an existing series.
Change some metadata. Do not save.
Try to close the series details.


snoesberger:
I did test this PR and the pop up with the warning about unsaved changes is displayed for all edit series tabs active in our test environment (Metadata, Access Policy, Theme, Tobira).

I just have one minor suggestion for which I did open an additional issue:
https://github.com/opencast/opencast-admin-interface/issues/1342

gregorydlogan (commented):
Should this warn you if you're swtiching tabs as well?  If I add an organizer, but switch to the ACL tab prior to actually saving I see a very misleading green checkmark, but if I switch back to the metadata tab my changes are gone.

Edit: @snoesberger's #1342 is exactly what I'm complaining about :)

Arnei:
This is intended to only add the warning when closing the modal, like it works for events already. Though I agree with both of you that having it on tab switch would be nice too. I would leave that to a different PR.

PR #1280, opened 2025-05-14
Add/Edit user dialog: aligned roles

Janni2006:
This PR fixes #1264 

The roles view is now alligned as descriped in the issue inspired by the possible solution. 
image

PR #1315, opened 2025-05-28
Fix the Roles view unaligned User Dialog

TIKaushik:


Arnei:
Duplicate of #1280, probably? Please provide a proper description so we can easily know what this is about.

If it really is a duplicate of #1280, please coordinate with the creator of that PR on how to move forward.

Arnei (approved):
Simple and entirely reasonable fix (given our messy CSS). Will merge this in favor of #1280, as it is a more straightforward fix.

Issue #1316, opened 2025-06-02
Filters Persist When Switching Tabs


priya-tik:
when a user applies filters in one tab (e.g., Events) and navigates to another tab (e.g., Jobs), then returns to the original tab, the filters are still applied.

Go to Events tab.
Apply a filter (e.g., status = finished).
Switch to Jobs tab.
Switch back to Events.
Filter is still active.

Arnei:
Closing as this as it constitutes intended behaviour. If you think it should not work like this, please reopen this with your argument.

PR #1317, opened 2025-06-02
Reset filters when series changed

priya-tik:
This PR fixes the issue where filters were persisting after switching tabs. Now, filters are reset when the user navigates between tabs.


Arnei:
Closing this for the same reason as #1316, see https://github.com/opencast/opencast-admin-interface/issues/1316#issuecomment-2929793544

PR #1322, opened 2025-06-04
Direct link to the Tobira video/series

wsmirnow:
Tobira uses its own IDs (a hash-like format) for events/series. We should use them in the Tobira tab in event/series details dialog. The URLs to the Tobira realms should also point to the resource.

This PR needs backend changes from https://github.com/opencast/opencast/pull/6789

Event Opencast ID: 18e4c988-e944-4aed-8376-3b25db6a6eb8
Event URL without this patch: https://tobira.opencast.org/!v/:18e4c988-e944-4aed-8376-3b25db6a6eb8

Event Tobira ID: evL52r4riq-hr (the ev prefix used for events in general)
Event URL after: https://tobira.opencast.org/!v/L52r4riq-hr


owi92 (approved):
Did some testing, everything works. Code looks reasonable.

PR #1324, opened 2025-06-10
Slightly improve indentation in `DetailsMetadataTab`

JulianKniephoff:
Most of this is just indentation, but see commit messages for some more details. (Not that many, though.)

If people are not happy with that change, we should at least keep the first commit, though, since that fixes indentation that is flat out wrong.


Arnei (changes requested):
Looks reasonable to me. Whatever makes this file more readable is a win in my book.

I think I would like to keep the brackets for `return`, if only because we use brackets everywhere else we return JSX. I don't care too much about that though.

JulianKniephoff:
> I think I would like to keep the brackets for `return`, if only because we use brackets everywhere else we return JSX. I don't care too much about that though.

You mean like that?

Arnei (approved):

PR #1325, opened 2025-06-10
Use proper keys in `DetailsMetadataTab`

JulianKniephoff:
Using loop indices as keys is often not a good idea, since it relies on the underlying data not chaning its order. Using some kind of identifier is often better. Plus, keys are only needed for lists.


Arnei (approved):
The choices for keys look reasonable to me.

PR #1334, opened 2025-06-13
Accessibility concerns in events stats

### LINKED ISSUE #1222: main-stats accessibility issues
An h1 is used in each of the buttons. This is difficult from an accessibility point of view, as there should generally only be one h1 tag per page. You could argue about this within articles, but it is not a good idea here.
Furthermore, it is not possible to recognise the function of the button using a screen reader, as only the number and the label are read out. It is not possible to recognise that it is a filter.

Arnei:
Fixes #1222.

- Replace `<h1>` with a `<div>` to avoid wrong semantic implications  
- Add tooltip and screen reader text

### DIFF
--- a/src/components/shared/Stats.tsx
+++ b/src/components/shared/Stats.tsx
@@ -13,6 +13,7 @@ import { loadEventsIntoTable } from "../../thunks/tableThunks";
 import { useAppDispatch, useAppSelector } from "../../store";
 import { fetchEvents } from "../../slices/eventSlice";
 import { ParseKeys } from "i18next";
+import { Tooltip } from "./Tooltip";
 
 /**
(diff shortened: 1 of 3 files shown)


NadiaUoM (changes requested):
Please don't mind can we have the tool tip text for example "Apply filter Today to the table" instead of "Apply filter Today to table".

NadiaUoM (commented):
IMG_7232

Translation is not working...

Arnei:
> Please don't mind can we have the tool tip text for example "Apply filter Today to the table" instead of "Apply filter Today to table".

Sure, changed it.

NadiaUoM:
Thank you so much for the needful.

NadiaUoM (approved):

PR #1335, opened 2025-06-13
Add aria-disabled to relevant buttons

### LINKED ISSUE #1217: Saved filter sets have issues with screenreader
- The close button is only read out as a button, but not its function

- The Save button is disabled, but this is not read out

Arnei:
Fixes #1217.

Make sure screen readers are properly informing their users that disabled buttons are disabled.

Note: The linked issue also notes that the function of the cancel button is not properly read. However, my screen reader reads "Cancel button", so that has probably been
fixed at some point.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/EventDetailsCommentsTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/EventDetailsCommentsTab.tsx
@@ -272,6 +272,15 @@ const EventDetailsCommentsTab = ({
 											isSavingComment
 										)
 									}
+									aria-disabled={
+										!!(
+											!newCommentText.length ||
+											newCommentText.length <= 0 ||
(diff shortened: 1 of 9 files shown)


NadiaUoM:
@Arnei : I can see that you have added aria-disabled to all the relevant buttons except the buttons i.e. Next, Submit present on the modal wizard... Is there any specific reason for it?

Arnei:
No specific reason, I simply missed those. Thanks for catching that!

NadiaUoM:
Apologies for bothering you again, but I found another instance where aria-disabled has not been added. Please refer to the screenshot below "Event Details" -> "Comments" -> "Submit".

image

Arnei:
Thanks for catching that as well, should be fixed now.

We really should add a base button component at some point, so we only have to make changes like this in one place.

PR #1338, opened 2025-06-16
Dock saved filter sets modal to the right

### LINKED ISSUE #1238: Saved filter sets: Window sometimes changes its position / anchor
This is a little nit-picking. However, here are the steps to reproduce:

1. Open Events table.
2. Click on the gear icon on the top right (tooltip: 'Saved filter sets') to open the _Saved filter sets_ window. 
3. Click on the filter icon (tooltip: 'Add filters') and add a filter with a name as long as possible (or two filters). The small _Saved filter sets_ window has changed its position (or anchor?):

Image

Initially, it had been on the far right:

Image

Oddly, if you add a filter with a short name, the window position will stay the same:

Image

I suggest that the _Saved filter sets_ window should always stay on the right.

Arnei:
Fixes #1238 .

Previously it was docked to the left, which would cause it to move in case the filter field was streching very long.

Bildschirmfoto vom 2025-06-16 16-37-51

### DIFF
--- a/src/styles/components/data-filter/_filter-profiles.scss
+++ b/src/styles/components/data-filter/_filter-profiles.scss
@@ -34,7 +34,7 @@
 @mixin filter-dd-content-mixin ($width) {
     // Base styles
     position: absolute;
-    top: 52px; left: 0;
+    top: 52px; right: 0;
     z-index: variables.$max-z;
     visibility: hidden;
     opacity: 0;


NadiaUoM (approved):

gregorydlogan:
While this is unlikely to affect *sane* users, we do have some additional length issues if someone does this:

https://github.com/user-attachments/assets/a0829c1a-281c-459c-a298-524abdfb96b0

PR #1340, opened 2025-06-18
Fix modal data not being loaded

Arnei:
One of the recent eslint PRs removed this by accident, causing modals to not load their data (i.e. the event details not loading metadata). This fixes that.

### DIFF
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -52,10 +52,16 @@ const NavBar = ({
 	const newResourceModalRef = useRef<ModalHandle>(null);
 
 	const showNewResourceModal = async () => {
+		if (create && create.onShowModal) {
+			await create.onShowModal();
+		}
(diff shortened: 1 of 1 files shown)


src/components/NavBar.tsx
marwyg, line 1: I think they don't do any harm, but couldn't this semicolons be removed behind the "if" block?

marwyg (changes requested):
I saw the lines that were previously removed. So this PR just introduces what was accidentally deleted I guess.

Looks good to me. I also tested it and it works. I think if the semicolons behind the if-blocks will be removed, it could be merged.

src/components/NavBar.tsx
Arnei: Done.

marwyg (approved):
Looks good now :) I would merge it like this, but I don't have this power.

PR #1352, opened 2025-06-25
Activate eslint no-unsafe-anything

Arnei:
Enables the eslint rules starting with `no-unsafe`.

Does not fix every problem theses rules surface, i.e. unchecked json parses, but it fixes some of them and therefore at least constitutes some progress.


owi92 (approved):
lgtm

PR #1360, opened 2025-07-02
Apply default actions to default roles for new events/series

### LINKED ISSUE #1354: The `default_actions` setting is not applied to the creator ACL when a new series is created.
**Preconditions:**
- custom actions are configured
- in the config file `/etc/opencast/listproviders/acl.default.create.properties` the `default_actions` parameter is configured, e.g. like this:
  ```
  # Default actions for every new role
  # Default: No actions
  default_actions=annotate,annotate-admin
  ```

**Problem:**
When I create a new series, the default actions are not automatically assigned to the series's creator ACL. However, if I manually add an additional user or role, the default actions are assigned correctly.

Image

Arnei:
Fixes #1354.

An admin can configure actions that are automatically added to each role you add to an ACL in the ACL tab for series and events. However, they were not applied to default user role the ACL is initialized with. This adds the actions to default user role.

### How to test this

Configure additional actions in your Opencast configuration. Check `etc/listproviders/acl.additional.actions.properties` and `etc/listproviders/acl.default.create.properties` and configure an action there. Then try creating an event and a series.

Bildschirmfoto vom 2025-07-02 16-26-43


snoesberger:
I tried to test this PR, but when I open the new series dialogue, I can't define the metadata and proceed to the next step.
grafik

I tried it with the same result in both our test environment and https://develop.opencast.org

Arnei:
> I tried to test this PR, but when I open the new series dialogue, I can't define the metadata and proceed to the next step. grafik
> 
> I tried it with the same result in both our test environment and https://develop.opencast.org

Likely fixed by https://github.com/opencast/opencast-admin-interface/pull/1340

snoesberger:
> Likely fixed by #1340

You are right, if I add this commit, I'm again able to create series.

However, I found an other problem. The custom actions are now added correctly for the default user. However, if I change the ACL template, the custom actions are removed again from the default user. This happens not only with new series, but also with existing ones. Additionally, when I change the template of an existing series, all the user ACLs are removed and only the editor's user ACL is added.

Arnei:
> However, if I change the ACL template, the custom actions are removed again from the default user. 

Thanks for catching that, should be fixed now.

> Additionally, when I change the template of an existing series, all the user ACLs are removed and only the editor's user ACL is added.

Afaik when switching templates, all changes made by the user are discarded. So after switching a template, only the roles specified by the template + the default user role are set. Wether that constitutes reasonable behaviour or not I cannot say. In any case, that discussion is probably better left for a different issue/PR.

snoesberger:
> > Additionally, when I change the template of an existing series, all the user ACLs are removed and only the editor's user ACL is added.
> 
> Afaik when switching templates, all changes made by the user are discarded. So after switching a template, only the roles specified by the template + the default user role are set. Wether that constitutes reasonable behaviour or not I cannot say. In any case, that discussion is probably better left for a different issue/PR.

In our case, we use the templates to regulate the access authorizations for Tobira. We use templates that contain roles which grant permissions based on users' organizational affiliations. For example, this means that we can only allow members of our institution to access certain series/events in Tobira.
Directly authorized users are owners of the corresponding series/events for us. Accordingly, we only want group authorizations to change when the template changes, not the owners. We have series with over 100 directly authorized users. If someone wants to change the template, for example, to make the series publicly available on Tobira instead of only to members of our institution, the 100+ users would lose their write access to the series.

By the way, in the old admin UI, user ACLs weren't removed when changing the template.

New issue to fix this: https://github.com/opencast/opencast-admin-interface/issues/1364

snoesberger:
As you mentioned in #1364, configuring `keep_on_template_switch_role_prefixes` in `etc/listproviders/acl.default.create.properties` should prevent the removal of user roles when switching templates. However, this isn't the case anymore with this PR.

Arnei:
Thanks for catching that, should be fixed now.

snoesberger:
Thanks @Arnei for fixing this. Default actions are now applied as expected and the configuration of `keep_on_template_switch_role_prefixes` is being considered again.

snoesberger:
I found one other minor issue. The very first time I open the create dialogue, the default actions are not applied. As soon as I select a template, the default actions are applied, as they are when I open the create dialogue a second time.

Arnei:
> I found one other minor issue. The very first time I open the create dialogue, the default actions are not applied. As soon as I select a template, the default actions are applied, as they are when I open the create dialogue a second time.

And I was so sure I had tested for that, sorry. Should be fixed now, please test again.

marwyg:
So I tested the following things:
* deleting cookies and website data (in firefox when clicking on the lock symbol next to the url)
* opening series dialog the first time -> default actions appear ✅
* opening events dialog -> default actions appear ✅

- deleting cache again (step 1)
- opening events dialaog -> default actions don't appear ❌
- opening events dialog a second time -> default actions do appear

Issue #1292, opened 2025-05-15
Improve workflow view in Admin UI


KatrinIhler:
From @marwyg:

> I would like to be able to see all jobs of a workflow in the Admin UI. Currently you can only see the operations, not the jobs that were spawned by them.
> 
> This would help when trying to find out where a workflow has failed = on which node it has failed. Maybe it's also possible to see the original Exception in the frontend. 

In our discussion we considered adding another level to see which jobs were spawned by the operation, were they ran and whether they succeeded or failed. It might also be nice to see the exception there instead of having to go back and enter a different section. See also https://github.com/opencast/opencast/issues/1223 in regards to what should be shown where and how.

PR #1362, opened 2025-07-08
Mini workflow tab rework

Arnei:
Helps with #1292.

Makes various changes to the workflow tab in the event details modal.

- When opening the workflow tab, start on the details page for the latest workflow.
  - Most of the time, users only care about processing status or failures of the latest workflow
  - Other workflow of course remain accessible
- Put links to operations at the top instead of the bottom and display the latest operation
  - Should make it easier to check workflow status
- Put errors just below operations. Directly display errors here (if applicable).
- Sort workflows by newest first (how it was in the old admin ui).

Bildschirmfoto vom 2025-07-08 10-19-17

### How to test this

No further changes required for testing. Check if this lets you access the information you are usually looking for faster than before, without obfuscating information.

-----
EDIT #1: The latest commit shows the related workflow operation on the error details page. This requires https://github.com/opencast/opencast/pull/6888, else the operation will not show. Not having the backend PR should NOT break this PR.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowDetails.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowDetails.tsx
@@ -239,7 +239,7 @@ const EventDetailsWorkflowDetails = ({
 							{/* 'Workflow Operation table */}
 							<div className="obj tbl-container more-info-actions">
 								<header>
-									{t("EVENTS.EVENTS.DETAILS.WORKFLOW_DETAILS.LATEST_OPERATION")}
+									{t("EVENTS.EVENTS.DETAILS.WORKFLOW_DETAILS.OPERATIONS")}
 								</header>
 
 								<table className="main-tbl">
@@ -322,6 +322,14 @@ const OperationsPreview = ({
 
 	const workflowId = useAppSelector(state => getModalWorkflowId(state));
 	const operationsEntry = useAppSelector(state => getLatestWorkflowOperation(state));
+	const workflow = useAppSelector(state => getWorkflow(state));
+
+	// Parse translation key to state
+	let workflowDone = false;
+	if ("status" in workflow) {
+		const workflowStatus = workflow.status.split(".").pop();
+		workflowDone = !(workflowStatus === "SUCCEEDED" || workflowStatus === "FAILED" || workflowStatus === "STOPPED");
+	}
 
 	const loadWorkflowOperations = async () => {
 		// Fetching workflow operations from server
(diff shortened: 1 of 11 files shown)


mtneug:
Looks nice for a quick improvement. Some thoughts:

> When opening the workflow tab, start on the details page for the latest workflow.

For me personally, this will speed up things. Not sure how intuitive it is for other users to navigate to other workflows. But let's just try and see if people complain.

> Put links to operations at the top instead of the bottom and display the latest operation

Big +1 for putting this on top. Not sure about displaying the latest operation. I guess the idea is to maybe show the failed operation? But when using exception handling workflows, this will not the the last operation. And showing `cleanup` in almost all cases doesn't help much? Although it might be handy to see the current operation during the workflow run (is latest operation == current operation?).

Arnei:
> Big +1 for putting this on top. Not sure about displaying the latest operation. I guess the idea is to maybe show the failed operation? But when using exception handling workflows, this will not the the last operation. And showing cleanup in almost all cases doesn't help much? Although it might be handy to see the current operation during the workflow run (is latest operation == current operation?).

The idea was to display the currently running operation, or the operation that ran last, to give users a quick impression of the state of the workflow without them having to look at the whole list of operations (and thus saving a click and a bit of time). I can see that it does not help much to if every completed workflow just shows `cleanup`. We could only display the "current" operation while the workflow is running and hide it otherwise?

The idea was not to show the failed operation. Ideally I would like to display the failed operation alongside the related error, but that is proving to be difficult.

Arnei:
"Current" operation is now hidden if the workflow is done.

src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowDetails.tsx
```diff
@@ -303,57 +304,96 @@ const EventDetailsWorkflowDetails = ({
...
+	// Parse translation key to state
+	let workflowDone = false;
+	if ("status" in workflow) {
+		const workflowStatus = workflow.status.split(".").pop();
+		workflowDone = !(workflowStatus === "SUCCEEDED" || workflowStatus === "FAILED" || workflowStatus === "STOPPED");
+	}
+
+	const loadWorkflowOperations = async () => {
+		// Fetching workflow operations from server
+		dispatch(fetchWorkflowOperations({ eventId, workflowId }));
```
gregorydlogan: This line doesn't have the workflow ID when it fires.  I'm seeing a 404 at `http://localhost/admin-ng/event/ID-portrait-1310279468/workflows//operations.json`, with a stack trace like this
```
(anonymous)	@	eventDetailsSlice.ts:1407
u	@	EventDetailsWorkflowDetails.tsx:336
(anonymous)	@	EventDetailsWorkflowDetails.tsx:341
```

src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowErrors.tsx
gregorydlogan, line 1: Similar to the above, I'm seeing a 404 on `http://localhost/admin-ng/event/ID-portrait-1310279468/workflows//errors.json` with a stack trace like this:
```
(anonymous)	@	eventDetailsSlice.ts:1464
(anonymous)	@	EventDetailsWorkflowErrors.tsx:50
```

gregorydlogan (changes requested):
A couple of unexpected errors, but the UI does work as expected - there are repeat calls to the relevant endpoints which occur *after* the workflow ID has been loaded which do work correctly.  I wouldn't have noticed these if I hadn't had the dev console open.

src/components/events/partials/ModalTabsAndPages/EventDetailsWorkflowDetails.tsx
Arnei: Thanks for reporting that, now it should not send faulty requests anymore.

PR #1365, opened 2025-07-09
Fix ACL role dropdown width

Arnei:
The role dropdown in the ACL tabs could extend over its column width, due to its own fixed width.
This fixes that.

Before:
Bildschirmfoto vom 2025-07-09 14-37-02

After:
Bildschirmfoto vom 2025-07-09 14-36-35


gregorydlogan:
Considering that this affects 17 (right?), should it be pointed there?

Arnei:
> Considering that this affects 17 (right?), should it be pointed there?

Thanks for pointing that out. I created a new PR against 17 here: #1370

PR #1368, opened 2025-07-10
Remove READ column from ACL modal UI

priya-tik:
This PR removes the READ column from the Access Control List (ACL) modal user interface.
image


gregorydlogan:
Discussion at the technical meeting suggested that this reads like the user can't do anything in the ACL.  I'm inclined to agree.  @Arnei suggested in the technical meeting that this was possible with config changes alone though, so perhaps this becomes something where we just change hte default?

Arnei:
To be precise, I did not say you can configure the removal of the READ column. I said that you can configure the read/write columns to be read-only. If read-only, the respective columns are still visible, but the checkboxes are greyed out and uninteractable. 

Config file: https://github.com/opencast/opencast/blob/develop/etc/listproviders/acl.default.create.properties
The READ column is read-only per default, you can see this behaviour on e.g. develop.opencast.org.

PR #1369, opened 2025-07-10
Reduce duplicate code in main page

Arnei:
Adds a single main page component, so pages don't have to import the different headers and footer again and again. Also adds a table page component, since almost all of our pages are just a big table.

###

Can be tested as usual, should result in no functional changes.


gregorydlogan (changes requested):
~Doesn't build, likely from changes in #1347 ~

This was caused by #1371, just ignore me...

PR #1370, opened 2025-07-11
Fix ACL role dropdown width

Arnei:
The role dropdown in the ACL tabs could extend over its column width, due to its own fixed width. This fixes that.

Replaces #1365

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx
@@ -269,7 +269,7 @@ const NewAccessPage = <T extends RequiredFormProps>({
 																						}
 																						skipTranslate
 																						optionHeight={35}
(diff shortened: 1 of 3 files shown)


wsmirnow (changes requested):
Event and series details dialog works now as expected. But new event and new series dialog is not fixed.

image

Organization > Access policies > Add access policy and details dialog is broken too.

@Arnei can you fix it please?

Arnei:
The new event/series/acl dialogs look fine to me. Can't yet think of reason why it is different for you.

Bildschirmfoto vom 2025-07-14 14-10-20

gregorydlogan:
I can reproduce Waldemar's experience, even after clearing my cache.

Arnei:
@wsmirnow Helped me fix my confusion, thanks for that. Should now also properly work for in the New Event/Series/Acl modal and the Acl Details modal.

wsmirnow (approved):
Works for me. Thank you @Arnei

Issue #1230, opened 2025-05-13
Performance issue with metadata


ppettit:
Open the metadata dialog (editing or creating an event) causes a huge file download and is very slow.

`/admin-ng/event/new/metadata` gets loaded  (including all series data) 4 times when opening the create event dialog, closing the dialog loads it a further 2 more times!

https://github.com/user-attachments/assets/f3ccc6d3-9b4b-4bc8-b974-3262f76f2454

Opening an existing event loads metadata only once but it is still slow because it loads all series metadata

https://github.com/user-attachments/assets/ad45dab3-764b-4819-9640-d0eddec0fb42

It would be better if as well as removing the duplicated loads we could use an autocomplete to search for series and not load the entire set.

Arnei:
As far as I can tell the biggest time waster is this code in the backend https://github.com/opencast/opencast/blob/e11426559dc3ad229f59b1e0d2f945ade7591b5a/modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/provider/SeriesListProvider.java#L157. It checks if two or more series have the same title and if true, modifies the series title to be unique. It does this for *all* series. On my system that takes about 4~5 seconds for about 10000 series, which makes for most of the time of the webrequest. It would already help a fair bit if we could find a way to not do that.

The query to get all series itself still takes a solid second on my system (for about 10000 series), so we'd still be pretty slow even if the above was not an issue.

For me, `/admin-ng/event/new/metadata` is fired once when opening the Create Event dialog and not at all when closing it. Maybe this has been fixed in the meantime.

I am also not seeing all series metadata being loaded when opening the dialog for an existing event. Only a huge array of "Series Title: Series ID" key value pairs. Which request is loading the metadata?

ppettit:
Hey @Arnei thanks for taking a look! I'm away from computers for a couple of weeks so can't look properly until I get back 😊.

I think the multiple requests were probably due to running the frontend in dev mode which I wasn't aware of at the time so that's probably not a problem now. 

I'll be back at work on July 7th so I'll do some more testing then.

EnsiyehE:
Hello everyone,

I've been looking into this issue and came up with a possible solution that I’ve implemented on the frontend from a user experience perspective.

It’s not perfect yet, but the idea is to lock the page using an overlay with a busy indicator while the metadata is still loading after clicking on the "Edit Event" panel. This prevents user interaction until the necessary data is available.

[...]

PR #1375, opened 2025-07-22
Make event metadata load faster by fetching series options asynchronously

Arnei:
***Breaks without the associated backend changes https://github.com/opencast/opencast/pull/6924***

Hopefully helps with #1230.

If you have many series in your Opencast, any modal that displays event metadata can take several seconds to load. This patch aims at pushing the loading time back into millisecond terrritory no matter how many series there are.

The problem is that when fetching event metadata from the backend, the backend also fetch all series options, which can take too long. So we change the backend to not do that, and change the frontend to fetch the series options separately.

Ideally we could also do this for other metadata fields with options (notably contributors and presenters), but the endpoint lacks the necessary filter capabilities.

### How to test this patch

Install together with the required backend PR. Try to change series for new or existing events.
If you want to check for performance benefits, use an Opencast with ~10000 series.

[Bildschirmaufzeichnung vom 2025-07-22 13-54-35.webm](https://github.com/user-attachments/assets/be516dfc-9611-4dab-b8f7-971c23fded11)

### DIFF
--- a/src/components/shared/DropDown.tsx
+++ b/src/components/shared/DropDown.tsx
@@ -45,7 +45,7 @@ const DropDown = <T, >({
 	ref?: React.RefObject<SelectInstance<any, boolean, GroupBase<any>> | null>
 	value: T
 	text: string,
-	options: DropDownOption[],
+	options?: DropDownOption[],
 	required: boolean,
 	handleChange: (option: {value: T, label: string} | null) => void
 	placeholder: string
@@ -66,7 +66,7 @@ const DropDown = <T, >({
 		optionPaddingTop?: number,
 		optionLineHeight?: string
 	},
-	fetchOptions?: () => { label: string, value: string}[]
+	fetchOptions?: (inputValue: string) => Promise<{ label: string, value: string }[]>
 }) => {
 	const { t } = useTranslation();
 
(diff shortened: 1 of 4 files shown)


gregorydlogan:
Should this be in develop when the backend changes will land in `r/18.x`?

When applied (front and back-end), new events don't seem to load the series at all - no series are ever listed.

Arnei:
Yes, they should go into the same branch, thanks for pointing that out.

Unless, of course, they are broken. Will check.

Arnei:
Not quite sure why the build workflow is failing now. Builds on my machine. The issue with the series not loading should be fixed now though.

gregorydlogan:
Could this be node related?  The workflows install a 20.x release, and when I use 20 I see this:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'vite@7.0.3',
npm WARN EBADENGINE   required: { node: '^20.19.0 || >=22.12.0' },
npm WARN EBADENGINE   current: { node: 'v20.12.2', npm: '10.5.0' }
npm WARN EBADENGINE }
```

Arnei:
Probably not? If I read the workflow logs correctly the workflow is using node v20.19.4 which should be fine with vite 7.

PR #1376, opened 2025-07-23
Fix selection of capture agent input selection in add events dialog

EnsiyehE:
Fix issue with user-selected inputs not being saved when posting new events

This PR addresses a problem where user choices for device inputs were not persisted despite appearing correctly in the form submission. The fix involved ensuring the correct property from the form values is accessed and included in the request payload.

To further improve code quality and prevent regressions, I can add unit tests covering this behavior in a follow-up PR if desired.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/EventDetailsSchedulingTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/EventDetailsSchedulingTab.tsx
@@ -92,7 +92,6 @@ const EventDetailsSchedulingTab = ({
 
 	const sourceStartDate = new Date(source.start.date);
 	const endStartDate = new Date(source.start.date);
-
 	useEffect(() => {
 		dispatch(removeNotificationWizardForm());
 		dispatch(checkConflicts({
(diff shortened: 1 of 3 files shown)


Arnei (approved):
Code looks reasonable to me and works. I think it would be adequate to have this target r/17.x instead of develop.

Now we just need to wait for Apereo to accept the CLA. This can sometimes take some time, so no need to worry.

Edit: You can disregard the github-actions message, it just pops up again if the "Check ICLA" workflow is run again. You don't need to act.

lkiesow:
I've just checked. The ICLA has been submitted on 2025-07-23. Can be merged.
Just checked: The test didn't work since the GitHub username in the ICLA was not actually the GitHub username but the link to the GitHub profile. Fixed that. From now on, the test shouldn't fail again.

Since @Arnei approved the code, I'm mergin this.

Arnei:
This got merged into `develop`, but I'm sure a lot of users would be happy to see this in `r/17.x`. Can we get it in there too?

PR #1379, opened 2025-07-25
Add Unit Tests for Application and New CI Workflow for Vitest Coverage

EnsiyehE:
This PR introduces the following enhancements to the application:

1. Unit Tests
Added new unit tests, including coverage for some components, thunks and events inside the application.

 2. New CI Workflow for Vitest Coverage
Created a new GitHub Actions workflow to run vitest with coverage reporting.

The CI job is configured to run automatically when a new version is released.

Coverage settings are added in package.json to keep it maintainable and version-aware.

 Review Request:
Please review the added unit tests and CI workflow. I'd appreciate your feedback on:

The test logic and structure

The setup of the CI job and coverage reporting

Any improvements or missing cases you notice

Thanks for your time and input!


src/components/events/partials/wizards/NewEventWizard.tsx
```diff
@@ -35,7 +35,10 @@ const NewEventWizard = ({
 }) => {
 	const dispatch = useAppDispatch();
 
-	const uploadSourceOptions = useAppSelector(state => getSourceUploadOptions(state));
+	const uploadSourceOptions = useAppSelector(state => {
+	console.log(state)
```
Arnei: Leftover logging statements in this file.

vitest.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+    test: {
+        environment: "jsdom",
+        globals: true,
+        setupFiles: "tests/setup.ts",
+    coverage: {
+        provider: "istanbul",
```
Arnei: Afaik v8 is the recommended coverage provider. Why choose istanbul?

vitest.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
```
Arnei: Afaik vitest can be configured in `vite.config.ts`. Why create a separate config?

tests/components/navBar.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+        org: { id: "org1", properties: {} },
+      },
+      events: {
+      uploadSourceOptions: [],
+      uploadAssetOptions: [],
+      isFetchingAssetUploadOptions: false,
+      metadata: { title: "event", flavor: "someFlavor", fields: [] },
+      extendedMetadata: [],
+      total: 0,
+  },
```
Arnei: Indentation is inconsistent in several places. I'm mentioning this here once, but I noticed this in other files as well.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,38 @@
+name: Vitest Coverage
```
Arnei, line 1: Why should we add code coverage reports to our releases? I feel like it will just unnecessarily bloat our releases with files that no one will ever look at.

tests/components/shared/tableSetupStore.tsx
```diff
@@ -0,0 +1,51 @@
...
+    notifications: [],
+  }),
+};
+
+export const store = configureStore({
+  reducer: rootReducer,
+});
+
+
+function renderWithProviders(
```
Arnei: The different setup store files read very similar to each other. Could we reduce duplicate code here, particularly in regards to `renderWithProviders`?

tests/components/usersSetupStore.tsx
```diff
@@ -0,0 +1,104 @@
...
+},
+  tableFilterProfiles: {profiles: []},
+  table: rootReducer.table(),
+};
+
+export function renderWithStore(ui: React.ReactElement, preloadedState = {},  route = '/users/') {
+  const testStore = configureStore({
+    reducer: userRootReducer,
+    preloadedState: {
+      ...defaultPreloadedState,
```
Arnei: Having defaultPreloadedState that may or may not gets partially overwritten by a preloadedState reads to me like it would lead to complicated bugs down the line. I would have each test pass the state itself, even if for now that means they're all passing the same state. What do you think?

.github/workflows/coverage.yml
Arnei, line 1: What about a workflow that runs the tests on pull requests? I imagine we don't want to merge pull requests if they break our tests.

tests/components/shared/table.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      ...rootReducer.table(),
+      pagination: {
+        ...rootReducer.table().pagination,
+        offset: 1,
+      },
+      pages: [
+        { number: 0, label: "1", active: false },
+        { number: 1, label: "2", active: true },
+      ],
+    }),
```
Arnei: I am not all experienced with writing these kinds of tests, but this looks like a lot of work for simply changing redux state. Isn't there a simpler way to achieve this, maybe using preloadedState?

tests/components/shared/tableFilters.test.tsx
```diff
@@ -0,0 +1,52 @@
+import { screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import renderWithProviders from './tableFiltersSetupStore';
+import TableFilters from '../../../src/components/shared/TableFilters';
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+const mockLoadResource = createAsyncThunk('resource/load', async () => Promise.resolve());
+const mockLoadResourceIntoTable = () => (dispatch:any) => {}; // dummy thunk does nothing
```
Arnei: Please avoid `any` if possible.

tests/components/users.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+          ],
+        })
+      );
+    }
+  }
+]);
+
+describe('username', () => {
+
+it('shows same name in table and header', async () => {
```
Arnei: Since the users displayed in the user table are unrelated to the currently logged in user, there is no reason to test if the same name is displayed both in the table and in the header. If you want to test if the header properly renders the name of the currently logged in user, I think that would be better done in a separate test for the Header.

tests/server.ts
```diff
@@ -0,0 +1,37 @@
...
+type HttpMethod = 'get' | 'post' | 'put' | 'delete';
+
+interface HandlerConfig {
+  method?: HttpMethod;
+  path: string | RegExp;
+  res: (
+    req: RestRequest,
+    res: ResponseComposition,
+    ctx: RestContext
+  ) => any;
```
Arnei: Avoid `any` if possible.

package.json
```diff
@@ -65,13 +67,20 @@
...
     "@types/lodash": "^4.17.19",
     "@types/node": "^22.15.18",
     "@types/react-dom": "^19.1.6",
     "@types/react-window": "^1.8.8",
     "@types/uuid": "^10.0.0",
     "@vitejs/plugin-react-swc": "^3.10.2",
-    "eslint": "^9.30.0",
+    "@vitest/coverage-istanbul": "^3.2.4",
+    "@vitest/ui": "^3.2.4",
+    "eslint": "^9.28.0",
```
Arnei: Looks like the eslint version got downgraded accidentally.

Arnei (changes requested):
Thanks for introducing unit tests to the admin interface. The overall approach looks good I think, but I still have a couple comments and questions for you (Some of which likely come across as terse. I assure you I am not trying to be rude though).

Running `npm test` produces a couple warnings and errors in the console. Are they expected? If so, would it be preferable to handle them in some way so as not to confuse others?

Please adapt the gitignore file to include files generated by the tests and coverage, e.g. `/coverage`.

The `npm run build` command is failing. Looks like eslint expects different syntax. Could you fix that?

```
Oops! Something went wrong! :(

ESLint: 9.32.0

A config object is using the "env" key, which is not supported in flat config system.

Flat config uses "languageOptions.globals" to define global variables for your files.

Please see the following page for information on how to convert your config object into the correct format:
https://eslint.org/docs/latest/use/configure/migration-guide#configuring-language-options

If you're not using "env" directly (it may be coming from a plugin), please see the following:
https://eslint.org/docs/latest/use/configure/migration-guide#using-eslintrc-configs-in-flat-config
```

tests/components/shared/tableSetupStore.tsx
EnsiyehE: I totally see the intention behind a generic render helper, but since our components often need very different store setups, trying to cover all cases in one function ends up adding boilerplate and complexity. It might be clearer and easier to maintain if we have separate, focused render helpers tailored to each feature’s store needs. This keeps tests simpler and more readable. What do you think?

vitest.config.ts
EnsiyehE: No worries, you’re right! Sorry about that—I actually added the Vitest config directly in the vite.config.ts file to keep things simpler. Thanks for pointing it out!

vitest.config.ts
EnsiyehE: Istanbul fitted my needs well because it provides detailed coverage per component, which helps me understand exactly what’s covered. Since I’m mainly using it for generating reports in CI, it works perfectly for now. But if you think switching to v8 would be better, please let me know!

tests/components/usersSetupStore.tsx
EnsiyehE: Thanks for the feedback! I understand the concern about partial overrides potentially causing issues. The approach I’m using provides a stable default state so components can render reliably, and tests only override what they need. This reduces duplication and keeps tests clearer. While there’s some risk, maintaining a well-defined default state helps minimize bugs. Each test can still fully specify its state if necessary.

tests/components/shared/table.test.tsx
EnsiyehE: I’m replacing just the table slice of the reducer here to have precise control over the pagination state in this test, while keeping the rest of the store intact. This seemed like a straightforward way to adjust only the part I needed without recreating the whole store.

I know using preloadedState is often a cleaner approach, but in this case, I felt that replacing the reducer slice was simpler for dynamically changing the state in one place. That said, I’m still learning and this is my first experience writing tests professionally, so I’m definitely open to suggestions or better ways to handle this!

tests/components/users.test.tsx
EnsiyehE: Thanks for the feedback! What I intended to do was to verify that the logged-in user’s name (displayed in the header) matches the name shown in the user table’s (name column). To make my intention clearer, I can share a snippet or picture showing how the user name appears in both places. This way, it’s easier to see what exactly I’m testing.

If you’d like, I can provide that snippet to clarify further.
pictureExample

vitest.config.ts
Arnei: I have no experience with either and was simply interested in your reasons for choosing istanbul. I'm assuming v8 does not work as well with CI or something? Anyway, thanks for sharing.

tests/components/shared/tableSetupStore.tsx
Arnei: I asked because the store setups did not seems very different but instead very similar to me. I agree, a generic setup should not be too generic, and some components may require very specialized setups that warrant their own setup code. Still, the `renderWithProviders`/`renderWithStore` and `configureStore`/`createTestStore` look to me like they could be abstracted into generic methods?

tests/components/usersSetupStore.tsx
Arnei: In the end this is semantics. I'm not against having a default state object, I'd just prefer seeing it used differently. Instead of having the default state added within the `renderWithStore` method, I would prefer if each test would explicitly pass the default state object to `renderWithStore` as preloadedState. This results in barely any code duplication and forces developers to express their intent ("Yes, I explicitly want this default state for my test"). It also makes it easier to pass a completely different state if desired, as you don't have to make sure to explicitly overwrite everything from the default state you don't want.

tests/components/shared/table.test.tsx
Arnei: Ah right, preloadedState would not work for dynamic changes, my bad. 

What about the reducers from tableSlice? Could we use `setPages()` or `setOffset()` instead of hacking the state?

tests/components/users.test.tsx
Arnei: I understand what you are testing. I do not think we should be testing it. It is not a requirement for user name in the header to also appear in the user table, or vice versa.

tests/components/shared/tableSetupStore.tsx
EnsiyehE: Thanks for the suggestion! I see your point—there definitely are similarities in the setup files. But I’ve found that keeping a separate renderWithProviders per component makes things more flexible and easier to reason about.

Each component can have slightly different needs—some require a specific preloadedState, others need routing, or even custom middleware. Trying to make a single shared helper to cover all of those cases tends to add a lot of conditional logic, which can make the tests harder to follow and maintain.

Redux Toolkit’s documentation actually encourages creating custom render helpers per component or test group, rather than forcing one global helper:
[🔗 https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function](https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function)

That said, if we start seeing repeated logic inside those helpers, we can definitely extract the common parts—like createTestStore—to keep things DRY without losing clarity.

What do you think?

tests/components/usersSetupStore.tsx
EnsiyehE: Thanks for your input! The reason I included a default state inside the renderWithStore helper is because most of the state slices are not directly related to the specific test goals, but we still need to pass them. This is because some parts of the Redux store depend on the global state structure, including slices that aren’t directly related to the component under test. Without providing this full structure, the component or connected logic might fail or behave unexpectedly.

This approach helps keep the tests focused and avoids repeating large amounts of irrelevant state in every test, while still ensuring the store shape is complete. Of course, I’m open to changing the pattern if you still prefer that each test explicitly passes the entire state — I want to find the best way that balances clarity and maintainability.

Let me know what you think!

tests/components/users.test.tsx
EnsiyehE: Here is the screenshot showing the logged-in user’s name updated in both the user table and the header after the change. This illustrates my intention to test that the header reflects the current user’s updated information consistently with the table.

Let me know if this helps clarify the test’s purpose!

example1111111

tests/components/shared/table.test.tsx
EnsiyehE: Also, just to add — when testing with Redux, it’s generally not recommended to mock hooks like useDispatch or useSelector. Instead, using a real test store with controlled state (like we do here, in this case it is only the structure not the real one in table as it is for sure clear in its setup store ) tends to give more reliable and realistic test results without unnecessary mocking complexity.

This is why I prefer directly manipulating the reducer state for tests rather than mocking dispatch or selectors.

Of course, happy to discuss other approaches if you think they might work better!

package.json
EnsiyehE: Sorry about that! I’ll fix the ESLint and Prettier issues using my VS Code setup, and also configure npm test not to show the warnings you mentioned. As for the build issues — it's a bit of a classic, but it does build fine on my side. That said, I’ll dig into it and figure out what’s going wrong.

If there's anything else that needs to be adjusted based on our earlier discussions, I’ll make sure to cover that too.

tests/components/shared/tableSetupStore.tsx
Arnei: > Each component can have slightly different needs—some require a specific preloadedState, others need routing, or even custom middleware. Trying to make a single shared helper to cover all of those cases tends to add a lot of conditional logic, which can make the tests harder to follow and maintain.

And I am trying to say that this it not the case here, that many components actually require an exceedingly similar setup and that for those components at the very least it would be nice to not have to copy and paste code all the time.

> Redux Toolkit’s documentation actually encourages creating custom render helpers per component or test group, rather than forcing one global helper:
[🔗 https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function](https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function)

Might be me being blind, but I don't see where it says that.

tests/components/usersSetupStore.tsx
Arnei: I still think what I said in my previous post.

tests/components/shared/table.test.tsx
Arnei: What's the capital of Madagascar?

I was not referring to mocking something, or at least that was not my intention. I was asking about using the reducers we get from the respective slice when setting up store in `configureStore`, if that is possible and if it would be useful.

EnsiyehE: Hi @Arnei,

Thanks again for your suggestions!

Just to clarify — I’m fully aware that Redux state is immutable. What I’m doing here is not mutating the state but providing a minimal mock of the TableState shape (based on the TableState type from tableslice.tsx) that the Table component expects to render properly in tests.

Since my goal is to verify whether the pagination buttons are enabled or disabled depending on the page state, I chose to mock only the relevant parts of the table slice state — like pages, pagination.offset, and a few others — to keep the test simple and focused.

I’m not using the full actual slice reducer or dispatching actions in this test. Instead, I override the reducer with a simple function returning the desired mock state. This way, I can simulate being on the first or last page directly.

This approach seemed more straightforward for this UI-specific test than involving thunk actions or mocking API calls, especially since I’m not concerned here with side effects or async data fetching.

I’m still quite new to frontend testing (about 2 months), so if you think there’s a more idiomatic or professional way to handle this, I’d really appreciate your advice and am happy to adjust!

Thanks again for your guidance and support!

tests/components/shared/tableSetupStore.tsx
EnsiyehE: Hi @Arnei,

Thanks for the discussion! I originally understood the recommendation to have separate render helpers per component as a way to keep concerns fully separate in unit testing.

For example, Navbar tests need preloaded state, User component tests require routing setup, while Table tests don’t rely on the real Redux slice or full store state—just the structure needed to support the UI behavior under test.

That sentence from the docs —
“Instead of copy-pasting the same store creation and Provider setup in every test, we can use the wrapper option in the render function and export our own customized renderWithProviders function that creates a new Redux store and renders a <Provider>, as explained in [React Testing Library's setup docs](https://testing-library.com/docs/react-testing-library/setup#custom-render).” — and [🔗 https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function](https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function)
along with other resources, made me think that “every test” here refers to each test block, not an integration test of Redux itself, but rather unit testing UI features of specific components.

If my understanding is correct, I’m happy to extract and share the common part of the renderWithProviders setup (store creation and Provider wrapper) while keeping component-specific needs separate.

Please let me know if you agree with this approach!

Thanks again for your guidance.

tests/components/usersSetupStore.tsx
EnsiyehE: > I still think what I said in my previous post.

Hi @Arnei,

Thanks for your continued feedback! I see your point about explicitly passing the state in each test.

However, in practice, each test would then need to repeat a very large structure of preloaded state—even when they only need a few slices—because the whole component must render first to produce the DOM or React tree we query (via getBy, findBy, etc.). Without this full state shape, rendering often breaks or behaves unpredictably.

So, having a stable default state inside the setup helps keep tests focused, reduces duplication, and avoids bugs caused by incomplete store shape. This is especially true when many slices aren’t directly related to the feature under test but are required for the app’s Redux structure. ( Which in the case of my undergoing test i found it true ! ) 

If my reasoning makes sense, I believe keeping the default state in the setup store is justified here.

Please let me know if you agree or if you see a better way to balance clarity, maintainability, and robustness.

Thanks again!

tests/components/shared/table.test.tsx
Arnei: Thanks for the detailed explanation. I was indeed under the impression that you were setting up our test store with the actual reducers from the different slices, although you really were just setting a minimal state. Sorry for the trouble and thanks for bearing with me, I am also still newish to frontend testing.

Then this makes sense to me. Personally, I would write the test somewhat like below, with the main difference being that we simulate clicking on the "next" button instead of directly changing redux state. For sure, that is not strictly necessary if all we want to this is if the buttons are enabled or disabled based on redux state. But it helps us test if our app is behaving like we expect (e.g. clicking the "next" button should switch to page 2). And helps me as a test writer, because I need to know less about how exactly the redux state causes the buttons to be enabled/disabled. All I care about as a tester is that the next button is disabled on the last page, not what value `offset` needs to be set to for that to happen, right? Let me know what you think!  

  
Below is my different approach to this test. Note that this is using a different `renderWithProviders`, which takes a preloadedState and has all reducers from `store.ts`.
```
describe('Pagination in Table Component', () => {

  const initialTableState = {
    rows: [],
    columns: [],
    pagination: {
      offset: 0,
      limit: 10,
      totalItems: 20,
      directAccessibleNo: 5,
    },
    pages: [
      { number: 0, label: "1", active: true },
      { number: 1, label: "2", active: false },
    ],
    resource: "default",
    sortBy: {},
    reverse: {},
    multiSelect: {},
    status: "idle",
  }

  it("Previous button is disabled on first page, nextButton on the second page", async () => {

    const { container } = renderWithProviders(
      <Table templateMap={{}} />,
      { preloadedState: {
        table: initialTableState,
      }}
    );


    let prevButton = container.querySelector(".prev");
    let nextButton = container.querySelector(".next");

    expect(prevButton).toHaveClass("disabled");
    expect(nextButton).not.toHaveClass("disabled");

    const user = userEvent.setup();
    await user.click(nextButton);

    prevButton = container.querySelector(".prev");
    nextButton = container.querySelector(".next");

    expect(nextButton).toHaveClass("disabled");
    expect(prevButton).not.toHaveClass("disabled");
  });
})
```

tests/components/shared/tableSetupStore.tsx
Arnei: I think that approach sounds good. We can probably heavily adapt from the docs you linked to get there.

In fact, I tried my hand at writing a more common `renderWithProviders` based off of the docs, just to see if it would work. It does not consider routing and is probably not optimal, but the general direction seemed about right to me.

Code, because why not. You should probably not directly copy this, it's just for inspiration.
```
import React, { PropsWithChildren, JSX } from "react";
import { render } from "@testing-library/react";
import { Provider } from "react-redux";
import { RenderOptions } from "@testing-library/react";
import { reducers, RootState } from "../../../src/store";
import { configureStore } from "@reduxjs/toolkit";

interface ExtendedRenderOptions extends Omit<RenderOptions, 'queries'> {
  preloadedState?: Partial<RootState>
  store?: AppStore
}
export type AppStore = ReturnType<typeof setupStore>

export const setupStore = (preloadedState?: Partial<RootState>) => {
  return configureStore({
    reducer: reducers,
    preloadedState
  })
}

function renderWithProviders(
  ui: React.ReactElement,
  {
    preloadedState = {},
    // Automatically create a store instance if no store was passed in
    store = setupStore(preloadedState),
    ...renderOptions
  }: ExtendedRenderOptions = {}
) {
  function Wrapper({ children }: PropsWithChildren<{}>): JSX.Element {
    return <Provider store={store}>{children}</Provider>
  }
  return { store, ...render(ui, { wrapper: Wrapper, ...renderOptions }) }
}

export default renderWithProviders;
```

tests/components/users.test.tsx
Arnei: 1. Your test would not achieve this. In `defaultPreloadedState`, the name of the user is "Test User".  Then, the user returned from mocking `'/admin-ng/users/users.json'` still has the name "Test User". So you cannot confirm whether the name actually changed or not, since it is still the same name.

2. We have no reason to force a relationship between the user reflected in the header and the users in the user table. They get their information from different endpoints, and how that information may or may not related in the backend is of no concern to us.
- Testing a user is properly displayed in the user table: :+1: 
- Testing the current users name is properly displayed in the aheder: :+1: 
- Testing the user name in the header also appears in the user table: :-1:

tests/components/usersSetupStore.tsx
Arnei: I do agree that we need to provide a default state, and that we don't want copy and paste the whole state definition all the time.

Instead of this
```
export function renderWithStore(ui: React.ReactElement, preloadedState = {},  route = '/users/') {
  const testStore = configureStore({
    reducer: userRootReducer,
    preloadedState: {
      ...defaultPreloadedState,
      ...preloadedState,
    },
  });
```

I would like this:
```
export function renderWithStore(ui: React.ReactElement, preloadedState = {},  route = '/users/') {
  const testStore = configureStore({
    reducer: userRootReducer,
    preloadedState:
  });
```
And in the tests:
```
it('some test', async () => {
  renderWithStore(<Users />, defaultPreloadedState, '/users/');
[...]
}
```

That way should still provide the full state shape in time so rendering does not break. Again, I feel this makes intent clearer and easier to pass completely different states if necessary. It also makes `renderWithStore` more similar to the other `renderWithStore`, which in turn should make it easier to abstract them into a common function.

On a slightly different note, there is a lot of the initialization in `defaultPreloadedState` for slices we do not want to care about. Like, we are fully initializing the state for `tableFilters`, with the same values as in `initialState` in `tableFilterSlice.ts`. We could avoid that by adding the reducer from `tableFilterSlice.ts` (or simply all reducers from `store.ts`) to our `configureStore` call. This sets all variables to the values from their initialState, allowing our `defaultPreloadedState` to become much slimmer.

.github/workflows/coverage.yml
pascalseeland: It was my idea to give the topic of testing a bit more visibility. If we only create them as artifacts of e.g. each action run, I think nobody will ever look at them. This was just a first idea. I think there are other options as well. Maybe just publishing them using github.io might also be a possibility. Other ideas are also welcome.

EnsiyehE:
I have fixed whatever we discussed according to your solutions , but still build command would give me error regarding something related to Linux , because my machine now is windows as an operating system , i am not sure how i can fix it , i would appreciate your hints and code review again Arnei , thank you !

.github/workflows/coverage.yml
Arnei: Adding test coverage to the releases is likely to get ignored, especially since we're aiming to automate away releases of the admin interface to Opencast as much as possible.

Another idea would be to add the coverage value to the Opencast release notes, somehow. Although I am not sure we want to proudly announce 5% code coverage in our release notes, so maybe not ^^'.

May be better if we could automatically post to github discussion, or even the matrix channel? No idea how to do that though.

EnsiyehE: With all due respect, what I see here already represents nearly 20% of the entire application—a considerable portion of a clearly large codebase. I believe this level of coverage provides a solid starting point that could even be built on by myself.

This is simply my opinion, shared with the intention of contributing constructively.

src/store.ts
```diff
@@ -35,61 +43,87 @@ import autoMergeLevel2 from "redux-persist/lib/stateReconciler/autoMergeLevel2";
...
-	eventDetails,
-	themeDetails,
-	seriesDetails,
-	recordingDetails,
-	userDetails,
-	groupDetails,
-	aclDetails,
-	userInfo,
-	statistics,
+  tableFilters,
```
Arnei, line 81: Please respect that we are using tabs for indentation instead of whitespaces. We are aware that this is not ideal and have plans for eventually switching to whitespaces. But pull requests should not arbitrarily change the current convention.

package.json
```diff
@@ -65,13 +66,22 @@
...
     "@types/react-window": "^1.8.8",
     "@types/uuid": "^10.0.0",
     "@vitejs/plugin-react-swc": "^3.10.2",
+    "@vitest/coverage-istanbul": "^3.2.4",
+    "@vitest/ui": "^3.2.4",
     "eslint": "^9.30.0",
+    "eslint-plugin-vitest": "^0.5.4",
+    "jsdom": "^26.1.0",
+    "msw": "^1.3.5",
+    "prettier": "^3.6.2",
```
Arnei: We can certainly discuss if we want to introduce an opinionated code formatter to the project, but let's do so in an issue separate from this PR. As such, please remove `prettier` from this PR and undo the automatic formatting it applied to previously existing files.

tests/components/shared/table.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+        { number: 1, label: "2", active: false },
+      ],
+      resource: "events",
+      sortBy: dummyResourceState,
+      reverse: dummyReverseState,
+      multiSelect: dummyMultiSelect,
+      status: "loading",
+      error: null,
+      predicate: "",
+      maxLabel: "",
```
Arnei, line 39: I wonder if we can shorten this state even more. Since we are using the reducer from `tableSlice.ts`, variables that we don't care about like maxLabel should already have an acceptable default value. And since preDefaultState is now typed as DeepPartial, we *should* in theory only need to define the variables we need, right?

src/store.ts
```diff
@@ -110,6 +144,8 @@ export type AppThunk<ReturnType = void> = ThunkAction<
   RootState,
   unknown,
   UnknownAction
->
+>;
+
+export const rootReducer = reducers;
```
Arnei, line 149: Why not directly export `reducers`? This feels needlessly convoluted.

tests/components/usersSetupStore.tsx
```diff
@@ -0,0 +1,22 @@
+import { configureStore } from "@reduxjs/toolkit";
+import { rootReducer } from "../../src/store"; // real root reducer
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router";
+import { render } from "@testing-library/react";
+
+export function renderWithStore(
```
Arnei, line 7: The renderWithStore from `setUpStore.tsx` now support routing, right? Why not use it here then?

tests/components/shared/tableFilters.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+        stats: [],
+        textFilter: [],
+        selectedFilter: "",
+        secondFilter: "",
+        currentResource: mockResource,
+      },
+      tableFilterProfiles: {
+        profiles: [],
+      },
+    } as DeepPartial<TestRootState>;
```
Arnei: Instead of casting here, we could now write `renderWithProviders<TestRootState>`. Or was that now why you made that generic?

tests/components/shared/tableFilters.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const user = userEvent.setup();
+
+    // initial preloaded state with empty textFilter
+    const preloadedState = {
+      tableFilters: {
+        data: [],
+        stats: [],
+        textFilter: [],
+        selectedFilter: "",
+        secondFilter: "",
```
Arnei: I wonder if we can avoid having to write most of these default initializations, as the slices should be doing that anyway.

tests/components/users.test.tsx
```diff
@@ -43,15 +43,17 @@ createServer([
...
+    expect(userInTable).toBeInTheDocument();
+  });
+
+  it("displays current user's name in the header", async () => {
+    renderWithStore(<Users />, {}, "/users/");
 
-    expect(tableCell).toBeInTheDocument();
-    expect(headerButton).toBeInTheDocument();
+    const headerUserName = await screen.findByText(/Test User/i);
+    expect(headerUserName).toBeInTheDocument();
```
Arnei: This is the same test twice. It just checks if it can find the text "Test User" anywhere, not if it's specifically in the header or in the table.

Arnei (changes requested):

package.json
Arnei: Version is fixed.

tests/components/users.test.tsx
Arnei: There are now two separate tests for header and table :+1: (although probably not quite working yet, see below).

tests/components/navBar.test.tsx
Arnei: Indentation in the new test files now looks good :+1: (Please see comment below about prettier and not mass-formatting unrelated files).

.github/workflows/coverage.yml
Arnei: I did not mean to belittle your work, this is definitely a large step up from our previous situation. Still, whether it is 5% or 20%, neither of these numbers is nearly large enough to inspire confidence in our adopters (and especially new adopters). Instead, I fear that it would have quite the opposite effect. Therefore, I think we should consider carefully how we want to draw attention to this.

tests/components/shared/tableSetupStore.tsx
Arnei: Setting up store for a given component is now easier :+1:

Arnei:
> I have fixed whatever we discussed according to your solutions , but still build command would give me error regarding something related to Linux , because my machine now is windows as an operating system , i am not sure how i can fix it , i would appreciate your hints and code review again Arnei , thank you !

It now builds fine, at least on my machine ^^'. Whatever you did seemed to help, thanks for that!

tests/components/shared/table.test.tsx
EnsiyehE: Hi Arnei 

I understand the suggestion, but in this case we're deliberately providing the full TableState because the test depends on multiple properties (like pagination, pages, etc.). Using the exact type here ensures correctness and clarity.

The goal of DeepPartial in our broader testing setup is to avoid enforcing full app state for unrelated slices — not to skip required fields when we're explicitly testing a specific slice. So from my side as the developer, reducing the state here feels like an unnecessary change — though I really appreciate you raising it!

src/store.ts
EnsiyehE: You're absolutely right, and I’ll work on removing Prettier from this PR and undoing the formatting changes as requested.

Apologies for the confusion — I think the formatting on my end may have used 4 tabs, which might be getting interpreted as spaces by others. Outside of using a formatter, I don’t really have another reliable way to keep indentation consistent, but I’ll do my best to adjust it manually this time. Thanks for the clarification!

package.json
EnsiyehE: You're absolutely right, and I’ll work on removing Prettier from this PR and undoing the formatting changes as requested.

Apologies for the confusion — I think the formatting on my end may have used 4 tabs, which might be getting interpreted as spaces by others. Outside of using a formatter, I don’t really have another reliable way to keep indentation consistent, but I’ll do my best to adjust it manually this time. Thanks for the clarification!

tests/components/users.test.tsx
EnsiyehE: Hi Arnei, apologies — I’d prefer to address all the issues related to users.test.tsx here in one place, as splitting them up gets a bit confusing from my side.

First, you're absolutely right — both tests are currently testing the same thing. I’ll revise the second one to check a different behavior in the component.

Second, as we discussed last week, my reason for keeping the setup local to this test is due to the component’s use of MSW. In this case, I’m intercepting API calls directly within the component’s context, which helps the test better reflect real-world behavior and makes it more robust. I understand the preference for a global testing setup, but in scenarios like this — where a route parameter or dynamic setup is involved — using a global configuration would introduce unnecessary indirection. I personally don’t see an issue with letting some components define their own renderWithProviders, especially when MSW is in use.

Third, regarding the rootReducer: I needed access to the actual combined reducer to configure the test store. Redux doesn't expose a clean way to get the inferred reducer type automatically, so exporting the rootReducer from the store was the most practical solution in this case. It allows me to keep the test aligned with the actual application store structure. I did this based on your earlier suggestions to keep a consistent store configuration across tests, but here I chose to use the full reducer structure because I wanted to test the connection between the users slice and related slices like userInfo and header. I realize this may be more advanced Redux usage, but from my perspective it’s a practical solution that serves the goal of integration-style testing.

So my recommendation would be to keep the renderWithProviders setup locally for this component’s tests — it helps keep things focused, avoids overcomplication, and supports the MSW setup better.

Thanks again for your feedback — let me know what you think, and I’ll proceed with the test updates accordingly.

tests/components/shared/tableFilters.test.tsx
EnsiyehE: To the best of my understanding — and I fully acknowledge I have less professional experience with Redux compared to you — using preloadedState or an initial state when rendering a specific component is necessary in cases like this.

In this particular test for the table component, the goal is to simulate dispatching an action (e.g., editing text). That requires the correct state shape to be present, especially since the initial state is already defined in the createSlice. Additionally, this component depends on the tableFilterProfiles state, either directly or indirectly — for example, it might render other components internally that rely on related filter states.

I completely understand your concerns about consistency and code structure, but I personally feel that these are more related to Redux-level details rather than to the core purpose of this PR, which is focused on component testing. Of course, if this setup proves inconsistent or problematic, I’m happy to revisit and refactor it — but my intention was to keep the test realistic and straightforward without overcomplicating the Redux structure just for the sake of testing.

Let me know what you think, and I appreciate your input as always.

tests/components/shared/tableFilters.test.tsx
EnsiyehE: The point of this setup is to keep each component test isolated and able to define only the parts of the store it depends on, rather than forcing a global shape or mocking irrelevant parts. That’s especially useful when writing multiple tests across components with different slice needs. It's not about enforcing stricter typing — it's about practicality in testing small, focused units.

Arnei: I'm not sure what you think my question was, but I am specifically talking about the type assertion `as DeepPartial<TestRootState>` (not type casting, excuse my Java). The assertion which is specifically there to pacify the typescript compiler. And that we could avoid entirely by passing `TestRootState` to `renderWithProviders`, because you explicitly changed the latter for that to be possible.

tests/components/shared/tableFilters.test.tsx
Arnei: I was not asking whether we can completely avoid having a `preloadedState` or an `initialState`. I agree that we absolutely need to have some state in redux. Instead I was wondering if we could slim down our `preloadedState` object here, as we are now using the ´tableFilters.reducer` from `tableFiltersSlice.ts`, which *should* come with it's own initial state. Or am I misunderstanding how that works?

tests/components/users.test.tsx
Arnei: > Second, as we discussed last week, my reason for keeping the setup local to this test is due to the component’s use of MSW. In this case, I’m intercepting API calls directly within the component’s context, which helps the test better reflect real-world behavior and makes it more robust. I understand the preference for a global testing setup, but in scenarios like this — where a route parameter or dynamic setup is involved — using a global configuration would introduce unnecessary indirection. I personally don’t see an issue with letting some components define their own renderWithProviders, especially when MSW is in use.

So from what I can tell, we want to pass `initialEntries` to `MemoryRouter`, but `renderWithProviders` from `setUpStore.ts` does not currently allow for that. Instead of changing `renderWithProviders` to allow for that, we write our own `renderWithProviders` since this is a specialized case. Eh, sure, fine by me. Can still revise it in the future should this special case become more common.

tests/components/shared/tableFilters.test.tsx
EnsiyehE: Thanks for clarifying! You’re right — using <TestRootState> on renderWithProviders would remove the need for as DeepPartial<TestRootState>. My initial reasoning for the assertion was to keep the preloaded state flexible and avoid repeating the generic everywhere, since the function already accepts DeepPartial. Both approaches work, but I agree that your suggestion improves type safety without much overhead. I’ll update the test to use the generic.

tests/components/shared/tableFilters.test.tsx
EnsiyehE: Hey Arnei,
You’re right — I realized that I was overestimating what needed to be in the preloadedState. It actually depends on the component being rendered, and for this test, we only really need the part of the state that the component interacts with.

I’ve now slimmed down the preloadedState to include only what’s strictly necessary for the TableFilters component to work and for the test to correctly apply the feature we’re testing (i.e., updating textFilter). Everything else from the slice will default from the slice’s initial state, so no unnecessary clutter remains.

Thanks for pointing this out — it definitely makes the test setup cleaner and more focused.

Arnei:
@EnsiyehE are you planning to address the outstanding requested changes?

A brief overview:
- Undo changes to formatting in unrelated code (for example switching tabs and whitespaces, line breaks).
- Directly export reducers from store.ts, unless there is an argument for not doing that.
- Use renderWithStore from setUpStore.tsx in usersSetupStore.tsx, if it makes sense to do so.
- Remove coverage.yml, or continue the discussion.
- Say what you think about including a github workflow that runs the tests on pull requests.

Arnei:
- This PR is too complex for what it does. Even tests that perform simple tasks like the ones provided are already difficult to understand. Therefore I would expect that other developers would not engage with them, and developers who break a test will have a hard time understanding what went wrong, ultimately meaning that the tests will cost us more time than they might save us.

Of course, if anybody disagrees, feel free to reopen this.

PR #1381, opened 2025-07-25
Textfilter reload on invalid

priya-tik:
This PR adds a guard in the `getTextFilter` selector to safely handle cases where `textFilter` is not an array. This prevents the following runtime error from crashing the app

### DIFF
--- a/src/selectors/tableFilterSelectors.ts
+++ b/src/selectors/tableFilterSelectors.ts
@@ -17,7 +17,14 @@ export const getFilters = (state: RootState, resource: string) =>
 export const getTextFilter = createSelector(
 	[getAllTextFilter, (state, resource: Resource) => resource],
 	(textFilter, resource) => {
-		const textFilte = textFilter.find(obj => obj.resource === resource);
-		return textFilte?.text ?? "";
+		if (Array.isArray(textFilter)) {
+			const textFilte = textFilter.find(obj => obj.resource === resource);
+			return textFilte?.text ?? "";
+		} else {
+			console.warn("clearing localStorage and reloading...");
+			localStorage.clear();
+			window.location.href = window.location.href + "?forceReload=" + Date.now();
+			return "";
+		}
 	},
 );


Arnei:
I have my reservations against this. First of all, if we want to ensure that textFilter can only ever be an array, a selector is not the place for that. The reducers should take care of that. Secondly, I don't want to add a catch block that deletes local storage everytime we change the type of a variable in local storage to every selector (and potentially other places) that use that variable.

priya-tik:
Thanks. You're right that selectors should stay pure and not cause side effects like clearing storage or triggering reloads. I added the check to prevent the crash users are seeing, but I agree it's better to fix this earlier, either in the reducer when loading state, or with app versioning logic. I'll move the validation out of the selector and update the PR. Thanks for the feedback

lkiesow:
I'd like to apply the “Be pragmatic” rule to #1395 or this pull request. I understand that it is annoying to deal with changed data structures in the browser's local storage, but the reality is that we have to if we don't want users to end up with a completely broken interface. I got contacted a bunch of times about develop.opencast.org being broken. Usually by technical people. They couldn't fix that locally (they didn't realize that they had to drop their in-browser storage) and just assumed Opencast was down.

I'm also fine with working out a more generic solution for this, like dropping the local storage automatically if users run into an exception. But we can't just reject valid solutions for the problem at hand just because there might be a better one. That's something for later. Opencast is kind of broken for users now.

Can we merge one patch and bring it in today's release?

mtneug:
IMO it's a better solution to clear the local storage automatically if there is data in an unexpected format. I'm would therefore prefer to merge this PR instead of #1395.

I'm fine with this going into today's release. I created #1414 to track the "better solution".

lkiesow:
I agreed and started testing this patch. Now I recommend we merge the other patch after all. This patch causes the UI to end up in an endless reload loop:


https://github.com/user-attachments/assets/7f6afe11-2e0d-41cf-aa84-3f0f736a9293

gregorydlogan:
We went with yet another alternative in #1418

PR #1383, opened 2025-07-30
Fix Metadata field getting cut off

Arnei:
Some metadata fields would not display their full content even if there was still space available. This patch should help with that.

Before:
Bildschirmfoto vom 2025-07-30 09-08-13

After:
Bildschirmfoto vom 2025-07-30 09-07-54

### How to test

Check that inputs/textareas in modals do not behave strangely.

### DIFF
--- a/src/styles/components/modals/_modal-base.scss
+++ b/src/styles/components/modals/_modal-base.scss
@@ -283,6 +283,7 @@
                 line-height: 25px !important;
                 padding-top: 4px;
                 padding-bottom: 4px;
+                word-break: break-word;
 
                 > a {
                     float: none;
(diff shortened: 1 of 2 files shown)


Piriya-VM:
Looks good to me. :)

grafik

No Code Review!

gregorydlogan (changes requested):
Hammering the keyboard and hitting enter randomly looks good on the Metadata tab, but when I get to the summary I see this:

image

The data is all there, and the modal will scroll right for this:

image


FTR, the metadata tab looks correct to me:
image

Arnei:
Added a commit that should handle terribly long words on the summary page.

PR #1385, opened 2025-07-30
Safari: Dropdown for multiselects opens immediately

Arnei:
In Safari, open the details of an event and click on one of the multi-select fields (e.g. presenters, contributors). This will open up a dropdown which hides already selected values. This is not desirable. Like in other browsers, the dropdown should only open after clicking the field an additional time.

This patch fixes the issue by removing autoFocus attribute from the input element. This has an annoying side effect in that while you now need two clicks in Safari, you need three clicks in other browsers (e.g. Firefox). Arguably a better solution would be to get rid of `<datalist>` entirely, as it is not well supported. But this quick fix should be acceptable for now.

### How to test this

Can be tested as usual. Test with different browsers, Safari and otherwise.

### DIFF
--- a/src/components/shared/wizard/RenderMultiField.tsx
+++ b/src/components/shared/wizard/RenderMultiField.tsx
@@ -166,7 +166,6 @@ const EditMultiSelect = ({
 						onChange={(e) => handleChange(e)}
 						placeholder={t("EDITABLE.MULTI.PLACEHOLDER")}
 						list="data-list"
-						autoFocus={true}
 					/>
 					{/* Display possible options for values as some kind of dropdown */}
 					<datalist id="data-list">


gregorydlogan:
I don't have Safari, but I do have Chrome and Firefox, neither of which seem to be behaving quite right.

Chrome:
Two clicks, but it's unclear if those suggestions are coming from Opencast, or my browser's suggestion list.  The ones at the bottom present in the particular test instance I'm using, but there's a space between them at the top ones which makes me wonder if Chrome is being helpful and adding its suggestions after the ones fetched from the system?

https://github.com/user-attachments/assets/6822ead7-cc04-44ae-895d-b9104394ace4


Firefox:
Similar to Chrome, except it takes *three* clicks to get the suggestions.  Same test instance, notably missing suggestions like `orga`, and `lkjh`.

https://github.com/user-attachments/assets/b4a4e4eb-9502-454c-b521-05d81c5d3659

In neither case does the UI look quite right to me - I would expect a React style dropdown the way that the License field works.  I guess I don't care if it looks this way, but it's out of place when compared to the rest of the UI.

Arnei:
Ah heck, thanks for testing. Then this PR just might not be the right approach and we'll have to get rid of `<datalist>` alltogether. That'll be more work though.

gregorydlogan:
Given that 17.x is pretty much EOL at this point, should this get either punted forward to 18, or refiled when it's ready?

Arnei:
I'd say refile it. I'll open an issue so that the problem does not get lost.

Issue #950, opened 2024-10-11
dialog: keyboard navigation


marcusm7:
- As mentioned in https://github.com/opencast/opencast-admin-interface/issues/65: the tab leaves the dialog and walks through any element of the table.
- There should be a way to navigate from one element (text field, radio button,...) to the next in any dialog. I.e. by means of the cursor keys or a sane tab-cycle (which then cycles through all elements of the dialog in order). 
In the create event dialog, this works really nice in the tab-bar and partially in the source-dialog, but i.e. not in the metadata - I can't reach series or presenter by means of the keyboard.
- ESC: Hitting Esc loses all the entered data without a chance to overthink your decision. Although I know this... behaviour, I accidentally close the dialog from time to time. The decision to do it this way is not at all common in other web-UIs. A hint like "Esc again to omit all entered data" or so would be great.

Arnei:
I believe all three points have been addressed over various pull requests, so I'm closing this.

PR #1387, opened 2025-07-30
Fix/modal keyboard nav 950

EnsiyehE:
This PR improves keyboard accessibility and behavior in modal dialogs, addressing the concerns raised in issue https://github.com/opencast/opencast-admin-interface/issues/950

Fixes tab navigation: Focus is now properly trapped within the modal, allowing users to cycle through form fields (e.g., text inputs, radio buttons) without jumping outside the dialog.

Improves Esc key behavior: Pressing Esc now shows a confirmation before closing the dialog, helping prevent accidental data loss.

These changes enhance usability and keyboard navigation consistency across all modals, including the metadata and source dialogs.

Tested in Google Chrome.


Arnei:
> Fixes tab navigation: Focus is now properly trapped within the modal, allowing users to cycle through form fields (e.g., text inputs, radio buttons) without jumping outside the dialog.

https://github.com/opencast/opencast-admin-interface/pull/1339 already added focus trapping by introduced the "focus-trap-react" library.

> Improves Esc key behavior: Pressing Esc now shows a confirmation before closing the dialog, helping prevent accidental data loss.

I am against adding this behaviour. Forcing users to *always* click away a dialog every time they want to close the modal reduces usability in my opinion. Furthermore, I expect users to eventually become numb to the (most of the time) pointless warning, automatically clicking it away without thinking, rendering the warning pointless.
I prefer showing such a warning only if there is actual data to be lost. While this is much more work, it is also much more useful to users. We already have this in a few places (event details metadata page I think), but it could definitely be expanded and improved upon.

EnsiyehE:
> > Fixes tab navigation: Focus is now properly trapped within the modal, allowing users to cycle through form fields (e.g., text inputs, radio buttons) without jumping outside the dialog.
> 
> #1339 already added focus trapping by introduced the "focus-trap-react" library.
> 
> > Improves Esc key behavior: Pressing Esc now shows a confirmation before closing the dialog, helping prevent accidental data loss.
> 

Hi,

Regarding pressing ESC in the NewResourceModal, I think the focus trap might still be enabled because after pressing ESC once, the modal seems to lose focus and it’s not possible to tab through all the input fields properly.

Could you please check this behavior and let me know? If the current modal setup isn’t desired, I can either remove that part or submit a new PR specifically to improve the tab navigation and focus trap handling.

Thanks for your time!

Arnei:
Let me see if I understood you. When the user first opens a NewResourceModal (for example the "Create Event" modal), focus is trapped inside the modal. When the user then focus an input or textarea field and pressed Escape, the field loses focus but also the focus trap is deactivated and the focus is now god-knows-where but not inside the modal anymore. And you want to prevent the focus from leaving the modal if Escape was pressed? If that is the case I think I agree. Pressing Escape on an input element probably should not let the focus leave the modal. Though that raises the question where the focus *should* go in that case.

EnsiyehE:
Hi,

Thanks for pointing that out! Regarding the ESC key: if we imagine a scenario where the user is prevented from closing the NewResourceModal on the first ESC press (or click), that part works as  i described and it will lose the focus from the modal window ( does not have focus trap anymore)

Separately, about the tab navigation—if you try tabbing through the inputs in the current main or develop branch, you’ll notice that the issue with navigating through all fields properly remains unresolved. meaning that you can only go until the language with pressing the tab key on your keyboard as it was described in the issue 


Thanks again!

Arnei:
Being able to properly tab through the metadata keys has nothing to do with pressing ESC though?

I think I am a bit confused on what you want to achieve, as it seems you want to achieve multiple things here. Would you do me a favor and summarize again what exactly your goals are?

EnsiyehE:
Hi,

In this fix for issue 950, my goal was to improve keyboard accessibility. Previously, tab navigation would stop at the language select element and not reach the rest of the fields. I’ve updated the focus trap so that tabbing now works properly across the entire modal,  and also as a second trial after pressing Esc to trigger the confirmation modal warning and the focus is still trapped inside the modal

Thanks for your time!

Arnei:
Thanks for clarifying. If I understand you correctly, there are three goals here.

1. Some form fields (i.e. the language selector) are not keyboard accessible. This patch makes them keyboard accessible. :+1:, that's great.
2. When pressing ESC while in a modal (and the modal does not close due to this event), the focus goes to some nonsensical element outside the modal. This patch somehow prevents that. :+1:, sounds good.
3. When pressing ESC while in a modal, the modal is closed. This patch adds a confirmation dialog that always appears before closing, to prevent accidental data loss. :-1:, I consider this bad UX as detailed in my first comment. Therefore I would like you to remove or change this.

EnsiyehE:
Hello,

Thank you very much for your reply ,


I have tried to remove the modal implementation for the warning message as you said for the losing data.

so this solution is only solving the keyboard navigation via pressing Tab button on your keyboard when your in the newResource modal.tsx

src/components/events/partials/ModalTabsAndPages/NewMetadataPage.tsx
```diff
@@ -18,7 +18,7 @@ const NewMetadataPage = ({
 	header?: ParseKeys
 }) => {
 	const { t } = useTranslation();
-
+	console.log(metadataCatalogs);
```
Arnei: Looks like a leftover logging statement.

src/components/shared/DropDown.tsx
```diff
@@ -209,6 +209,7 @@ const DropDown = <T, >({
 		<AsyncSelect
 			ref={selectRef}
 			{...commonProps}
+            openMenuOnFocus={false}
```
Arnei: Please respect that we are using tabs for indentation instead of whitespaces. We are aware that this is not ideal and have plans for eventually switching to whitespaces. But pull requests should not arbitrarily change the current convention.

src/components/shared/NewResourceModal.tsx
```diff
@@ -11,67 +11,79 @@ import { Modal, ModalHandle } from "./modals/Modal";
 /**
  * This component renders the modal for adding new resources
  */
-export type NewResource = "events" | "series" | "user" | "group" | "acl" | "themes";
+export type NewResource =
```
Arnei, line 14: Please don't reformat code unrelated to the topic of the commit (or even PR). It makes reviewing a lot more work, and future `git blame`s a lot more difficult.

Arnei (changes requested):
Kind of works. I can now correctly tab onto, over or into dropdowns in the metadata tab, which is nice. Dropdowns don't automatically open anymore when tabbed onto, but that might be for the better anyway?

I can also tab onto and over multiselect fields (e.g. presenters, contributors). However, I cannot edit multiselect fields anymore. Tab/Enter/Space will not trigger the edit mode, and clicking onto the field with the mouse will only briefly flash the edit mode before closing it again. This needs to be fixed, the comments further below are just nitpicks.

Bildschirmfoto vom 2025-08-06 16-19-58


Tested with Firefox 141.

EnsiyehE:
Hi Arnei 

You're right — I’ve been mistaken.

I misunderstood the original question, and I ended up fixing the tabbing issue inside the Create Event modal, specifically the problem where focus wasn’t moving correctly after the language field. So while it turns out the issue wasn’t actually about this part, I did discover and fix a real problem in the Add Event flow anyway — so something good came out of it!

Let me know if you'd like me to adjust this part of the code as well cause i have not seen the problem there at all.

tabbing issue after language field
this had problamatic tabbing issue

Arnei:
I would still like you to fix all of the changes I requested above. In particular, fixing tabbing over Dropdown-fields at the cost of completely breaking MultiSelect-fields is not acceptable. (Note: Multi-Select fields currently only appear in the "Event details" modal, not the "Create event" modal).

src/components/shared/NewResourceModal.tsx
EnsiyehE: What changed and why:

I originally added a warning modal in this file as part of an earlier attempt to handle user confirmation, but after we agreed this would hurt user experience, I removed that change. The only remaining differences in this file are formatting-related:

Line endings: They were converted from CRLF (Windows default) to LF, because my development environment is on Windows and Git warned about inconsistent line endings.

Indentation: A few lines may have used spaces instead of tabs unintentionally. This happened because I usually rely on Prettier for consistent formatting, but since installing Prettier isn’t allowed in this project, I had to format manually, which increases the risk of such mistakes.

I’ll make sure to correct the indentation by converting spaces back to tabs and keep LF endings to match the repository’s conventions.

EnsiyehE:
Hi @Arnei 

Summary of changes:
I worked on fixing the multi-select field behavior inside the Create Event window. Initially, I was addressing an issue where contributors and presenters were still displayed incorrectly. If I misunderstood the problem or caused confusion regarding the multi-select functionality, I sincerely apologize. Could you please explain that to me ??

Request:
Please review the updated code when you have time.
Thank you for your support and feedback!

Arnei:
Can confirm that this works now. I can access and change all metadata fields with keyboard as well with mouse.

src/components/shared/wizard/RenderMultiField.tsx
```diff
@@ -197,14 +201,28 @@ const ShowValue = ({
...
 	return (
-		<div onClick={() => setEditMode(true)} className="show-edit">
+	<div
+  		tabIndex={0}
+  		onClick={() => setEditMode(true)}
+		onFocus={() => setEditMode(true)}  // <-- activate edit mode on focus
+  		onKeyDown={e => {
+    	if (e.key === "Enter" || e.key === " ") {
+      	setEditMode(true);
+      	e.preventDefault();
```
Arnei: Please do not mix spaces and tabs when indenting. If the file is using the tab character for indentation, please continue to do so.

If you have trouble telling which character is used for indentation, I recommend checking the settings of your IDE or text editor. Most will let you display whitespace characters, and some will let you configure which whitespace character (and how many) are used when pressing the `tab` key on the keyboard (per project, per file).

Arnei (changes requested):

PR #1391, opened 2025-08-01
More explanations for embedd modal

### LINKED ISSUE #1272: Embedcode Dialog needs more text for screen readers
Image

Only the dimensions of the videos at the counters are read out here. It is not clear what pressing the switch does. 
The field in which the embed code is entered also has no labelling, so it is unclear what the field is used for.

Overall, the dialogue could do with more labels or explanations. Even if it is an admin interface, the UI should be as self-explanatory as possible.

Arnei:
Fixes #1272.
This mostly adds additional screen reader text to the embedding code modal, but also some tooltips and
an explanatory text on the modal itself.

### How to test this

Ideally you try to go through the modal with a screen reader and report if and what is still unclear.


gregorydlogan:
This seems to work well for me, though I'm not a regular screen reader user.

gregorydlogan (approved):

PR #1394, opened 2025-08-04
Fix crowdin upload branch name

Arnei:
Since we adapted the branching structure of the main opencast repository we do not use main anymore. This patch adapts the run command from the same workflow in the mian repo.


Arnei:
Merging this on account of it being infrastructure.

PR #1395, opened 2025-08-05
Fix filter cache breaking interface

lkiesow:
Investigating people complaining develop.opencast.org being broken, it seems like there was a change to the filters which cannot handle the previous data which might still be stored in browsers. Like here means, it completely crashes the interface and users just see a blank page.

The error they get is:

```
Uncaught TypeError: t.find is not a function
    fk tableFilterSelectors.ts:23
    D reselect.mjs:647
    i reselect.mjs:584
    P reselect.mjs:659
    i reselect.mjs:584
    u TableFilters.tsx:55
    <anonymous> Redux
    L use-sync-external-store-with-selector.production.js:40
    h use-sync-external-store-with-selector.production.js:63
    React 2
    useSyncExternalStoreWithSelector use-sync-external-store-with-selector.production.js:74
    n Redux
    qc TableFilters.tsx:55
    React 14
    K scheduler.production.js:152
tableFilterSelectors.ts:23:32
```

This patch makes the code a bit more resilient so it can deal with the previous content.

EDIT: Rebased to `r/18.x` since the problem also appears there.

### DIFF
--- a/src/selectors/tableFilterSelectors.ts
+++ b/src/selectors/tableFilterSelectors.ts
@@ -20,7 +20,7 @@ export const getFilters = createSelector(
 export const getTextFilter = createSelector(
 	[getAllTextFilter, (state, resource: Resource) => resource],
 	(textFilter, resource) => {
-		const textFilte = textFilter.find(obj => obj.resource === resource);
+		const textFilte = (textFilter || []).find(obj => obj.resource === resource);
 		return textFilte?.text ?? "";
 	},
 );


Arnei:
Addresses the same issue as #1381, albeit a bit differently.

I'd really rather not have to account for both the old and the new type everywhere in the code in case the type of a variable that is stored in local storage changes between versions. That sounds like a headache.

While this indeed allows the admin ui to load, the value in local storage for textFilter is still of the wrong type, no? And that could still cause the admin ui to break when other code than the selector is interacting with it, e.g. when fetching events or when trying to change the textFilter value?

lkiesow:
I'd like to apply the “Be pragmatic” rule to #1381 or this pull request. I understand that it is annoying to deal with changed data structures in the browser's local storage, but the reality is that we have to if we don't want users to end up with a completely broken interface. I got contacted a bunch of times about develop.opencast.org being broken. Usually by technical people. They couldn't fix that locally (they didn't realize that they had to drop their in-browser storage) and just assumed Opencast was down.

I'm also fine with working out a more generic solution for this, like dropping the local storage automatically if users run into an exception. But we can't just reject valid solutions for the problem at hand just because there might be a better one. That's something for later. Opencast is kind of broken for users now.

Can we merge one patch and bring it in todays release?

gregorydlogan:
Superceded by #1418

PR #1396, opened 2025-08-07
Change text for "time in queue"

Arnei:
"Time in queue" is one of the variables shown for worklow operation details. However, the name is misleading, as this does not refer to the time the operation (or the job representing it) was waiting before starting. Instead, it represents the time all the jobs the operation has spawned were waiting.
Which is why "time in queue" so often has the value "0ms", as many workflow operations simply don't spawn an jobs.

This patch attempts to make this more clear by changing "time in queue" to "total subjobs time in queue". Suggestions are welcome.


gregorydlogan:
While I applaud the idea, I'm wondering if that's going to be too long to fit comfortably in the header of a table.  Could we change the text to `Total time in queue`, with a hover-over explaining things?  I'm assuming that's a no-no accessibility wise?

Arnei:
I think it kinda still fits, but fair enough, we don't need to cut it close (especially considering translations).

Hover text is less of an accessibility thing and more of a UX thing. We can always add screen reader text and what not to accommodate accessibility, but expecting users to hover an element they have no innate reason to hover over (unlike e.g. a button) is unintuitive for many users.

gregorydlogan:
Is this going to be moving forwards, or should we drop this?

Arnei:
What would this need to move forwards?

I don't think that the text is too long by the way, we still have quite a bit of space here:
Bildschirmfoto vom 2026-03-19 10-06-30

Issue #1077, opened 2025-01-16
Scheduling: Multiple event scheduling nitpicks


gregorydlogan:
Multiple minor issues:
- Multiple event scheduling allows scheduling events in the past.  Single event scheduling (correctly) refuses to let me do this.
- Modal does not check if the date range, and day-of-the-week settings make sense.
  - I would expect settings like Monday and Tuesday, but restricting the date range to 25-01-16 -> 25-01-17 to throw an error in the UI rather than after the final submit to the backend
  - Assuming at least one day and date combination match things work as expected.
- Changing the hour (or duration) such that the end time is tomorrow does not display the date next to the end time the way that the single-event scheduling does.  Obviously we can't put a date here, but something like `+ 1 day` or something next to it would clarify.
  - Not supporting this at all would also be an option here.  This functionality raises a bunch of fun edge cases.

EnsiyehE:
@Arnei I believe this issue is merged already , am i right ? 

it can be closed ,

bests 

Ensiyeh

PR #1398, opened 2025-08-11
fix /Scheduling: Multiple event scheduling nitpicks #1077

EnsiyehE:
I have thoroughly reviewed the issues listed in #1077 and implemented fixes addressing each point mentioned. This PR aims to improve the user experience and correctness of the multiple event scheduling feature by resolving all identified nitpicks.

Changes include:

Please review and let me know if there are any questions or further adjustments needed.


Arnei:
My testing observations so far (Note that I merged the current state of develop into this for my testing, as otherwise the datepickers were broken for me):

- I can still schedule events in the past by setting the start date in the past and the end date in the future. That should not be possible I think?
- If no day of the week is in the date range, the error "The repeat days must be within the scheduled date range." is displayed. I think this what #1077 wanted, so that works :+1: . Maybe change the error message to "At least one repeat day must be within the scheduled date range" to more precisely describe the error condition to the user?
- The "+1 day" message appears even if the end time is not on the next day, which should not happen. It seems the "+1 day" message appears if the duration is greater than an hour?

src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
```diff
@@ -366,6 +366,56 @@ const Schedule = <T extends {
 }) => {
 	const { t } = useTranslation();
 	const currentLanguage = getCurrentLanguageInformation();
+	const getEndDateForSchedulingTime = () => {
+  	const {
```
Arnei: The code in this function is not properly indented.

Arnei (changes requested):

src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
EnsiyehE: done

EnsiyehE:
> My testing observations so far (Note that I merged the current state of develop into this for my testing, as otherwise the datepickers were broken for me):
> 
  Done the same now , update the local develop first and then this branch

> * I can still schedule events in the past by setting the start date in the past and the end date in the future. That should not be possible I think?
     fixed , you can not scheduale something in the past for both multiple or single ( it was not part of this task )

> * If no day of the week is in the date range, the error "The repeat days must be within the scheduled date range." is displayed. I think this what [Scheduling: Multiple event scheduling nitpicks #1077](https://github.com/opencast/opencast-admin-interface/issues/1077) wanted, so that works 👍 . Maybe change the error message to "At least one repeat day must be within the scheduled date range" to more precisely describe the error condition to the user?

   it is changed to the same sentence you have suggested 

> * The "+1 day" message appears even if the end time is not on the next day, which should not happen. It seems the "+1 day" message appears if the duration is greater than an hour?

Now you can test it as before. After a duration that exactly leads to the next calendar day — for example, starting at 12 noon and lasting 12 hours — it should show +1 day. It should not show +1 day for durations like 11 hours and 55 minutes. This applies to both single and multiple event scheduling.

thanks 
bests 
Ensiyeh

Arnei (approved):
My concerns have been addressed. Works as expected and looks good to me :+1:

PR #1402, opened 2025-08-14
Basebutton tooltip params

priya-tik:
This PR fixes an issue where tooltipParams were ignored when rendering the tooltip in BaseButton.
Previously, placeholders like {{filterName}} in translations were not being replaced even when tooltipParams was provided.

image

### DIFF
--- a/src/components/shared/BaseButton.tsx
+++ b/src/components/shared/BaseButton.tsx
@@ -30,7 +30,7 @@ const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(({
 	if (tooltipText) {
 		return (
 			<Tooltip title={tooltipParams ? t(tooltipText, tooltipParams) : t(tooltipText)}>
-			{buttonComponent}
+				{buttonComponent}
 			</Tooltip>
 		);
 	}
--- a/src/components/shared/BaseButton.tsx
+++ b/src/components/shared/BaseButton.tsx
@@ -29,8 +29,8 @@ const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(({
 
 	if (tooltipText) {
 		return (
-			<Tooltip title={tooltipParams ? t(tooltipText) : t(tooltipText, tooltipParams)}>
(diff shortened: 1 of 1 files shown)


src/components/shared/BaseButton.tsx
```diff
@@ -29,8 +29,8 @@ const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(({
 
 	if (tooltipText) {
 		return (
-			<Tooltip title={tooltipParams ? t(tooltipText) : t(tooltipText, tooltipParams)}>
-				{buttonComponent}
+			<Tooltip title={tooltipParams ? t(tooltipText, tooltipParams) : t(tooltipText)}>
+			{buttonComponent}
```
Arnei: This looks like an accidental removal to me?

Arnei (changes requested):
Can confirm that this fixes the issue, thanks for that! Since the bug is already occurring in `r/17.x`, could you rebase your branch onto `r/17.x` and point the PR against it instead of `develop`? (Since very recently we now have the same versioning in the admin interface as we have in Opencast. r/17.x is the oldest and we forward merge our versions, so if it goes into r/17.x it will go into all versions.)

src/components/shared/BaseButton.tsx
priya-tik: This wasn’t accidental. I updated the <Tooltip> expression to tooltipParams ? t(tooltipText, tooltipParams) : t(tooltipText) because the previous version caused an error. The change ensures that tooltipParams are passed to t() only when they exist, otherwise it falls back to the plain translation string.

Arnei: I see know that the line I marked did not all make clear what I wanted to refer to. I am very sorry for the confusion, of course updating the expression as you did makes perfect sense. I was only referring to the removal of the tab character ^^`.

priya-tik: ok got it

Arnei (approved):

PR #1404, opened 2025-08-15
Prevent tobira tab crash when ID is missing

### LINKED ISSUE #1403: Event details Tobira tab not working for normal users
**Preconditions:**

Opencast is configured with a working Tobira instance.

**Problem:**

A normal admin UI user without admin rights cannot open the Tobira tab in the event details modal. As soon as the user clicks on the Tobira tab, the screen turns completely white. The following error message is shown in the browser console (OC 17):
```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at aXe (DetailsTobiraTab.tsx:210:27)
    at OE (react-dom-client.production.js:3502:21)
    at qd (react-dom-client.production.js:6239:15)
    at h2 (react-dom-client.production.js:7243:14)
    at fT (react-dom-client.production.js:10855:14)
    at gR (react-dom-client.production.js:10736:37)
    at TT (react-dom-client.production.js:10717:7)
    at bm (react-dom-client.production.js:10322:40)
    at vm (react-dom-client.production.js:11634:3)
    at FE (react-dom-client.production.js:11505:17)
```

The same test with the current develop branch shows the following similar error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at DetailsTobiraTab (DetailsTobiraTab.tsx:63:80)
    at Object.react_stack_bottom_frame (react-dom_client.js?v=32453963:17424:20)
    at renderWithHooks (react-dom_client.js?v=32453963:4206:24)
    at updateFunctionComponent (react-dom_client.js?v=32453963:6619:21)
    at beginWork (react-dom_client.js?v=32453963:7654:20)
    at runWithFiberInDEV (react-dom_client.js?v=32453963:1485:72)
    at performUnitOfWork (react-dom_client.js?v=32453963:10868:98)
    at workLoopSync (react-dom_client.js?v=32453963:10728:43)
    at renderRootSync (react-dom_client.js?v=32453963:10711:13)
    at performWorkOnRoot (react-dom_client.js?v=32453963:10359:46)
```

For the series details modal, the Tobira tab is working correctly.

owi92:
I recently observed that in some cases, the Tobira API does not send an ID for its respective asset (i.e. video or series).
I have not been able to reproduce this anymore, so it's hard to identify the unterlying issue. But since others have reported this as well, in order to prevent this leading to crashes, this now uses optional chaining for `id` operations.

This won't break anything, it just means that in cases where the (Tobira) ID is actually missing, the Opencast ID is used in links to these assets instead.

Fixes #1403

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/DetailsTobiraTab.tsx
@@ -60,7 +60,7 @@ const DetailsTobiraTab = ({ kind, id }: DetailsTobiraTabProps) => {
 
 	const i18nKey = kind === "series" ? "SERIES" : "EVENTS";
 	const prefix = kind === "series" ? "s" : "v";
-	const directTobiraLink = tobiraData.baseURL + `/!${prefix}/` + (tobiraData.id.length > 2 ? tobiraData.id.substring(2) : `:${id}`);
+	const directTobiraLink = tobiraData.baseURL + `/!${prefix}/` + (tobiraData.id?.length > 2 ? tobiraData.id.substring(2) : `:${id}`);
 
 	const getBreadcrumbs = (currentPage: TobiraPage) => {
 		const homepage = {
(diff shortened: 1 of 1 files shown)


snoesberger:
Unfortunately this PR doesn't fix #1403. I'm still facing the same problem with the same error message in the browser console.

owi92:
I will take another look and see if I can reproduce this.

owi92:
Sorry about that. Turns out, this was more or less unrelated to the missing `id`, and more to the missing `everything`. I.e. Tobira/Opencast didn't send the necessary info because the user is lacking access.

If anyone runs into this, you need to add

```
 <sec:intercept-url pattern="/admin-ng/event/*/tobira/pages" method="GET" access="ROLE_STUDIO" />
```

(use any role that you want to give access to the Tobira tabs in event details).

mtneug:
@owi92 Should this be added to the default security configuration and do we need a dedicated role?

owi92:
@mtneug yes, I think that makes sense. I'll do a follow up PR.

Issue #1406, opened 2025-08-20
Text entry is slow in the create/edit event dialogue if there are a lot of series.


snoesberger:
On an Opencast system with a large number of series (>4,000), entering text in the create or edit event dialogue box is slow. For example, when I type into the title field, the characters appear slowly, one after the other. Take a look at the attached video, in which I type randomly and quickly on my keyboard.

https://github.com/user-attachments/assets/18cb9587-23a4-4df0-9a55-f9a9cb450d57

Edit: To be more precise, the form is only slow for admins or users who have access to a lot of series.

Arnei:
> @Arnei, do you know if this is also caused by the backend code that you mentioned and fixed with https://github.com/opencast/opencast-admin-interface/pull/1375? Or is this a completely different issue?

From https://github.com/opencast/opencast-admin-interface/issues/1230#issuecomment-3200793060.

#1375 might indirectly help with this particular issue, but it will not solve this issue (and it does not aim to do so).

Arnei:
#1408 has helped with this, but I think we can do even better by making sure that our code does not cause an unnecessary amount of rerenders and only rerenders the necessary components (instead of the whole form or even modal).

PR #1408, opened 2025-08-21
Use formik 'FastField' instead of 'Field' for performance reasons

snoesberger:
On an Opencast system with a large number of series (>4,000), entering text in the create or edit event dialogue box is slow. Using the formik 'FastField' instead of 'Field' improves the performance.

Fixes https://github.com/opencast/opencast-admin-interface/issues/1406

### DIFF
--- a/src/components/shared/Field.tsx
+++ b/src/components/shared/Field.tsx
@@ -1,4 +1,4 @@
-import { Field as FormikField } from "formik";
+import { FastField as FormikFastField } from "formik";
 import { FieldAttributes } from "formik/dist/Field";
 
 /**
@@ -6,7 +6,7 @@ import { FieldAttributes } from "formik/dist/Field";
  */
 export const Field = (props: FieldAttributes<any>) => {
   return (
-    <FormikField
+    <FormikFastField
       {...props}
       onKeyDown={(event: KeyboardEvent) => {
         // Handler for basic html inputs to remove focus, if no custom component is passed


mtneug:
@snoesberger can you explain why the text inputs are slow when there are many series? This seems very strange to me, and I'm wondering if there is something else that should be optimized. The [`FastField` documentation](https://formik.org/docs/api/fastfield) has a warning before use paragraph, but I don't know enough React to really comprehend this.

snoesberger:
> @snoesberger can you explain why the text inputs are slow when there are many series? This seems very strange to me, and I'm wondering if there is something else that should be optimized. The [`FastField` documentation](https://formik.org/docs/api/fastfield) has a warning before use paragraph, but I don't know enough React to really comprehend this.

Each time you enter an additional character in one of the text fields, the entire form is rendered again. And the more series you have, the slower it gets. After searching the internet, I found a proposal to use 'FastField' instead of 'Field'. I also read the Formik documentation and saw the warning you mentioned. According to the explanations, using 'FastField' can cause problems if one field depends on another. Having checked the admin UI code, I found no such dependencies; the user password and password confirmation fields are also still working correctly. Or do you see maybe dependencies between fields that I have missed?

mtneug:
> Or do you see maybe dependencies between fields that I have missed?

I don't really dabble in frontend code. So others have to answer that. This issue was just surprising to me.

So the underlying problem seems to be that all series are loaded and the fields check for dependencies among each other during each update? I guess if fields don't depend on each other, this is a valid quickfix, but the real scaling problem is having all series in-memory / in the DOM instead of loading only a partial list based on the search term.

Arnei:
I've not used `FastField` before, but it looks like it might be a valid band-aid solution for our problem here.

I do agree that the better long term solution would be to figure out why 

> Each time you enter an additional character in one of the text fields, the entire form is rendered again.

is happening, and then make it not do that. If a field is changed, only that field needs re-rendering should re-render, not the entire form. `FastField` achieves this, but in my mind we should not even need `FastField` for this behaviour and there is a more underlying issue in our code we should fix instead. I wrote https://github.com/opencast/opencast-editor/pull/1608 to combat a similar issue in the editor, not sure how applicable this is to the admin ui though.

But maybe it's better to leave the long term solution stuff to a different PR, and just use `FastField` in the meantime (provided it does not break anything :D).

lkiesow:
Tested this and it really helps quite a bit :rocket:
If no one has a strong objection, I suggest merging this. We can always revert it if someone comes up with a better solution.

snoesberger:
I agree. If someone sees a better solution, I'd love to see that. For now, though, the admin UI is hardly usable for admins on systems with many series. Therefore, I would greatly appreciate it if this PR could be merged.

lkiesow (approved):

Issue #1076, opened 2025-01-16
Scheduling: Scheduling an event currently in progress gives you the wrong error message


gregorydlogan:
Scheduling an event which is currently in progress gives you an error that 'The event has already ended'.

Open questions:
- Is this something that's reasonable to support *at all*?  Lots of CAs wouldn't check their calendars fast enough to catch many already-in-progress events, so this is *probably* something we can just refuse to allow via the UI?
- If we do want to support this, do we have an existing string or does it need a new one?

Noticed in testing #1069

gregorydlogan:
Clarifying here, that I mean something which was scheduled yesterday, and is currently recording, rather than an event which starts *right now*.  This is handling editing of an event *during* the event, not scheduling a new event.

PR #1411, opened 2025-08-27
Fix that scheduling an event currently in progress gives you the wrong error message

EnsiyehE:
Hello all,

In this pull request, I added a new error message to address the issue described in #1076. Specifically, I improved the scheduling conflict detection by:

Keeping the existing date-only check for past events unchanged.

Adding a separate hour-only check for events that are currently in progress.

This ensures that ongoing events are correctly detected without affecting the logic for past events.."


gregorydlogan (changes requested):
While the error box is at the top of the modal, it doesn't prevent me from saving the change.

EnsiyehE:
> [gregorydlogan](/gregorydlogan)

Hi Greg,

Just to confirm my understanding — when you mentioned( in the issue #1076)  that scheduling an event which is currently in progress gives the error "The event has already ended", did you mean events that have already started but not yet finished (so they’re ongoing right now)?


If that’s the case, my “Add Event” window already handles this by showing an error message at the top of the modal and preventing the user from saving. I just want to make sure I’m interpreting the issue correctly and not misunderstanding what you meant.


Thanks for clarifying!

gregorydlogan:
Sorry, yes I was unclear.  What I meant was that I scheduled an event, and then after it starts but before it finishes I change end time.  Effectively, shortening the event while it's in progress.

In theory this is *doable*, but I would wager that most CAs don't update their schedules quickly enough to reliably be able to stop early.  Thus, we should just not allow you to shorten an event which has already started.

EnsiyehE:
> Sorry, yes I was unclear. What I meant was that I scheduled an event, and then after it starts but before it finishes I change end time. Effectively, shortening the event while it's in progress.
> 
> In theory this is _doable_, but I would wager that most CAs don't update their schedules quickly enough to reliably be able to stop early. Thus, we should just not allow you to shorten an event which has already started.

Hello Greg,

I hope you are doing well.

I have now fixed the issue you mentioned. Please test the fix as follows:

Set the start hour and minute to the current time (not a time in the past) so that the event is scheduled to start immediately.

Choose your desired end hour and minute specifically

You will see that if the event is already in progress and not yet finished, the end time cannot be shortened by the user.

Best regards,
Ensiyeh

EnsiyehE:
Hello Again ,

Sorry for the confusion i thought my previous solution might work , but my logic to disable the end time field was somehow wrong completely , now please test this PR as follow :

if you choose a time like for example : 14:23 and the event has already getting started in one minute and then you choose the end time like 15:23 or 14:36 , after the event is started , you are not able to change the end time and you will see the following error message :  " This event cannot be modified because it is currently in progress."

please review and let me know the mistakes or points 

@Arnei  @gregorydlogan 

bests 
Ensiyeh

gregorydlogan (changes requested):
Closer, but still not quite right.

I'm seeing `The schedule could not be updated: You cannot schedule an event to be in the past.`, rather than the expected message.

Also I can still change the scheduled time, even with the message on the modal.


https://github.com/user-attachments/assets/6b134eca-4bee-4816-98b9-ca97807c5441

EnsiyehE:
> Closer, but still not quite right.
> 
> I'm seeing `The schedule could not be updated: You cannot schedule an event to be in the past.`, rather than the expected message.
> 
> Also I can still change the scheduled time, even with the message on the modal.
> 
>  Screencast.from.2025-11-18.11-12-08.mp4

Hi Greg,

Thank you for your feedback. It seems I may have misunderstood the requirements. I thought we needed to prevent changes to the end time once an event had already started. As you can see in the video, when I schedule an event that is about to start in one minute, once it begins I’m unable to change or shorten the end time that I originally set.

Could you please take a look at the video and let me know your thoughts?

Regarding setting an event in the past, I believe that’s related to another PR, not this one—right?

bests 
Ensiyeh

[Aufzeichnung 2025 neu error issue-11-18 183320.zip](https://github.com/user-attachments/files/23610994/Aufzeichnung.2025.neu.error.issue-11-18.183320.zip)

gregorydlogan:
Ok, I see whats happening.  There are two (or more) bugs in play, and we're talking about different things.

- This PR, attempting to keep the user from scheduling a new event starting *right now*.  This makes sense since there's no way for a CA to get the calendar and start recording in time.  This PR does what it says, my apologies for dragging this out and misunderstanding.
- #1076 should probably be retitled "Rescheduling an event currently in progress", in the sense that we need to properly handle changes in event start/end times in the case that the event is already actively recording.  The goal here is to prevent the user from editing an existing event once it has already started since there's no guarantee that the CA is checking the calendar at all, much less checking it often enough to handle the change properly.  Neither start, nor end time modification should be allowed IMO.

So, leaving #1076 alone here, this PR does solve a valid bug.  One further thing I would test though is what happens if you schedule something  for 18:33, but don't actually click all the way through to the final submit button until 18:33 (or later)?  Perhaps insisting on a small buffer would make sense if we're looking at times?  120 seconds?  @Arnei, do we have other checks like this elsewhere?

I would hope the backend would reject requests that close to *now*, but we should probably check that *that* doesn't work too...  There were buffer requirements a long time ago, but I'm unsure if they're still enforced.

EnsiyehE:
Hi Greg,

Sorry about the confusion — I misunderstood this issue twice.

About the case where someone schedules an event for 18:33 but doesn’t click the final submit until 18:33 or later: I’m not sure what the correct solution is. Maybe we need a small buffer (like a few minutes), but I don’t know.

What I did in this PR is: if an event has already started, the user cannot shorten it anymore. I thought this made sense because the problem seemed to me that it happens in the “Create Event " Schedule tab. ( as i said maybe a confusion from my side , sorry ) 

If you think this PR doesn’t really help, I can close it and instead work on the scheduling logic inside the Event Details Modal.

Arnei:
> @Arnei, do we have other checks like this elsewhere?

I believe we do not have other checks like this in the admin interface. Can't think of anything besides scheduling that would need that.

gregorydlogan:
Rechecked that this works, let's just get this merged.

PR #1412, opened 2025-08-27
Fixing that the selection may change if events are modified in the background 

### LINKED ISSUE #1253: Selection may change if events are modified in the background
https://github.com/user-attachments/assets/80327ffc-bbf2-43a1-84d9-80ea66880796

- Go to events table
- Sort by date
- Select all
- In the background add a new event or update an event date so that there is a new event on page 1
- The selection has changed. There is now one unselected event in your list

Options:

- If one event is no longer selected, the select all checkbox should not be selected
- We could freeze (not update) the table if something is selected
- We could keep the selected events, even if they are not on the page anymore. But we should indicate that somehow

EnsiyehE:
Hello all,

This pull request fixes the events table selection behavior: when new events appear on the current page, any previously selected events that are no longer selected will correctly update the ‘select all’ checkbox to remain unselected.

This fixes #1253 

thank you for your time
Ensiyeh


gregorydlogan:
The issue here is that it's going to unselect *everything* if even a single event gets added to the list.  I think the original intent here was that the select all button was deselected, but the rest of the selected items *remain* selected.  Is that something we can do?

EnsiyehE:
> The issue here is that it's going to unselect _everything_ if even a single event gets added to the list. I think the original intent here was that the select all button was deselected, but the rest of the selected items _remain_ selected. Is that something we can do?

Hi Greg,

Thank you for the review!

I’ve updated the code accordingly so that adding a new event will now only deselect the “Select All” checkbox, while keeping the previously selected items intact.


Best regards,
Ensiyeh

Issue #1409, opened 2025-08-25
Design a landing page suitable for non-technical users


lkiesow:
Technically, the current landing page of https://admin-interface.opencast.org does, what it is supposed to do: redirect users to different versions of the admin interface. However, given the very technical look'n'feel of the site and that we do redirect new and non-technical adopters to this page, it may easily be interpreted as “the site is broken”.

This issue is basically the same as https://github.com/opencast/opencast-editor/issues/1623

### To-do

- Make the page look nice
  - Maybe pick up some elements from the editor UI?
- Explain what the versions do
- Maybe don't display a path (`./17.x`), but the corresponding version (`17.x`)?
- Build this into the CI/CD code generating the page

> @ltekbas, can you assign yourself once you accepted the invitation?

ltekbas:
@lkiesow  
Same as  issue [#1623](https://github.com/opencast/opencast-editor/issues/1623)

- the header should now be the same size, and added second row header 
- fixed the naming of the versions
- added "Select a Version" to the text container, so the user knows what to do
- overall changed some colors and designed the list as "column buttons"

<img width="449" height="344" alt="Image" src="https://github.com/user-attachments/assets/360b9839-5466-408d-8113-09b8a07a655d" />

Arnei:
This has been completed by #1415.

PR #1415, opened 2025-08-28
Landing page suitable for non-technical users

ltekbas:
**Made new Pull Request for r/17.x**
Issue: https://github.com/opencast/opencast-admin-interface/issues/1409

Edit the Generate index.html command to add css sytling for the file.
Command includs now css href, some new classes and tags and new Text for the User to know what the page is about.
Also added new assets folder to seperate newly created CSS file from the other files


.github/workflows/deploy-test.yml
```diff
@@ -111,7 +111,7 @@ jobs:
       - name: generate index.html
         working-directory: admin-interface-test
         run: |
-          echo '<html><body><ul>' > index.html
+          echo '<html><head><link rel=stylesheet type=text/css href=assets/index.css /></head><body><div class="head-container"><img src=https://admin-interface.opencast.org/develop/assets/opencast-white-B3KJ3o-a.svg /></div><div class="navbar-container"></div><div class="text-container"><p>Test deployment for the latest development versions of the Opencast admininterface.The branches listed here correspond to Opencast's own branches.</br><b>Please select a version.</b></p></div><ul>' > index.html
```
lkiesow: This looks good and should update the `index.html`. We need to copy the `index.css` to the web server, though. You placed it in the editor's code which will not make it end up at the correct place. If you look at [what is in the gh-pages branch, you can basically see what's in the web server file system](https://github.com/opencast/opencast-admin-interface/tree/gh-pages).

To copy the `index.css` to the correct place, we can add a new task like this:
```yml
- name: update CSS and other assets
  if: github.ref == 'refs/heads/develop'
  run: cp -r .github/assets admin-interface-test/
```
The `if:` causes only the newest version of the assets to be deployed

.github/workflows/deploy-test.yml
```diff
@@ -111,12 +111,16 @@ jobs:
...
           find . -maxdepth 2 -name '*_*' -type d \
             | sort -r \
             | sed 's/^\(.*\)$/<li><a href=\1>\1<\/a><\/li>/' >> index.html
           echo '</ul></body></html>' >> index.html
 
+      - name: update CSS and other assets
+        if: github.ref == 'refs/heads/develop'
+        run: cp -r .github/assets admin-interface-test/
```
lkiesow: Just noticed that my suggested path was not correct. This copies `.github/assets/`, but the `index.css` is in `.github/workflows/assets/`. My suggestion would be to actually move it to `.github/assets/`, though, since the workflow folder usually only contains workflow (YAML) files.

.github/workflows/deploy-test.yml
gregorydlogan, line 1: Subtle issue here, but **this** workflow generates the *test* instances you see in PR comments.  If you're looking to make `https://admin-interface.opencast.org` pretty then you need to do the same things to `deploy-main.yml` instead.

A word of warning, once the `gh-pages` branch is checked out (line 68 currently) then the contents of the `.github` directory are no longer accessible.  You'll want to copy the contents of `.github/assets` somewhere convenient prior to switching branches!

gregorydlogan (changes requested):

.github/workflows/deploy-test.yml
ltekbas: I added the command to the ``deploy-main.yml``
I'm not quite sure what you mean by coping the content of the assets somewhere else. I get the problem, but I don't know how to handle it

.github/workflows/deploy-main.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           rm -rf $BRANCH
           mv build $BRANCH
           #Generate an index, in case anyone lands at the root of the test domain
-          echo '<html><body><ul>' > index.html
+          echo '<html><head><link rel=stylesheet type=text/css href=.github/assets/index.css /></head><body><div class="head-container"><img src=https://admin-interface.opencast.org/develop/assets/opencast-white-B3KJ3o-a.svg /></div><div class="navbar-container"></div><div class="text-container"><p>Deployment for the latest development versions of the Opencast admininterface.The branches listed here correspond to Opencast's own branches.</br><b>Please select a version.</b></p></div><ul>' > index.html
```
gregorydlogan: This is a BASH-ism issue, and it took me some futzing to figure out.  The `'` in `Opencast's` is breaking the overall enclosure of this whole string, but simply escaping it with `Opencast\'s` *does not work* because of... subtle bash things.  This appears to be the cleanest solution short of inverting the use of `'` and `"`:

```suggestion
          echo $'<html><head><link rel=stylesheet type=text/css href=assets/index.css /></head><body><div class="head-container"><img src=assets/opencast-white.svg /></div><div class="navbar-container"></div><div class="text-container"><p>Deployment for the latest development versions of the Opencast admin interface.The branches listed here correspond to Opencast\'s own branches.</br><b>Please select a version.</b></p></div><ul>' > index.html
```

Also adjusted the paths to the CSS and header image.  Github won't serve things in a `.github` path, and I just don't trust that the suffix on the header image won't change and break things without notice.

.github/workflows/deploy-test.yml
gregorydlogan: Ok, so when this workflow starts it checks out (as an example) `r/17.x`.  This branch has `.github/assets/index.css` in it once this PR is merged.  **But**, part way through we get to modifying the `gh-pages` branch, and to do that we check it out with `git checkout gh-pages` on (currently) line 68.  This *removes* the content of `r/17.x` on disk, and replaces it with `gh-pages` - what we want, except that `.github/assets/index.css` does not exist in the `gh-pages` branch!  So we end up with the HTML file, but no css.

To solve this we need to somehow add the css file to the `gh-pages` branch, which you've tried with the workflow step after the index generation.  There are a few issues with that step:
- `.github/assets/index.css` does not exist at this point, so we can't copy it anywhere.  We can solve this by copying/moving it somewhere else *prior* to `git checkout gh-pages`, then copying/moving that copy to the right place since the `checkout` only touches files that are added to the git history.
- You copy the file into the right place, but git doesn't actually pay attention to it unless you tell it to.  You need to `git add $filename`
- You only add the assets for `develop`.  This works great as long as the workflow gets run in `develop` right away, but when I'm testing this *only* in `r/17.x` it doesn't add the assets.  This took me *way* longer than it should have to figure out haha.  I like this approach though, since presumably any future branding change lands in `develop`, and probably not in the older versions - we don't want an older version's branding clobbering the newer version's branding since whoever ran most recently would be the branding applied.

I'll make suggestions, see what you think.

.github/workflows/deploy-main.yml
```diff
@@ -81,5 +81,9 @@ jobs:
         env:
           BRANCH: ${{needs.detect.outputs.branch}}
 
+      - name: update CSS and other assets
+        if: github.ref == 'refs/heads/develop'
+        run: cp -r .github/assets admin-interface-test/
```
gregorydlogan: Github won't let me make suggestions past line 70, so I'm going to put it here instead.  Insert this at 68 (before the `git checkout`):
```
cp -r .github/assets assets_temp
cp src/image/opencast-white.svg assets_temp
```

This will save the relevant asset folder somewhere that the `git checkout` won't touch.  We know that we saved the assets into a safe place earlier, now we just tell `git` to pay attention to those files.  If the files are the same this ends up doing nothing, but if they're *not* then we end up with whatever is current in `develop`

```suggestion
        run: |
          rm -rf assets
          mv assets_temp assets
          git add assets
          git diff --staged --quiet || git commit --amend -m "Build $(date)"
```

.github/workflows/deploy-test.yml
```diff
@@ -111,12 +111,16 @@ jobs:
       - name: generate index.html
         working-directory: admin-interface-test
         run: |
-          echo '<html><body><ul>' > index.html
+          echo '<html><head><link rel=stylesheet type=text/css href=.github/assets/index.css /></head><body><div class="head-container"><img src=https://admin-interface.opencast.org/develop/assets/opencast-white-B3KJ3o-a.svg /></div><div class="navbar-container"></div><div class="text-container"><p>Test deployment for the latest development versions of the Opencast admininterface.The branches listed here correspond to Opencast's own branches.</br><b>Please select a version.</b></p></div><ul>' > index.html
```
gregorydlogan: All of the changes to this file can be dropped since they won't be used.

gregorydlogan (changes requested):
Hope this clears things up.  I can file a PR against your PR if you're time constrained, but I figured it would be more helpful to teach rather than show.

.github/workflows/deploy-test.yml
ltekbas: Thank you for the help and explanation on how to handle this. 
I changed the files like you suggested.

PR #1417, opened 2025-08-28
Revert 1416

gregorydlogan:
This PR reverts #1416, which breaks the UI when uploading an event with a series.  This was initially missed because I was looking to verify that the series ACL is the default (which does work!), and the UI isn't otherwise obviously broken.  The more subtle issue is if you try and *change* the ACL, which does *not* work and in fact throws the UI into some kind of infinite loop.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/NewAccessPage.tsx
@@ -29,7 +29,7 @@ import ModalContentTable from "../../../shared/modals/ModalContentTable";
  * This component renders the access page for new events and series in the wizards.
  */
 interface RequiredFormProps {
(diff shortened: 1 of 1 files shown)


gregorydlogan:
@KatrinIhler the underlying loop appears to originate with https://github.com/opencast/opencast-admin-interface/pull/1417/files#diff-cafaf44cbf912f0492fb3e36ac8b7c20385fe3946288cad3d0bd63451a60a4cbR81, which ends up in `src/slices/seriesDetailsSlice.ts`'s `fetchSeriesDetailsAcls` which (I suspect) ends up somehow updating the state, which re-triggers the the call to `useEffect` in `NewAccessPage`, etc.  It doesn't bust the stack - thought the debug console will eventually just close on its own - but it does loop infinitely as demonstrated by browser CPU usage being pinned while the ACL page is open

PR #1419, opened 2025-08-29
Fix series rights for new events - properly this time

KatrinIhler:
Reverts #1417 and fixes #1416.


jduehring:
Tested against https://legacy.opencast.org 
Looks good to me. 
As you mentioned series ACLs are being loaded twice for whatever reason, but do not reload infinitely anymore. 
The ACLs also seem to correctly apply upon upload.

jduehring (approved):
Approved as mentioned above.

PR #1423, opened 2025-09-02
Fix series filters being overwritten by existing event filters

### LINKED ISSUE #1246: Events and Series: Overwrite existing event filters when filtering for series
Steps to reproduce:

1. Open Events table and add any filter other than series.
2. Switch to Series table and click on a series name to filter for that series in the Events table.
3. In the Events table, the previous filter does not get deleted (Status: Finished):

Image

It would be more intuitive to delete existing filters on the Events table when a user filters for a series _out of the series table_.

EnsiyehE:
Hello all,
this PR Try to fix issue stated in https://github.com/opencast/opencast-admin-interface/issues/1246 as follows :
Clear existing filters on Events table when applying a series filter from Series table.

This fixes #1246


Arnei (approved):
Looks reasonable and fixes the issue :+1:

Issue #1405, opened 2025-08-20
Access policy (role): Highlight the correct role in the drop-down


oas777:
When opening the drop-down for "role", it's always the first role in that list highlighted in blue - not the role currently selected. That's not good.

https://github.com/user-attachments/assets/c88779dc-f758-43c1-af6e-ab84234d9b74

EnsiyehE:
Hi Olaf,

I looked into the issue where the dropdown for “role” always highlights the first role in the list instead of the currently selected one. At first, I thought this was a UI problem, but after checking the code, I found that we filter the dropdown options with this function:

filterRoles(roles, policies)....


This means we remove any roles that are already used in the ACL policies to avoid duplicates. Because of that, the currently selected role isn’t in the dropdown list at all, so the component highlights the first available role instead.

This seems to be the reason why the behavior looks wrong. I can change it so that the currently selected role is added back into the dropdown options when editing, so the user can see what they already have selected.

@Arnei ,  does this make sense as the intended behavior, or should we include the current role in the dropdown when editing?

oas777:
Thanks for looking into this, @EnsiyehE - I didn't realise the role wasn't actually in the list. If we (have to) keep it like that, I would suggest to remove the highligh - but let's hear what @Arnei has to say first.

mtneug:
IMO what is selected by default is a preference and not a hard rule. However, I think it is strange that the current value can no longer be found in the list. The filtering makes sense in general, but I would also say that adding the current value after filtering (or skipping it during the filter process) makes sense.

oas777:
Maybe not a rule, but there must be some strong (UX/UI) recommendations for this: If the selected value is part of the list, that value should be highlighted: 

<img width="725" height="654" alt="Image" src="https://github.com/user-attachments/assets/0b2a7fc7-cde0-4e00-b034-4a9bb0dafee5" />

Arnei:
Having the current role also show in the dropdown and having it highlighted in some fashion seems very reasonable to me. We are already doing this for e.g. the language selector in the event metadata, so we probably want similar behaviour here.

I have no hard opinion on whether the current role should also be initially focused in the dropdown, or if focusing the first element in the dropdown is better. Would need to look into the UX and accessibility concerns.


As an aside, we probably still want to keep other set roles beside the current role out of the dropdown, I think? So if I my UI looks like this

[...]

PR #1427, opened 2025-09-05
fix the missing current  chosen value inside the Access- policy -role drop down

EnsiyehE:
Hello all, 

I have achieved to fix the missing current chosen value access policy role to the drop down options , but due to fix size and some other issues i have not been able to change the focus to the current selected value , so this PR only solves part of the the issue mentioned in #1405

please review and let me know about my issues

thank you for your time
best regards
Ensiyeh


gregorydlogan:
Either this isn't working for me, or I'm completely missing what changed.

EnsiyehE:
Hello Greg,

To explain briefly: if you go to the Series tab, then open Series Details and the Access Policy tab, each dropdown previously did not include the currently chosen value from the input field among its options. I have now fixed that part. However, in the original issue, it was also requested that the dropdown should focus on that selected value, which I believe by default always focuses on the first option and i could only reach to bring back the previously chosen value inside the options

Thank you for the feedback

bests
Ensiyeh

src/components/shared/DropDown.tsx
```diff
@@ -210,6 +211,15 @@ const DropDown = <T, >({
 
 		// @ts-expect-error: React-Select typing does not account for the typing of option it itself requires
 		components: { MenuList },
+		filterOption: createFilter({ ignoreAccents: false }), // To improve performance on filtering
```
Arnei, line 214: I'm curious, did you test if this actually performance on filtering and if so, by how much?

src/components/shared/DropDown.tsx
```diff
@@ -210,6 +211,15 @@ const DropDown = <T, >({
...
+
+		isOptionSelected: (option, _selectValue) => {
+		  const typedOption = option as DropDownOption;
+		  if (isAclDropDown) {
+		    return false;
+		  }
+		  return typedOption.value === value;
+		},
```
Arnei, line 222: So from what I understand, what this does is that if the Dropdown is defined as an AclDropdown, there is never an option selected. I don't understand why we wouldwant that, what is this supposed to achieve?

Arnei (changes requested):
Selected roles now appear in the dropdown of their select field, as such this PR achieves what it intends to do.

This breaks the "Users" and "Roles" split in the ACL tab that is active when user role sanitization is disabled. It should not break that.

Arnei:

PR #1428, opened 2025-09-08
Fix Tobira paste bug

owi92:
Fixes an issue where pasting a value in the Tobira path segment input would not register that value when submitting the form.

For reviewing: Either turn off white space changes or ignore the first commit, which fixes the indentation of a large code block.


wsmirnow (approved):
Works for me.

I'm not go through the indentation patch. I would expect same code reformatted.

owi92:
@Arnei since we talked about this yesterday in the dev meeting, I checked and can confirm that other input fields work fine with copy/paste.

wsmirnow:
This PR should probably go to r/18.x?

PR #1431, opened 2025-09-08
Css color consistency

Arnei:
Includes #1400. #1400 Should be reviewed and merged first.

This PR replaces the "hardcoded" color values that are flying around everywhere in our scss files with the color variables defined in `_variables.scss`. Color values that are not represented by our variables are either added to the variables file, or replaced with a variable with a similar value.

Goals:
- Make colors easier to work with. Having our colors all grouped in one place should make it easier to e.g. implement a dark mode.
- Reduce the number of different colors. We really should not need 50 different shades of grey (number approximated).
- Increase contrast. I used the opportunity to slightly change some colors to make the ui more readable.

### How to test this

This touches every element that has color, so unfortunately pretty much everything. Quickly click through everything ui element you can find and check if any colors appear "off".


src/styles/mixins/_button.scss
ferishili, line 1: I noticed there's a whole block of code commented out. Is that intentional?
If so, could you please add a short comment above it explaining why it's being kept? Also, it would be better to use a proper block comment instead of multiple single-line comments for readability.

ferishili (changes requested):
Thanks @Arnei,
Your changes look fine, however a few things need your attention!

First and most important one is:
While running the UI click-through tests, I noticed the contrast in the calendar (month and year selection) looks off, as shown below. Could you please take a look at that as well?

Image

src/styles/mixins/_button.scss
Arnei: Thanks for the review! I added a comment on why I thought this might be worth keeping, even if commented out. Will also take a look at the calendar.

Arnei:
> While running the UI click-through tests, I noticed the contrast in the calendar (month and year selection) looks off, as shown below. Could you please take a look at that as well?

Contrast should be better now!

Bildschirmfoto vom 2026-03-05 10-53-05

ferishili (approved):
Everything look fine, thanks @Arnei

Issue #336, opened 2024-04-29
Registration Nag Screen Missing


gregorydlogan:
The behaviour added by https://github.com/opencast/opencast/pull/4321 is missing in the new UI still.

Edit: I meant the modal which pops up, rather than the bell, but I can't find the relevant PR which turned that on.  Regardless, the new admin UI should pop up the modal if the user hasn't registered, and hasn't told the modal to go away.

EnsiyehE:
Hi,
I’m working on the new admin UI and noticed that the behavior introduced by [opencast/opencast#4321](https://github.com/opencast/opencast/pull/4321)
 seems to be missing. Specifically, I understand there should be a modal that pops up if the user hasn’t registered and hasn’t dismissed it. Since I’m new and haven’t seen the previous implementation, could you please share a screenshot (or any visual reference) from the original feature so I can ensure it’s implemented correctly in the new UI?

Also, just to confirm—this modal pops up whenever the user is not an admin or not related, correct?

Thank you!

@gregorydlogan  @Arnei

gregorydlogan:
Right, so the modal should appear, provided:
- The user is an admin user (ROLE_ADMIN, as enforced by the security config)
- `window.location.hostname` is not `localhost` (don't annoy developers)
- The system is not registered yet (check in the `/admin-ng/adopter` namespace for this)
- The terms of use that's been agreed to matches the latest terms of use available (`/admin-ng/adopter/latestToU`)
- It's been 30 or more days since the modal was last dismissed AND the system is unregistered

The previous code can be found at:
- https://github.com/opencast/opencast/blob/d20e7d659eb7894edb804e6877e445b38c04de5e/modules/admin-ui-frontend/app/scripts/shared/controllers/applicationController.js#L127
- https://github.com/opencast/opencast/blob/r/15.x/modules/admin-ui-frontend/app/scripts/shared/resources/adopterRegistrationResource.js
- https://github.com/opencast/opencast/blob/r/15.x/modules/admin-ui-frontend/app/scripts/shared/resources/termsOfUseResource.js

Ideally we'd have a way to have this pop up for org admins as well since not all adopters use global admins.

The modal actually still exists, and works.  It's under the ? button in the top right of the admin ui, under Adopter Registration.  It just doesn't automatically pop up like it used to in the angular ui.

EnsiyehE:
Hello Greg Logan,

Thank you for your hint and reply.

[...]

PR #1434, opened 2025-09-11
Feature/registration nag screen 336

EnsiyehE:
Hello 

please review the code , i have been trying to fix the issue which is stated at #336

thank you for your time 
bests 
Ensiyeh


src/components/Header.tsx
```diff
@@ -132,6 +133,22 @@ const Header = () => {
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, []);
 
+	useEffect(() => {
+  			if (!user) { return; }
+
+  			const isAdmin = user.isAdmin || user.isOrgAdmin;
+			// un comment the below line after the test
+  			// const notLocalhost = window.location.hostname !== "localhost";
+			// for testing purposes on the local host
```
gregorydlogan:
```suggestion
  			const notLocalhost = window.location.hostname !== "localhost";
```

I would also suggest flipping the logic here and making it this:
```suggestion
  			const isLocalhost = window.location.hostname == "localhost";
```
such that the if stanza on :148 reads `if (isAdmin && dismissedLongEnough && !isLocalhost) {`, which (IMO) reads cleaner

gregorydlogan (changes requested):

src/components/Header.tsx
EnsiyehE: Hi,
Done 🙂

PR #1436, opened 2025-09-11
Remove unnecessary html tags from table cells

Arnei:
Removes html tags that are not needed from table cell components. This won't really make any sort of noticable dent performance wise, but it still results in a cleaner dom, so why not.

### How to test this

Check that the tables look exactly like before.


KatrinIhler:
FYI: In the future I would prefer if we put code refactoring changes like these into the stable or develop branch, because it doesn't improve the user experience and could potentially break something, which we want to avoid for the legacy release.

PR #1437, opened 2025-09-12
Avoid unnecessary table rerenders

Arnei:
This patch prevents certain actions from causing all table rows to rerender, if the actions did not actually affect the table rows. It does this by refactoring the table component so that redux state updates only trigger rerenders in the
components that actually use the update data.

Actions include:
- Selecting a single row.
- Opening the page size dropdown.
- Displaying a tooltip by hovering over a hoverable table element.

This should not affect inital table load times.

### How to test this

This patch changes our logic on how we access row information a bit, so make sure that all table functionality still works as expected.

### DIFF
--- a/src/components/events/Events.tsx
+++ b/src/components/events/Events.tsx
@@ -55,7 +55,6 @@ const Events = () => {
 	const editMetadataEventsModalRef = useRef<ModalHandle>(null);
 
 	const user = useAppSelector(state => getUserInformation(state));
-	const showActions = useAppSelector(state => isShowActions(state));
 	const events = useAppSelector(state => getTotalEvents(state));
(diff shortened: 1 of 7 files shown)


gregorydlogan (changes requested):
With this patch, when I select any rows on the event table and then wait a couple of seconds, the selections un-set themselves.  I suspect this is happening on the normal table re-render when fetch updated data from the backend.  In my case, the system is otherwise idle, so row order is not changing.

Arnei:
Thanks for the review. I now also noticed that this breaks the adding/removing columns feature, so this may take some time to fix.

Arnei:
Ok nvm, this was more straightforward to fix than I thought. Both issues should be fixed now.

PR #1440, opened 2025-09-16
Fix series ACLs not being applied to new events

KatrinIhler:
#1006 renamed the value and this part was missed.

How to test: Create a series with at least one role in the ACL, start to create an event and select that series, check if the roles from the series show up in the ACL tab during event creation.

Sponsored by the University of Jena :white_flower:


jduehring:
Tested against r/18.x and develop and works as expected. Thanks! (Fixes https://github.com/opencast/opencast-admin-interface/issues/1421)

PR #1443, opened 2025-09-19
Improvement: Update the submitter display in the event details tab

miesgre:
This PR adds the submitter name in the events details tab.
Captura de pantalla 2025-09-19 a las 11 48 00

I've made this PR to r/17.x. If it's not correct, please let me know and I'll change it.


lkiesow:
This is how the view looks right now:
Screenshot from 2025-09-22 13-17-08

With this patch applied, this looks weird if a user has only a username but no full name:
Screenshot from 2025-09-22 13-17-17

Can we then fall back to showing the username only without the braches?

Also, I think it's a problem that the column becomes too long. I don't have a partivular long name and I have collegues with far longer names (and email addresses), but even for me the column is already quite wide:

Screenshot from 2025-09-22 13-23-39

Can we remove the email here? It's available if you go into the workflow details, if you need it.  Or can we at least make it a tooltip instead?

miesgre:
I have modified the PR to display the submitter name in a tooltip.

Captura de pantalla 2025-09-22 a las 14 04 11

Issue #984, opened 2024-11-19
Add next/previous tab hotkeys


lkiesow:
The old admin ui featured two hotkeys that would let you move forward and backward between the different steps in a create modal. It could be nice to have this again. We had an attempt at doing this in #830, but that faced some issues.

EnsiyehE:
Hello Lars & Arne,

I have tried to find a solution for this issue , 

but during that i have found another issue underhood , which i thought to state it here and ask your opinion if it would be better to open a seprate issue or work on the same branch , 

when a user submit a title and wants to press Enter ( to temporary take the auto focus from the first input field in the modal addEvent and edit an event ) , or by Pressing Escape key on the keyboard , sometimes we lose the focus trap of the modal .. which i think it is not a good Ux To see for anyone  ( The focus would be gone to the browser elements so to say ) 

could you please try this and look and my PR and share your opinions regarding both ?

the PR Is #1441 

thanks for your time
bests
Ensiyeh

PR #1447, opened 2025-09-29
Feature Add next/previous tab hotkeys

EnsiyehE:
Hello all,

as stated in the problem #984 , there was no possibility like the old version of admin user interface of open cast to navigate between different steps like ( edit event modal ) (eg, metadata to publications ) ,

by this PR i tried to suggest the solution which i would appreciate your view and opinions , if i am in the currect track or even close to it that i can continue my solution to other wizard modals inside the Admin UI ,

now it is possible to move forward whenever the focus is not on the input type fields by pressing: alt + Enter and back ward : alt + Backspace , to test this PR please first open the edit window modal , and change the focus by tabbing from the input field to the next one and then use the hotkeys as above i have mentioned them

thank you all for your Time
Ensiyeh


src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -2086,6 +2086,8 @@
 				"SERIES_VIEW": "Series",
 				"NEW_EVENT": "Add event",
 				"NEW_SERIES": "Add series",
+				"NEXT_STEP": "Go to next step",
```
gregorydlogan: I'm not sure that 'next/previous step' are the right words here.  To my ears, 'next/previous pane/panel' are what I would expect here, but I'm not sure what we've called those elsewhere in the UI.

gregorydlogan (dismissed):
Works, nitpicking about the shortcut descriptions

src/i18n/org/opencastproject/adminui/languages/lang-en_US.json
EnsiyehE: Hi ,

thank you for the hint 

I have changed it , in the files you mentioned to the next and previous panel

gregorydlogan: It's not about the key name, although changing those makes sense too.  When you look at the keyboard shortcut summary, it says 'Go to next step', which should probably read 'Go to next pane'.

EnsiyehE: Done 🙂

src/components/events/partials/modals/EventDetails.tsx
```diff
@@ -229,6 +231,46 @@ const EventDetails = ({
 		        dispatch(openModalTab(tabNr, "workflow-details", "entry"));
 		}
 	};
+	const wizardTabs = tabs.filter(tab =>
+	  !tab.hidden && tab.page !== EventDetailsPage.Tobira && tab.page !== EventDetailsPage.Statistics,
```
Arnei: Why are we explicitly avoiding the Tobira and Statistics tab?

src/components/events/partials/modals/EventDetails.tsx
```diff
@@ -229,6 +231,46 @@ const EventDetails = ({
...
+	  const currentIndex = wizardTabs.findIndex(tab => tab.page === page);
+	  if (currentIndex <= 0) { return; }
+	  openTab(wizardTabs[currentIndex - 1].page);
+	};
+	// NEXT STEP
+	useHotkeys(
+	  availableHotkeys.general.NEXT_PANEL.sequence,
+	  event => {
+	    const target = event.target as HTMLElement;
+	    if (!["INPUT", "TEXTAREA"].includes(target.tagName)) {
```
Arnei: I believe we don't need to explicitly check for "INPUT" or "TEXTAREA" here, the `useHotkeys` hook should do that per default.

src/components/shared/modals/Modal.tsx
```diff
@@ -69,7 +69,7 @@ export const Modal = forwardRef<ModalHandle, PropsWithChildren<ModalProps>>(
 
 	return ReactDOM.createPortal(
 		isOpen &&
-			<FocusTrap>
+			<FocusTrap focusTrapOptions={{ escapeDeactivates: false }}>
```
Arnei: This seems unrelated to the issue at hand?

src/components/shared/modals/ModalContentTable.tsx
```diff
@@ -15,14 +16,24 @@ const ModalContentTable = ({
...
+  	const lastFocusedRef = useRef<HTMLElement | null>(null);
+    const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    const active = document.activeElement as HTMLElement | null;
+      if (event.key === "Escape") {
+        if (active && event.currentTarget.contains(active)) {
+          lastFocusedRef.current = active; // store the last focused element
+          active.blur();
+        }
+      }
+    };
```
Arnei: This also seems unrelated to the issue at hand?

Arnei (changes requested):
Seems to work fine. The linked issue requested the next/previous hotkeys for the create modals though, not for the details modals. Not that having them in the details modal would be bad, but they're more useful in the create modals where users will need to move between adjacent tabs more often.

src/components/shared/modals/Modal.tsx
EnsiyehE: Hello,

As I mentioned earlier in the issue: when a user submits a title and presses Enter (to temporarily remove the autofocus from the first input field in the Add Event or Edit Event modal), or presses Escape, we sometimes lose the focus trap of the modal.

This means the focus shifts to browser elements instead of staying within the modal — which I believe is not a good UX experience.

I’ve now fixed this issue with my latest code update. When you press Escape for the first time, the modal retains its focus trap as intended. On the second press, it behaves as expected and closes the modal.

Previously, as I mentioned in the issue, pressing Escape once would immediately break the focus trap — which I think was a poor UX experience.

src/components/shared/modals/ModalContentTable.tsx
EnsiyehE: I have mentioned the reason on the above discussion already 

bests 
Ensiyeh

src/components/events/partials/modals/EventDetails.tsx
EnsiyehE: Hello 

As I see it on my machine (please correct me if I’m wrong), these elements are not part of the Event Details tab, so I couldn’t find a reason for them to be included in the logic. This was just an experiment for this specific modal.

My goal was actually to generalize the behavior, and I mentioned this in the issue to you and Lars, who opened it, but I unfortunately didn’t receive any feedback.

I believe that trying to generalize this inside the parent element could lead to bigger issues, because the Modal component handles different types of modals — such as confirmations and others — which have different types and behaviors.

eventdetails tab names

src/components/events/partials/modals/EventDetails.tsx
EnsiyehE: The reason is :::  the hotkey doesn’t trigger while the first input is autofocused is due to how react-hotkeys-hook handles keyboard events by default. Keyboard events are sent to the element that currently has focus. By default, this library ignores events coming from input, textarea, or other editable elements to avoid interfering with typing.

You can see this explained in the docs here :::  https://react-hotkeys-hook.vercel.app/docs/api/use-hotkeys
example

As you previously mentioned in the underlying issue, the same behavior occurs, and this could have been a suggestion for a reasonable solution for the issue:


issue explainations

src/components/events/partials/modals/EventDetails.tsx
Arnei: > As I see it on my machine (please correct me if I’m wrong), these elements are not part of the Event Details tab, so I couldn’t find a reason for them to be included in the logic. This was just an experiment for this specific modal.

These elements are part of the event details, but do not show per default. Much like the extended metadata tab for example.They would appear if their `hidden` condition evaluates to `false`.

But if this was just an experiment and is bound to change in the future, no need to sweat over it.

src/components/events/partials/modals/EventDetails.tsx
Arnei: Oh so we want the hotkeys to be able to trigger while an input or textarea element is in focus. Makes sense to me. Then you have to invert your if condition I think. Or make use of `enableOnFormTags`, that is probably cleaner.

src/components/shared/modals/Modal.tsx
Arnei: That may all be true, but still has nothing to do with adding next/previous hotkeys. Please submit this as a separate PR. Having different features as separate PRs has several benefits:
- It is easier for reviewers to tell which code change belongs to which feature, speeding up the review process.
- Having separate features as separate commits leads to a cleaner commit history.
- If the first feature is approved but the second feature is controversial, the first feature is not blocked by the second feature.

src/components/events/partials/modals/EventDetails.tsx
EnsiyehE: I think there was a misunderstanding on my side. The goal of this condition was to allow the hotkeys to trigger even when an input or textarea is focused. We used this if (!["INPUT", "TEXTAREA"].includes(target.tagName)) check as a practical workaround because, due to how the hotkey listener is attached globally, event.target is observed as the body element even when the input is focused. This makes the hotkeys fire correctly in that context. Essentially, the condition ensures that the hotkeys work as intended inside input or textarea fields, despite the event target not being the input itself.


! makes it fire inside input fields — because target.tagName is "BODY" even when an input is focused.

src/components/events/partials/modals/EventDetails.tsx
EnsiyehE: > These elements are part of the event details, but do not show per default. Much like the extended metadata tab for example.They would appear if their `hidden` condition evaluates to `true`.
> 
> But if this was just an experiment and is bound to change in the future, no need to sweat over it.

Thanks for the explanation! I’ve made the tabs that are not hidden at the moment as the wizardTabs 😊

src/components/events/partials/modals/EventDetails.tsx
Arnei: > ! makes it fire inside input fields

But it does not. If I open the event details, focus an input field and then press "Alt + Enter", the input field simply looses focus like if I had just pressed "Enter".

EnsiyehE: > > ! makes it fire inside input fields
> 
> But it does not. If I open the event details, focus an input field and then press "Alt + Enter", the input field simply looses focus like if I had just pressed "Enter".

I’m not entirely sure what you mean. As far as I can see, my branch is up to date and everything is working correctly. I’m attaching a video in this comment to show that we can move forward and backward while still keeping the focus on the field. Isn’t that the main point of our discussion — that when the focus is on a field, we should still be able to navigate between the tabs?

[Aufzeichnung 2025-11-17 092834.zip](https://github.com/user-attachments/files/23577615/Aufzeichnung.2025-11-17.092834.zip)

Arnei: Does not work for me. First "Alt+Enter" unfocuses the field, second "Alt+Enter" moves forward:

[Bildschirmaufzeichnung vom 2025-11-17 09-49-56.webm](https://github.com/user-attachments/assets/50e9b84b-32ad-45d9-8eac-2681014af70b)

EnsiyehE: > Does not work for me. First "Alt+Enter" unfocuses the field, second "Alt+Enter" moves forward:
> 
>  Bildschirmaufzeichnung.vom.2025-11-17.09-49-56.webm
I’m not entirely sure what you mean. If I flip the condition, the logic stops working completely because, as we discussed, the event target is not always the actual input element. Due to event delegation and bubbling, the event often comes from a parent element rather than the input itself.

Because of this, the input element is not treated as the event target, so the reversed condition never passes. As a result, the handler never runs and we can’t move to the next or previous step. This matches what we talked about earlier and what the documentation explains: with the default behavior, the focus is placed on the first input, but the event still bubbles from a parent element.

Given that this is the natural browser behavior, I’m not sure what outcome you’re expecting when the condition is reversed. Could you clarify what behavior you want to achieve exactly?

Arnei: > Because of this, the input element is not treated as the event target, so the reversed condition never passes. As a result, the handler never runs and we can’t move to the next or previous step. This matches what we talked about earlier and what the documentation explains: with the default behavior, the focus is placed on the first input, but the event still bubbles from a parent element.

Tbh I have no idea what you are trying to tell me here.

> Given that this is the natural browser behavior, I’m not sure what outcome you’re expecting when the condition is reversed. Could you clarify what behavior you want to achieve exactly?

I am currently assuming that we want to be able to move forward by hotkey even while an input or textarea is focused, which is not the case.

EnsiyehE: Let me try to explain it more clearly:

I’m not saying anything complicated — only that the browser does not always report the focused <input> or <textarea> as the event target. Even when the input is focused, the key event often comes from a parent element because of event bubbling. That means the event target is usually a wrapper element or even the body, not the input itself.

This is why, when the condition is reversed, it never matches and the handler never runs. That’s the only point I’m making.

Right now, I’m not fully sure what behavior you expect.
Are you expecting that the user should be able to trigger the hotkey while an input or textarea is focused? If yes, that is not possible with the browser’s default behavior, since the event is not coming from the input as we assume.

So could you please clarify what the expected behavior is from your perspective?

According to the documentation of react-hotkeys-hook  as i mentioned in here https://react-hotkeys-hook.vercel.app/docs/api/use-hotkeys 

the key event does not fire on the same element if it is an input or textarea. This is the default behavior of the library and the browser.

In other words, if the focus is on an input or textarea, we cannot trigger the hotkey — this is not something we can change. That’s why reversing the condition breaks the functionality: the event simply never fires when the input is focused. 

so there is no solution for what you are saying unfortunately ...

EnsiyehE: Imagine ::: 

If we removed the condition entirely, the initial focus on the first input field (which is always set by default) would prevent us from moving forward or backward between tabs or steps. Without this feature, the hotkeys simply wouldn’t work when the input is focused.

As I mentioned before, the first time you press Esc, it would remove the focus trap from the window completely( could you try it please on your develop branch now on the same window ?? eventdetails ) . I have already solved that issue, but it should be tracked separately in another ticket.( you mentioned in one of the discussions ) 

This condition is necessary to allow navigation( by keyboard ))  while respecting the default behavior of focus on input fields.

Arnei: > Are you expecting that the user should be able to trigger the hotkey while an input or textarea is focused?

That is the behaviour I think is desirable. I am not saying that your code does this (it does not), but I think that is the behaviour we would ultimately want.

> In other words, if the focus is on an input or textarea, we cannot trigger the hotkey — this is not something we can change. 

I think we can. This should do the trick:
```javascript
	useHotkeys(
		availableHotkeys.general.NEXT_PANEL.sequence,
		event => {
				goNextStep();
		},
		{ enableOnFormTags: ["INPUT", "TEXTAREA"] },
		[goNextStep],
	);
```

Arnei: > If we removed the condition entirely, the initial focus on the first input field (which is always set by default) would prevent us from moving forward or backward between tabs or steps. Without this feature, the hotkeys simply wouldn’t work when the input is focused.

Previously you claimed that we cannot make hotkeys work while the input element is focused. Now you seem to claim that due to "the condition", the hotkeys work while the input element is focused. That sounds like a contradiction to me, so I probably still don't understand what you are trying to explain. What exactly do you expect "the condition" to do?

EnsiyehE: with all due respect : 

I prefer to use your code for now to avoid further discussions, while I focus on other issues. I will revisit this in the coming days . If it doesn’t work, I will try to explain my intent again. As I’ve already mentioned in the explanation of the issue:

Currently, it is possible to move forward whenever the focus **is not** on an input field by pressing Alt + Enter (forward) and Alt + Backspace (backward). To test this PR: first open the edit window modal, then tab from one input field to the next, and use the hotkeys as described above.

When I said we can solve it “whenever we are focused on the input,” I meant exactly this scenario.

If we removed the condition entirely, the initial focus on the first input field (which is set by default) would prevent us from moving forward or backward between tabs or steps. Without this condition, the hotkeys wouldn’t work when the input is focused.

As I mentioned before, pressing Esc the first time currently removes the focus trap from the window completely. Could you please try this on your develop branch using the same window (e.g., EventDetails)? I have already solved that issue, but it should be tracked separately in another ticket, as you noted in one of our discussions.

This condition is necessary to allow keyboard navigation while still respecting the default behavior of focusing on input fields.

Arnei: > I prefer to use your code for now to avoid further discussions, while I focus on other issues

Understandable, good luck with your endeavours.

> Currently, it is possible to move forward whenever the focus is not on an input field by pressing Alt + Enter (forward) and Alt + Backspace (backward).

I can confirm that.

> To test this PR: first open the edit window modal, then tab from one input field to the next, and use the hotkeys as described above.

If I for example open the event details modal, then tab from title to subject, then press the "Next" hotkey, it will not take me to the next tab. Only after pressing the "Next" hotkey a second time will it take me to the next tab. Is that what you expect to happen?

> If we removed the condition entirely, the initial focus on the first input field (which is set by default) would prevent us from moving forward or backward between tabs or steps. Without this condition, the hotkeys wouldn’t work when the input is focused.

From my testing, removing the condition (we are talking about `if (!["INPUT", "TEXTAREA"].includes(target.tagName))`, right?) does not have any effect, the behaviour stays the same.

> As I mentioned before, pressing Esc the first time currently removes the focus trap from the window completely. Could you please try this on your develop branch using the same window (e.g., EventDetails)?

With your PR, the focus trap stays intact as far as I can tell. Pressing ESC while inside an input field and then pressing tab will put the focus on the close-modal button. And I cannot tab outside the modal.

EnsiyehE: > > I prefer to use your code for now to avoid further discussions, while I focus on other issues
> 
> Understandable, good luck with your endeavours.
> 
> > Currently, it is possible to move forward whenever the focus is not on an input field by pressing Alt + Enter (forward) and Alt + Backspace (backward).
> 
> I can confirm that.
> 
> > To test this PR: first open the edit window modal, then tab from one input field to the next, and use the hotkeys as described above.
> 
> If I for example open the event details modal, then tab from title to subject, then press the "Next" hotkey, it will not take me to the next tab. Only after pressing the "Next" hotkey a second time will it take me to the next tab. Is that what you expect to happen?
> 
> > If we removed the condition entirely, the initial focus on the first input field (which is set by default) would prevent us from moving forward or backward between tabs or steps. Without this condition, the hotkeys wouldn’t work when the input is focused.
> 
> From my testing, removing the condition (we are talking about `if (!["INPUT", "TEXTAREA"].includes(target.tagName))`, right?) does not have any effect, the behaviour stays the same.

as i showed in my recording , from the screen , first it would remove the focus from current field and then goes to the next tab as you mentioned by the current condition related to event.target and how and where you handle this event as i mentioned ..

so yea , the behavior is the same as you mentioned , first the focus one second would go back from current field and then next tab .. 


> 
> > As I mentioned before, pressing Esc the first time currently removes the focus trap from the window completely. Could you please try this on your develop branch using the same window (e.g., EventDetails)?
> 
> With your PR, the focus trap stays intact as far as I can tell. Pressing ESC while inside an input field and then pressing tab will put the focus on the close-modal button. And I cannot tab outside the modal.

So my question is: if the focus trap remains active after pressing Esc, do you still think I should split the code, considering these two accessibility issues are so closely related ?????

EnsiyehE:
@Arnei 
Hi again!
I tested the code you shared, and it works just as you said. When the input is focused, I’m able to switch to the next tab using the keyboard.
Thanks a lot for your time and the helpful tip!

src/components/events/partials/modals/EventDetails.tsx
Arnei: > So my question is: if the focus trap remains active after pressing Esc, do you still think I should split the code, considering these two accessibility issues are so closely related ?????

I still believe these two accessibility issues are unrelated.

EnsiyehE: Ok

Fixed ,

I removed it

Arnei (approved):
Works and looks good to me.

As EnsiyehE mentioned in the inital post, this does not completely resolve https://github.com/opencast/opencast-admin-interface/issues/984, but works as a starting point for further development.

PR #1476, opened 2025-11-11
Fix workflow parameters in scheduled events

marwyg:
If the workflow parameters for a scheduled event are not set, the default values didn't get used. The values also couldn't be changed.

With this change, the default values from the workflow parameters should be set for the workflow parameters in the workflow tab of the scheduled event. Here a screenshot of the tab:

image

To test this, you can schedule an event and change the workflow properties in the database and also set different workflow parameters in the workflow xml:

You could change the database with this sql for example:
```mysql
update oc_scheduled_extended_event 
set workflow_properties = '{"not_existing_property1":"true","foobar":"false"}' 
where mediapackage_id = "af041d5d-1b75-4a6d-9f57-45321a80c4b4";
```
You could also set an empty list for testing.

In the workflow xml you can edit the configuration_panel_json and set different default values. If a value is not present in the database, the default value from the workflow xml should be used.


Arnei:
`workflowPanelUtils.ts` has similar code for getting default workflow parameters. Could we reuse that here and if not why not?

Arnei:
@marwyg poke

marwyg:
I will work on this in the near future

PR #1477, opened 2025-11-12
Go to first page when changing filters via Stats

ciegler:
Currently the Admin UI does not reset the offset and the page number when changing filters via the Stats bar. 

With this the offset and the page number is set to 0 when a filter is changed via the Stats bar, without reloading all events as `goToPage(0)`  would do.

### DIFF
--- a/src/components/shared/Stats.tsx
+++ b/src/components/shared/Stats.tsx
@@ -10,6 +10,7 @@ import {
 	removeTextFilter,
 } from "../../slices/tableFilterSlice";
 import { loadEventsIntoTable } from "../../thunks/tableThunks";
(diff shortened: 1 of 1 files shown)


gregorydlogan:
This should probably be targeted earlier.  In discussion at the dev meeting, this should go to 17.x.

ciegler:
I rebased my branch to r/17.x and changed the PR to target the branch r/17.x.
This was my fist time changing the PR target branch, so I hope that  these were the right steps to do.

marwyg:
I encountered this problem while testing something different. Found your PR and tested it. Seems to work :+1:

PR #1479, opened 2025-11-19
Fix language and licence selection in EventDetails tab

jduehring:
Fixes https://github.com/opencast/opencast-admin-interface/issues/1471 . 

I think in https://github.com/opencast/opencast-admin-interface/pull/1375 the options field for non series fields has been missed and wasn't parsed at all. This resulted in no language or licence options showing up in the dropdown. 

This patch reintroduces the parsing of the options for those fields.

### How to test this patch

With the patch installed, open the event details / new event tab and check if you can see and select language, licence and series options in the dropdown selection.

### DIFF
--- a/src/components/shared/wizard/RenderField.tsx
+++ b/src/components/shared/wizard/RenderField.tsx
@@ -230,14 +230,14 @@ const EditableSingleSelect = (props: EditableSingleSelectProps) => {
 		text={text}
 		form={form}
 		options={
-		metadataField.collection
-			? metadataField.collection.map(item => ({
-				label: item.label ?? item.name,
-				value: item.value,
-				order: item.order,
-			}))
-			: []
-		}
+			metadataField.collection
+				? metadataField.collection.map(item => ({
+					label: item.label ?? item.name,
+					value: item.value,
+					order: item.order,
+				}))
+				: []
+			}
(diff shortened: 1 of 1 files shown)


src/components/shared/wizard/RenderField.tsx
```diff
@@ -229,6 +229,15 @@ const EditableSingleSelect = (props: EditableSingleSelectProps) => {
...
 		form={form}
+		options={
+			metadataField.collection
+				? metadataField.collection.map(item => ({
+					label: item.label ?? item.name,
+					value: item.value,
+					order: item.order,
+				}))
+				: []
+			}
```
owi92, line 240: you can also do
```suggestion
		options={metadataField.collection?.map(option => ({
			...option,
			label: option.label ?? option.name,
		}))}
```

but it's functionally equivalent, so not super necessary to change.

owi92 (approved):
This works and can be merged now.

There is still an issue where the dropdown for the language field shows the languages in the currently selected UI language but the filtering is happening based on the english word.
I.e. if the Admin UI is set to german, the dropdown shows the "german" option as "Deutsch", but typing "de" won't produce any suggestion. You still need to type "ge" to get the "Deutsch" option. But I'd suggest that should be fixed in another PR.

PR #1496, opened 2026-01-05
Fix multi-select values in metadata not removable (Safari)

Arnei:
Fixes a bug where values could not be removed from multi-select fields, like presenters or contributors in the event metadata. When clicking on the "X" of the blue box with the value, the field would leave edit mode but the value was not removed. This is due to a timing issue most notable in Safari. This patch fixes the issue by making the field not leave edit mode when clicking on "X", which is how multi-select fields behaved in 18.x as well.

This fix has an unpleasant side effect, where multi-select fields will not properly blur anymore when left via tab navigation. I am not sure how to properly fix this with the current code.

### DIFF
--- a/src/components/shared/wizard/RenderMultiField.tsx
+++ b/src/components/shared/wizard/RenderMultiField.tsx
@@ -206,27 +206,26 @@ const ShowValue = ({
 	showCheck,
 	onBlur,
 }: {
-    setEditMode: (e: boolean) => void
+	setEditMode: (e: boolean) => void
 	form: FieldProps["form"]
 	field: FieldProps["field"]
 	showCheck: boolean,
 	onBlur: () => void
 }) => {
 	return (
-	<div
-		tabIndex={0}
-		onClick={() => setEditMode(true)}
-		onFocus={() => setEditMode(true)}  // <-- activate edit mode on focus
-		onKeyDown={e => {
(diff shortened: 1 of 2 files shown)


gregorydlogan:
This doesn't break anything obvious to me, but I don't have a Mac to test with.  @mtneug @karendolan @rute-santos @snoesberger I think you're all on macs.  There's a test above, don't even have to build anything.

snoesberger:
I did a quick test on my Mac using Safari. I can now successfully remove values from the multi-select fields. I also successfully tested this with Firefox and Chrome.

But I had to use the docker image to test this. [test.admin-interface.opencast.org/1496/2026-01-05_11-52-27/  ](https://test.admin-interface.opencast.org/1496/2026-01-05_11-52-27/) doesn't allow me to change metadata because of CORS policy:

`Access to XMLHttpRequest at 'https://develop.opencast.org/admin-ng/event/ID-portrait/metadata' from origin 'https://test.admin-interface.opencast.org' has been blocked by CORS policy: Method PUT is not allowed by Access-Control-Allow-Methods in preflight response`

It would be nice if the CORS policy could be adapted to allow such tests.

gregorydlogan:
Thanks @snoesberger.  I've filed an issue to track the CORS stuff!

PR #1501, opened 2026-01-15
Download correct crowdin translation branch

gregorydlogan:
This PR causes GHA to download and apply the correct translations from crowdin.  Previously *all* branches applied the translations from `main`, which is both defunct and no longer updating with new keys.


KatrinIhler:
I will not cut an Admin UI release just for this, can you repoint this to 18?

PR #1509, opened 2026-02-19
Unify date formatting, base it on locale

### LINKED ISSUE #1279: On THE Event details dialog  'Start date' and 'Created' date don't have the same format
Start date format is currently DD/MM/YYYY, Time
Created date format is currently MM/DD/YY, Time (with am/pm) 

Ideally, they should both have the same format and consistent to how dates are displayed anywhere else.

Image

Arnei:
This patch aims for a consistent date display across the entire ui. That it is not always consistent can be seen for example in the event details metadata, where the start date is formatted differently from the created date.

Furthermore, this patch aims to base date formatting more on the chosen locale to create a more intuitive experience for users. No more A.M./P.M. in the german locale for example.

To achieve this, date-fns is chosen as the gold standard and momentjs is removed from the project completely. Not only because moment is legacy, but also because our datepickers were using date-fns already and will not use anything else.

Additionally, this removes our custom date formatting from the translation files and replaces it with date-fns tokens. In my opinion, we do not need to try and be smarter than our datetime library when it comes to formatting dates. Translators retain the ability to enforce their favourite formatting should they so chose,
but will have to be aware that it will not be respect everywhere, e.g. not in the datepickers.

This patch contains changes to all translation files. I am aware that everything except en_US will get overwritten, the changes to other languages are merely for ease of reviewing. If this patch is merged, the translations should be adapted asap as per this example.

Before:
Bildschirmfoto vom 2026-02-19 13-50-34

After:
Bildschirmfoto vom 2026-02-19 13-50-42

Fixes #1279

### How to test this

Can be tested as is. Check anything with dates. Special attention should be payed to the statistics tab and scheduling. Ideally we get people who use other locales apart from the english ones to test this, so they can mention if anything looks very odd in their favourite locale now.


marcusm7:
Screenshot_20260219_184654

There is an inconsistency when English UK is selected.

Arnei:
Thanks for testing! Interesting, I'm not seeing that for UK on my end.

Bildschirmfoto vom 2026-02-20 08-55-52

Could you try clearing the cache for the website, just in case?

marcusm7:
> Thanks for testing! Interesting, I'm not seeing that for UK on my end.

> Could you try clearing the cache for the website, just in case?

Indeed, interesting.

- Firefox. Disabled all extensions. English UK in OC. Cache cleared. Reload with ctrl+f5 -> AM/PM in popup
- Version 145.0.7632.45 (Official Build) -> OK 

# Details Firefox:
``` about:support
Name: Firefox
Version: 147.0.4
Build ID: 20260212191108
Distribution ID: canonical
User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:147.0) Gecko/20100101 Firefox/147.0
OS: Linux 6.14.0-123037-tuxedo #37~24.04.1tux2 SMP PREEMPT_DYNAMIC Fri Jan 23 22:31:21 UTC 2026
OS Theme: Breeze / Adwaita
```

ypatios:
we noticed this when selecting UK English and still getting the (questionable :P) US date format. thx for fixing

Arnei:
As far as I can tell, the issue marcusm7 discovered is a problem with react-datepicker. When setting the datepicker to use an input field for the time selection in the dialog, the input field does not respect the set formatting and locale. Instead the browers formats the `<input type="time">` field in whatever way it sees fit. This explains why for marcusm7 there is always a 12h format, while for me it is always a 24h format, independent of the locale we chose.

This seems to be a known issue in the react-datepicker project, but there seem to be no plans to fix it. react-datepicker allows us to use our own, custom time input component, so we could implement that. However, properly implementing a time input component is non-trivial (and a reason to use a library). So I do not want to implement such a custom component in this PR, or at all really. But even if with this issue present, I think this PR constitutes an improvement over the current state of things and would still like to see it merged.

PR #1517, opened 2026-03-05
Fix scheduling due to missing device id

### LINKED ISSUE #1515: Scheduling Broken
Via https://github.com/opencast/opencast/issues/7420, the admin UI is no longer sending which CA should be scheduled against in the scheduling request, which obviously breaks the backend since it can't tell which agent to schedule against.

This affects 19.x and develop right now.

Arnei:
Fixes #1515.

The function that creates the scheduling request for the backend was looking for the device id in the wrong place.

### How to test
Check if scheduling an event works again with this PR. You can create a dummy capture agent if none is available with the /capture-admin/agents/{name} endpoint.

### DIFF
--- a/src/slices/eventSlice.ts
+++ b/src/slices/eventSlice.ts
@@ -451,6 +451,7 @@ export const postNewEvent = (params: {
 		scheduleStartHour: string,
 		scheduleStartMinute: string,
 		sourceMode: string,
+		location: string,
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Normally I would wait for a technical meeting review pass to get this in, but it's directly affecting test infra and likely adopters.

Issue #1506, opened 2026-02-06
DropDown List Jerkiness/Jumps


ypatios:
Actual Behaviour:
Dropdown menu repopulates, scroll position resets to top, causing jerkiness/jumps.

Expected Behaviour:
Open menu retains scroll position during background refreshes.

Steps to Reproduce (eg for Workflow DropDown):
1. Open event upload or bulk action dialog.
2. Open workflow dropdown (effect is stronger with long lists).
3. Scroll down in the open dropdown.
4. Wait for Admin UI table polling refresh (default 5'')

https://github.com/user-attachments/assets/be3134d2-f23e-4720-817b-5b9937cf84de

—
Possible Cause:
[DropDown](https://github.com/opencast/opencast-admin-interface/blob/develop/src/components/shared/DropDown.tsx) component uses `AsyncSelect` for static options; `defaultOptions` and `loadOptions` create new references on every parent render, triggering menu reload.

Suggested Fix:
Use synchronous `Select` for local options (`!fetchOptions`), memoize formatted options with `useMemo`, stabilize `loadOptions` with `useCallback`.

marcusm7:
Hi,
- I can't reproduce it in 17.8. Select Workflow Dropdown opened -> last entry selected. Everything as expected
- On https://stable.opencast.org (OC19) I can observe the described misbehaviour (with Firefox 147.0.3 in wayland).

ypatios:
after reading [this](https://github.com/opencast/opencast-admin-interface/issues/1507)  –which might even be related–, I made a little demo screen rec and added it to the description above.

mwuttke:
I think I can reproduce this behaviour in our test system. We are using OC 19.3.

ypatios:
> I think I can reproduce this behaviour in our test system. We are using OC 19.3.

last time I tried it was reproducible on both stable.opencast.org and legacy.opencast.org (ie 19.x and 18.x).
I believe the bug was introduced with 18.4-ish.

mwuttke:
Is there a chance to fix this issue in OC 19.4?

ypatios:
> Is there a chance to fix this issue in OC 19.4?

afaik not yet, unfortunately. i could give it another try but my chances of success are low, i'm afraid

Issue #1442, opened 2025-09-19
Are dropdown the best way to let users select from thousands of options?


Arnei:
In our app we have several dropdowns that are likely to contain thousands of options on big production systems. For example:
- The series and contributor dropdowns in the metadata tab in the event modal.
- The role dropdowns in the access policy tab in event/series modal.
- The series dropdown in the events table filter.

The good thing about these dropdowns is that they are searchable, so users can filter through the options by typing.

The bad thing about these dropdowns is that they required us to load all options, often causing the app to lag (see also #1425). When I think on how we can best solve this issue on a design level, I always end up with paginated, searchable lists, e.g. the events table. I reckon there is a better way to solve this problem. Some way to let users who know what they are looking for *and* users who don't quite know yet what they are looking for find what they need without them noticing any lags. Suggestions are welcome.

ltekbas:
First, I want to note that the dropdown styling for the `Contributer(s)` and `Presenter(s)` is inconsistent with the other examples provided in this issue.
If the dropdown component is retained, its styling must be unified across all instances. While I understand the potential reasoning behind having the dropdown body open to the side of the input field, maintaining  styling consistency  is a huge factor for user experience and overall usability.
In the presentation, I propose an alternative styling option that better integrates tags alongside selection from the dropdown. 

The following is a [short presentation](https://www.figma.com/deck/5UMw3aFNhWqImxYA75GPda/Opencast-Dropdown?node-id=1-25&viewport=-6642%2C-192%2C1&t=PizISC2moVt2m0Eu-1&scaling=min-zoom&content-scaling=fixed&page-id=0%3A1) and overview of the design options I've developed. While there may not be a single 'perfect' solution, these suggestions offer a more robust approach to handling large datasets

PR #1523, opened 2026-03-13
Fix role getting unselected when changing write

### LINKED ISSUE #1507: New Policy field in Series Details acts weird
If a New policy is selected in the Access Policies, and a checkbox for Read or Write is checked afterwards, the selected policy vanishes.

This was working properly on 16.x. I saw the problem in 17.11 and 19.x on stable.opencast.org

Image
Image

Arnei:
Fixes #1507.

There is a bug in the access policy tab of event details that can be reproduced like so:
1. Add a new policy.
2. Set a role for the new policy.
3. Change the write rights for the new policy.

The role is now unset. This patch should fix that.

### DIFF
--- a/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
+++ b/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
@@ -547,11 +547,11 @@ export const AccessPolicyTable = <T extends AccessPolicyTabFormikProps>({
 																		handleChange={element => {
 																			if (element) {
 																				const matchingRole = roles.find(role => role.name === element.value);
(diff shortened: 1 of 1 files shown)


ypatios:
thanks for this :D

ypatios:
> This pull request is deployed at [test.admin-interface.opencast.org/1523/2026-03-13_10-01-29/ ](https://test.admin-interface.opencast.org/1523/2026-03-13_10-01-29/). It might take a few minutes for it to become available.

a quick test here confirms fix 👍 
at the same time I have to note the sluggishness of that particular tab and the grave effects of #1506; see also #1442 

https://github.com/user-attachments/assets/944a861f-04b3-4a25-8147-89fdb7cf437d

PR #1542, opened 2026-03-31
Less queries

geichelberger:
Filter-driven reloads are now centralized to prevent multiple data
fetches at once. Further, added a feature to cancel ongoing requests and
ensure that older responses don’t replace newer data.

After applying filters, there’s a temporary pause on auto-refresh to
prevent interruptions while users are still filtering.

Removed duplicate fetch behavior in search/filter apply logic by
avoiding page-navigation fetch plus manual reload in the same
interaction.


src/components/shared/TableFilters.tsx
```diff
@@ -201,7 +205,9 @@ const TableFilters = ({
 				setFilterSelector(false);
 				setSelectedFilter("");
 				// Reload of resource after going to very first page.
-				dispatch(goToPage(0));
+				dispatch(deselectAll());
+				dispatch(setOffset(0));
+				dispatch(setPageActive(1));
```
Arnei, line 210: Should this not set to `0` instead of `1`?

src/components/shared/TablePage.tsx
```diff
@@ -40,37 +40,89 @@ const TablePage = ({
...
+		if (source === "auto" && currentLoadSource.current === "filters") {
+			return;
+		}
+
+		const requestId = ++latestLoadRequestId.current;
+		currentLoadSource.current = source;
+
+		currentLoadRequest.current?.abort?.();
+
+		const fetchRequest = dispatch(fetchResource()) as Promise<PayloadAction<any, string>> & { abort:() => void };
```
Arnei, line 85: I think we can do without the type assertions here to make this a bit easier to read?

Arnei (changes requested):
Reads sound to me and works from my testing. I would like to see a brief comment in the code that explains the intent behind these functions (basically what you already wrote in the commit message).

geichelberger:
@Arnei, is there a way to resolve the conflicts without reverting those changes? https://github.com/opencast/admin-interface/pull/1559/changes/0eff28d01704e1c839a933872bf9bf059ccdfc73
We need the synchronization in the layer above. Or should that be done at the redux level?

Arnei:
I made the changes before looking at this PR. It could arguably be fine to revert them? 

Not sure how this could be done at the redux level. You'd still have to have state somehow, but could you still keep that in refs? Would you want to use redux state? I don't know.

gregorydlogan:
We're staring down the barrel of having 19 move to the legacy release slot.  Are we wanting to get this into 19, or leave it for 20+?

src/components/shared/TableFilters.tsx
geichelberger: Changed it to 0

geichelberger:
Changed the PR target to r/20.x

Arnei:
By this time I feel like this should be done at the redux level. Having to manage this manually with six different refs feels like there should be a better solution, maybe redux toolkit actually offers this functionality already in some shape or form? Also TablePage.tsx feels like the wrong spot for this since this arguably only concerns tables (my fault, I know). Maybe tablethunks.tsx should be involved in this somehow?

geichelberger:
This is far beyond my knowledge of frontend tinkering :)

PR #1561, opened 2026-04-02
Filter available roles in dropdown by config

Arnei:
Adds a feature that reduces the number of available options in the dropdowns in the access policy tab. It does this bx matching prefixes that are defined in the
config in the backend.
This can for example be used to filter out all roles with "ROLE_GROUP" prefix, to make the dropdowns more usable.

*Will not break anything without the backend PR: https://github.com/opencast/opencast/pull/7541*

### How to test this

Requires you to be able to make config changes in the backend, else you can only test that everything works as usual.

### DIFF
--- a/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
+++ b/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
@@ -467,8 +467,8 @@ export const AccessPolicyTable = <T extends AccessPolicyTabFormikProps>({
 
 	// Filter available options by custom prefixes from the config
 	if (aclDefaults) {
-		if (aclDefaults["display_role_filter_prefixes"]) {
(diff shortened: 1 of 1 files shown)


src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
```diff
@@ -465,6 +465,16 @@ export const AccessPolicyTable = <T extends AccessPolicyTabFormikProps>({
 		return newRole;
 	};
 
+	// Filter available options by custom prefixes from the config
+	if (aclDefaults) {
+		if (aclDefaults["display_role_filter_prefixes"]) {
```
mtneug: `filter` is ambiguous. Is it "filter out" or "filter for". Maybe change the config name to `display_role_filter_blacklist_prefixes`?

gregorydlogan (approved):

PR #1566, opened 2026-04-07
Preselect all capture agent inputs when scheduling new event

Arnei:
Originally https://github.com/opencast/admin-interface/pull/979. I had trouble finding the relevant commits and rebasing, so I figured it might be simpler to just create a new commit and PR.
Be sure to check https://github.com/opencast/admin-interface/pull/979 for further info and discussion.
Also this will again need some testing.


snoesberger:
I successfully tested this PR with an Extron SMP and pyCA. The Extron SMP inputs were preselected and the recording was fine. For pyCA ,without any inputs, the recording was also fine.

PR #1568, opened 2026-04-07
Use react-select for page size dropdown

### LINKED ISSUE #1215: Pagination Items per Page cannot be operated via keyboard
The element can be accessed via a tab, but no selection can be made. Neither enter and space nor cursor keys work.

Arnei:
*Originally #1530*

Fixes #1215.

We are currently using our own custom styled dropdown thingy for the page size selector. This patch instead uses "Select" from react-select for the page size selector. The goal is to get the accessibility features of using a proper select component "for free", instead of having to painstakingly code our own. Apart from that, the page size selector should work as before.

### How to test this

Can be tested as is. The page size selector is the element at the bottom right of the page.

### DIFF
--- a/src/utils/componentStyles.ts
+++ b/src/utils/componentStyles.ts
@@ -163,6 +163,10 @@ export const pageSizeStyles: StylesConfig<any, false> = {
 		height: 28,
 		width: 75,
 		borderRadius: 4,
+		"&:hover": {
+			borderColor: "#378dd4",
+			boxShadow: "0 0 0 1px #378dd4",
+		},
 		paddingLeft: 10,
 
 		cursor: "pointer",
--- a/src/utils/componentStyles.ts
+++ b/src/utils/componentStyles.ts
@@ -147,3 +147,68 @@ export const dropDownSpacingTheme = (theme: Theme) => ({
 		baseUnit: 2,
 	},
 });
+
+/**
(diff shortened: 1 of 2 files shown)


geichelberger:
Should we use a different border color? It doesn't fit the other colors. I think the border color should be #378dd4

grafik

Arnei:
The border color of what specific part of the select are you referring to? They look fine to me.

geichelberger:
This one:

grafik

src/utils/componentStyles.ts
```diff
@@ -147,3 +147,69 @@ export const dropDownSpacingTheme = (theme: Theme) => ({
...
+		position: "absolute",
+		right: "20px",
+	}),
+	control: (base, state) => ({
+		...base,
+		minHeight: 28,
+		height: 28,
+		width: 75,
+		borderRadius: 4,
+		borderColor: state.isFocused ? "#378dd4" : "hsl(0, 0%, 80%)",
```
geichelberger:
```
		"&:hover": {
				borderColor: state.isFocused ? "#378dd4" : "hsl(0, 0%, 80%)",
		},
```

Shouldn't it be like that? 
But there is a box shadow with the wrong color, which is quite strange.

gregorydlogan (approved):

PR #1569, opened 2026-04-07
Switch to react-tooltips

### LINKED ISSUE #1446: Replace MUI Tooltips with something more performant
According to https://github.com/opencast/opencast-admin-interface/pull/1433, mui tooltips take too much time to render when used in large quantities, i.e. in a large table. It would still be nice to have tooltips in our tables though. So the goal of this issue is to get tooltips in our codebase that render quickly enough to not significantly delay the rendering of a 1000 line table. This will likely mean finding a different library than mui tooltips altogether.

Arnei:
*Originally #1485*

Fixes #1446.

Replaces our mui tooltip wrapper with react-tooltip. Also adds back tooltips that were commented out.

The reason for this change is that tooltips from react-tooltip are much more performant
when used in larger quantities. It will still add about ~100ms to the render time of events or series
tables with a 100 elements, which is not great but hardly noticable with our load times of over a second.

### How to test this

Does not require any special set up for testing. Check that tooltips appear where you expect them and behave as you would expect them to.


geichelberger:
I don't know what it is, but I find the tooltips annoying. Maybe it's because they appear too fast or because they don't disappear if you move to the next element?

gregorydlogan (approved):

Arnei:
Given that 1 out of 2 reviewers is unhappy with these tooltips, maybe we want to revert this?

geichelberger:
I wouldn't revert it; let's wait for user feedback.

PR #1571, opened 2026-04-08
LifeCycle Policies - Squashed

Arnei:
*Originally #1004.*

*The old PR for lifecycle policies was old, had many commits and many merges. Creating a squashed commit seemed like the simplest option to bring lifecycle policies from the old history to the new history.*

Instead of only being able to view lifecycle policies, this commit lets you edit them, create new ones and even delete them. 

Depends on changes to the backend: https://github.com/opencast/opencast/pull/6139. These changes are currently pointed at develop.

To test this, you'll need an Opencast with backend changes mentioned above.

Demonstration:
[Bildschirmaufzeichnung vom 2024-12-04 10-45-28.webm](https://github.com/user-attachments/assets/ab48f8cb-6d4c-47aa-b3b5-3ee466676c60)

### DIFF
--- a/src/components/events/partials/wizards/NewLifeCyclePolicyWizard.tsx
+++ b/src/components/events/partials/wizards/NewLifeCyclePolicyWizard.tsx
@@ -1,0 +1,181 @@
+import { useEffect, useState } from "react";
+import { Formik } from "formik";
+import NewAccessPage from "../ModalTabsAndPages/NewAccessPage";
+import WizardStepper from "../../../shared/wizard/WizardStepper";
+import { useAppDispatch, useAppSelector } from "../../../../store";
+import { getUserInformation } from "../../../../selectors/userInfoSelectors";
+import { UserInfoState } from "../../../../slices/userInfoSlice";
+import { postNewLifeCyclePolicy } from "../../../../slices/lifeCycleSlice";
+import NewLifeCyclePolicyGeneralPage from "../ModalTabsAndPages/NewLifeCyclePolicyGeneralPage";
+import NewLifeCyclePolicySummary from "./NewLifeCyclePolicySummary";
+import { LifeCyclePolicySchema } from "../../../../utils/validate";
+import { initialFormValuesNewLifeCyclePolicy } from "../../../../configs/modalConfig";
+import { parseTargetFiltersForSubmit } from "../../../../utils/lifeCycleUtils";
+import { ParseKeys } from "i18next";
+
+/**
+ * This component manages the pages of the new event wizard and the submission of values
+ */
+const NewLifeCyclePolicyWizard = ({
+	close,
+}: {
+	close: () => void
+}) => {
+	const dispatch = useAppDispatch();
+
+	const user = useAppSelector(state => getUserInformation(state));
+
+	const initialValues = getInitialValues(user);
+
+	const [page, setPage] = useState(0);
+	const [snapshot, setSnapshot] = useState(initialValues);
+	const [pageCompleted, setPageCompleted] = useState<{ [key: number]: boolean }>({});
+
+	// Caption of steps used by Stepper
+	const steps: {
+		translation: ParseKeys,
+		name: string,
+		hidden: boolean,
+	}[] = [
+		{
+			translation: "LIFECYCLE.POLICIES.NEW.GENERAL.CAPTION",
+			name: "general",
+			hidden: false,
+		},
+		{
+			translation: "EVENTS.EVENTS.NEW.ACCESS.CAPTION",
+			name: "access",
+			hidden: false,
+		},
+		{
+			translation: "EVENTS.EVENTS.NEW.SUMMARY.CAPTION",
+			name: "summary",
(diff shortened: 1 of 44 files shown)


ferishili:
There are currently two major errors that prevent me from continuing with deeper testing:


---
1. The `sort` parameter is `undefined`:
   `GET | http://localhost:3000/api/lifecyclemanagement/policies?limit=10&offset=0&sort=undefined:undefined`

Screenshot 2026-04-20 at 13 28 34

---
2. In the **Create** modal, as soon as I navigate to the **Access Policy** step/tab, an error appears:

> In HTML, `<td>` cannot be a child of `<td>`.

Screenshot 2026-04-20 at 13 31 09

---
That would nice to have a rebase here as well!

Thanks @Arnei

Arnei:
For the first error, try clearing cookies and website data for the page.
For the second error, that currently appears in every Access Policy tab (not just lifecycle policies), even outside of this PR. It should not prevent you from testing.

src/components/events/partials/wizards/NewLifeCyclePolicyWizard.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+	const handleSubmit = (values: typeof initialValues) => {
+		const fixedValues = {
+			...values,
+			targetFilters: parseTargetFiltersForSubmit(values.targetFiltersTransformed),
+			accessControlEntries: values.policies,
+		};
+
+		const response = dispatch(postNewLifeCyclePolicy(fixedValues));
+		console.info(response);
```
ferishili, line 94: Is this info log just for development, or should everyone be able to see it?

src/slices/lifeCycleDetailsSlice.ts
```diff
@@ -0,0 +1,278 @@
...
+	policies: { acl: { ace: Ace[] } }
+}, { dispatch }) => {
+	const { id, policies } = params;
+
+	const data = new URLSearchParams();
+	data.append("accessControlEntries", JSON.stringify(policies.acl.ace));
+
+	await axios.put(`/api/lifecyclemanagement/policies/${id}`, data)
+		.then(response => {
+			console.info(response);
```
ferishili: Here as well?

src/slices/lifeCycleDetailsSlice.ts
```diff
@@ -0,0 +1,278 @@
...
+		} else if (stringified === Object(stringified)) {
+			stringified = JSON.stringify(stringified);
+		}
+		// @ts-expect-error: ???
+		data.append(key, stringified);
+	});
+
+	await axios.put(`/api/lifecyclemanagement/policies/${policy.id}`, data)
+		.then(response => {
+			console.info(response);
```
ferishili: And here?

src/slices/lifeCycleSlice.ts
```diff
@@ -0,0 +1,186 @@
...
+		// @ts-expect-error: ???
+		data.append(key, stringified);
+	});
+
+	data.delete("accessControlEntries");
+	data.append("accessControlEntries", JSON.stringify(prepareAccessPolicyRulesForPost(policy.accessControlEntries).acl.ace));
+
+	await axios.post("/api/lifecyclemanagement/policies", data)
+		.then(res => {
+			console.info(res);
```
ferishili: And here too?

src/slices/lifeCycleSlice.ts
```diff
@@ -0,0 +1,186 @@
...
+			console.error(res);
+			dispatch(addNotification({ type: "error", key: "LIFECYCLE_POLICY_NOT_SAVED" }));
+		});
+});
+
+export const deleteLifeCyclePolicy = createAppAsyncThunk("lifeCycle/fetchLifeCyclePolicies", async (id: string, { dispatch }) => {
+	await axios
+		.delete(`/api/lifecyclemanagement/policies/${id}`)
+		.then(res => {
+			console.info(res);
```
ferishili: What about here?

ferishili (changes requested):
While checking the browser dev console, I noticed several info logs that only show the response object from a request, for example:

`Object { data: "", status: 200, statusText: "OK", headers: {...}, config: {...}, request: XMLHttpRequest }`

Because of that, I commented out the `console.info` statements introduced in this PR so we can decide whether they should remain or be removed before merging.

Another issue I ran into: when editing an existing record and clicking **Save**, the dialog does not close, although the notification is displayed. Also, when specifically changing the **"Active"** parameter, the change is not applied.

Additionally, when clicking **Cancel**, I would expect the dialog to close.

Video reference:
[https://github.com/user-attachments/assets/745d75eb-afeb-4124-9adf-d26527ce7e16](https://github.com/user-attachments/assets/745d75eb-afeb-4124-9adf-d26527ce7e16)

src/slices/lifeCycleSlice.ts
Arnei: Okay okay, I should remove my debug logging statements, I get it :P

PR #1575, opened 2026-04-10
Update vite to new major version 8

Arnei:
Supercedes #1554.

Updates vite to major version 8 and handles the necessary changes.

Kicks out vite-plugin-svgr because we are not using it.


ferishili (approved):
Thanks @Arnei,

I can confirm that the changes not only show `found 0 vulnerabilities` during installation 🤩
they also reduced the build time from `✓ built in 6.33s` to `✓ built in 2.04s` 😍

PR #1576, opened 2026-04-13
Fix roles getting unselected in ACL tab

Arnei:
In the access policy tab, when adding new roles, old roles would be removed. Basically, changing the roles in some manner would result in wonky behaviour.

This patch fixes that, mostly by reverting 9d0a7b2203ee09e836bce57d08b230db87b2e32c and making sure the policy key id is more reliable than index.

### How to test

Can be tested as is, with any event. Make sure that #1523 is still fixed as well.

### DIFF
--- a/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
+++ b/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
@@ -530,7 +530,7 @@ export const AccessPolicyTable = <T extends AccessPolicyTabFormikProps>({
 											{formik.values.policies.length > 0 &&
 												policiesFiltered.map(
 													(policy, index) => (
-														<tr key={index}>
+														<tr key={policy.role + index}>
 															{/* dropdown for policy.role */}
 															<td className="editable">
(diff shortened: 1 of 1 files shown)


ferishili (changes requested):
Thanks @Arnei,

I can confirm that the changes work as expected. However, the **Access Policy** tab currently has a noticeable delay. Click-Actions are applied only after about 1-2 seconds at best, and I also encountered a white page after adding more than three roles with several action (checkbox) changes.

My suggestion would be to move the inline functions into static methods, although I'm not sure whether that would fully resolve the issue.

Arnei:
The Access Policy tab seems to have the issues you noticed also on `r/19.x` I think. Could you confirm?

ferishili:
> Could you confirm?

Yes, I can confirm that as well, but it seems to me as it works just a little bit smoother!

Arnei:
> However, the Access Policy tab currently has a noticeable delay

@ferishili This PR should help somewhat: https://github.com/opencast/admin-interface/pull/1589

ferishili (approved):

PR #1577, opened 2026-04-13
Show again that there are open comments

### LINKED ISSUE #1544: Visual Feedback about Open Comments has Disappeared
As an OC-Admin, I want to have visual feedback about open comments / Todos in the admin ui.

---

Pre OC 19.x there used to be visual feedback about the presence of open commends of events (TODOs). This feature has disappeared in 19.x.

Here's how it used to look like (test done on legacy.opencast.org -> 18.x):

Image

...
In 19.x, a new set of icons was introduced (see below; taken from stable.opencast.org -> 19x).
Now, the messages icon remains the same for events with open and for those with closed comments:

Image

BTW: notice the red dot on the scissors icon that indicates an open "needs cutting" comment. This works as it used to.

Arnei:
Fixes #1544.

In the events table, a "comment" icon is displayed for events with comments. Before we switched to lucide icons, there were two different comment icons, depending on if the event still had unresolved comments or not. This behaviour was lost.

This patch brings the behaviour back. If there are still unresolved comments, the speech bubble icon now has an exclamation point in it.

Why not a filled icon like before? Because lucide does not have that.

### How to test this

Can be tested as is. Add comments to an event in the event details, then watch the icons on the right in the events table.

### DIFF
--- a/src/components/events/partials/EventActionCell.tsx
+++ b/src/components/events/partials/EventActionCell.tsx
@@ -17,7 +17,7 @@ import { openModal } from "../../../slices/eventDetailsSlice";
 import { ActionCellDelete } from "../../shared/ActionCellDelete";
 import { Modal, ModalHandle } from "../../shared/modals/Modal";
 import ButtonLikeAnchor from "../../shared/ButtonLikeAnchor";
-import { LuFileSymlink, LuFileText, LuFolderOpen, LuLink, LuMessageCircle, LuScissors, LuTriangleAlert } from "react-icons/lu";
+import { LuFileSymlink, LuFileText, LuFolderOpen, LuLink, LuMessageCircle, LuMessageCircleWarning, LuScissors, LuTriangleAlert } from "react-icons/lu";
 import { SeriesDetailsPage } from "./modals/SeriesDetails";
 
 /**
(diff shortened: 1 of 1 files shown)


ferishili (approved):
Thanks @Arnei,

works fine and can be merged right away!

Just a small (if possible) request:
Would it be possible to add something like a small red dot indicator, similar to the scissor icon?

Arnei:
>  Would it be possible to add something like a small red dot indicator, similar to the scissor icon?

I would first like to see how users like this change. If it is not good enough (because it might not be visible enough or something), we can discuss how to do it better in a follow up PR, and then maybe use the red dot indicator.

ypatios:
thanks for fixing this :)

ypatios:
> This pull request is deployed at [test.admin-interface.opencast.org/1577/2026-04-13_14-58-26/ ](https://test.admin-interface.opencast.org/1577/2026-04-13_14-58-26/). It might take a few minutes for it to become available.

any ideas why I get a 404 here?

ypatios:
> Just a small (if possible) request: Would it be possible to add something like a small red dot indicator, similar to the scissor icon?

that sounds good and would provide a unified visual xp. +1
as @Arnei said, let's see what the community thinks before any changes :)

PR #1579, opened 2026-04-14
Add UI for playlist management

### LINKED ISSUE #1399: Add UI for managing playlists
Playlist is now a third major type in the current Opencast data model (besides event and series). It is already used in Tobira and Stud.IP and maybe other LMS will follow. As such, the Opencast Admin UI should also be able to manage playlists.

CC https://github.com/opencast/opencast/issues/6976

### LINKED ISSUE #6909: Integration of Playlists into Admin UI
Focus: Currently, playlist management is handled separately from the core Admin UI. 

Integrating playlist creation, editing, and management functionality directly into the Admin UI could provide a unified interface for content and playlist administration. 

A clear, streamlined approach to content management will improve admin efficiency and reduce errors.

owi92:
Originally #1537 

This adds the necessary interface elements that are needed to use playlists in the admin interface.
Structure and UI design follows what is already present for events and series, and reuses components whenever possible.

Needs https://github.com/opencast/opencast/pull/7496 to work properly.

Closes https://github.com/opencast/admin-interface/issues/1399
Closes https://github.com/opencast/opencast/issues/6909

### DIFF
--- a/src/components/events/partials/EventsNavigation.tsx
+++ b/src/components/events/partials/EventsNavigation.tsx
@@ -4,18 +4,23 @@ import { ParseKeys } from "i18next";
  * Utility file for the navigation bar
  */
 export const eventsLinks: {
-	path: string,
-	accessRole: string,
-	text: ParseKeys
+  path: string,
+  accessRole: string,
+  text: ParseKeys
 }[] = [
-	{
-		path: "/events/events",
-		accessRole: "ROLE_UI_EVENTS_VIEW",
-		text: "EVENTS.EVENTS.NAVIGATION.EVENTS",
-	},
-	{
-		path: "/events/series",
-		accessRole: "ROLE_UI_SERIES_VIEW",
-		text: "EVENTS.EVENTS.NAVIGATION.SERIES",
-	},
+  {
+    path: "/events/events",
+    accessRole: "ROLE_UI_EVENTS_VIEW",
+    text: "EVENTS.EVENTS.NAVIGATION.EVENTS",
+  },
+  {
+    path: "/events/series",
+    accessRole: "ROLE_UI_SERIES_VIEW",
+    text: "EVENTS.EVENTS.NAVIGATION.SERIES",
+  },
+  {
+    path: "/events/playlists",
+    accessRole: "ROLE_UI_PLAYLISTS_VIEW",
+    text: "EVENTS.PLAYLISTS.TABLE.CAPTION",
+  },
 ];
--- a/src/components/events/partials/wizards/NewPlaylistWizard.tsx
+++ b/src/components/events/partials/wizards/NewPlaylistWizard.tsx
@@ -1,0 +1,191 @@
+import { useState, useEffect } from "react";
+import { Formik } from "formik";
+
+import NewMetadataCommonPage from "../ModalTabsAndPages/NewMetadataCommonPage";
+import NewAccessPage from "../ModalTabsAndPages/NewAccessPage";
+import NewPlaylistSummary from "./NewPlaylistSummary";
+import WizardStepper, { WizardStep } from "../../../shared/wizard/WizardStepper";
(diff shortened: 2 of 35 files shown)


src/components/events/partials/PlaylistTitleCell.tsx
```diff
@@ -0,0 +1,8 @@
+import { Playlist } from "../../../slices/playlistSlice";
+
+/**
+ * This component renders the title cells of playlists in the table view
+ */
+const PlaylistTitleCell = ({ row }: { row: Playlist }) => <span>{row.title}</span>;
```
Arnei: If you are just rendering plain text, you don't need to create an extra cell for that. You can do the same thing you did with `description` in `playlistTableConfig.ts`

src/components/events/partials/EventsNavigation.tsx
```diff
@@ -4,18 +4,23 @@ import { ParseKeys } from "i18next";
...
+    text: "EVENTS.EVENTS.NAVIGATION.EVENTS",
+  },
+  {
+    path: "/events/series",
+    accessRole: "ROLE_UI_SERIES_VIEW",
+    text: "EVENTS.EVENTS.NAVIGATION.SERIES",
+  },
+  {
+    path: "/events/playlists",
+    accessRole: "ROLE_UI_EVENTS_VIEW", // TODO
```
Arnei: Found a TODO. Probably want a unique role here?

src/components/events/partials/wizards/NewPlaylistWizard.tsx
```diff
@@ -0,0 +1,207 @@
...
+import { removeNotificationWizardForm } from "../../../../slices/notificationSlice";
+import { getUserInformation } from "../../../../selectors/userInfoSelectors";
+import { getAclDefaultActions, getAclDefaultTemplate } from "../../../../selectors/aclSelectors";
+import { AclTemplate } from "../../../../slices/aclSlice";
+import { UserInfoState } from "../../../../slices/userInfoSlice";
+
+
+/**
+ * Manages tabs of the new playlist wizard and submission of form values.
+ * Currently supports metadata and access policy steps.
```
Arnei: Mildly outdated comment, as this also supports the "entries" step. I'd suggest just removing this line, we don't need to keep track of this in a comment.

Arnei (changes requested):
Works and makes sense to me. I do like the drag and drop for playlist entries. Below are nitpicks.

I would like to see the PlaylistDetails Icon be blue instead of green. Green is for series.

Arnei (changes requested):
Actually, pagination might be broken? If I have more then 10 playlists, I still only get one page with 10 items.

owi92:
Thanks for the review. Pagination should be fixed now.

Edit: Sorry, ~~idk why the `compare` button isn't doing its thing.~~ I have a rebase onto develop in there as well. I have addressed the points you mentioned in the review, I think most of them were in the first commit.

Edit 2: Also forgot about the green vs. blue icon question. Matthias also mentioned that in another issue I believe. But if green is for series, you could also argue that blue is for events, right?
I decided for green in the first place because I feel that playlists have more in common with series than with events, with both being collections of videos. Should playlists get a new color? (I don't think so, but if they do, which one)?
I guess I am putting too much thought into that, while not having a strong opinion. So let's just make it blue.

Arnei:
Pagination works now.

The playlist titles in the playlists table do not render. I swear this should work without an extra TitleCell, just like for description >.>

You could argue that blue is for events, or that it is just for everything else (see users, groups). Series is probably a non-blue color, because there was a need for a series details button in the events table, and they wanted it to be visually more distinct from its neighbouring event details button. Is that good design? I dunno *shrug*

owi92:
> The playlist titles in the playlists table do not render

Yeah you need to clear the browser cache first. Had the same issue.
Re color: Let's just go with blue for now.

Arnei (approved):

PR #1580, opened 2026-04-16
Fix file upload drag and drop

Arnei:
https://github.com/opencast/admin-interface/pull/1529 reworked the file upload. This lost us the ability to add files to the file upload dialog via drag and drop (from the file explorer).
This patch fixes that by adding appropriate callbacks.

### How to test this

Can be tested as is. When creating an event, try to upload files by dragging the desired file onto the file upload button.

### DIFF
--- a/src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
+++ b/src/components/events/partials/ModalTabsAndPages/NewSourcePage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import Notifications from "../../../shared/Notifications";
 import DatePicker from "react-datepicker";
@@ -241,7 +241,7 @@ const Upload = <T extends RequiredFormPropsUpload>({
 }) => {
 	const { t } = useTranslation();
 
-	const handleChange = (e: React.ChangeEvent<HTMLInputElement>, assetId: string) => {
+	const handleChange = (e: { target: { files: FileList | null } }, assetId: string) => {
 		if (e.target.files) {
 			if (e.target.files.length === 0) {
 				formik.setFieldValue(assetId, null);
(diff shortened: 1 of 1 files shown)


ferishili (changes requested):
Thanks @Arnei,

Drag & drop upload works fine!

However, unlike the regular upload button, there is currently **no file type limitation**.

Because of this, it's possible to upload files like an `.mp4` as a subtitle file (for example, a large video). After completing all steps and clicking the **Create** button, the files are uploaded first (which can take some time). Only afterwards does the error message appear, and the whole process has to be started again.

It would be great if the same file type validation could be applied to drag & drop as well.

The error message is also not very descriptive and disappears after a few seconds.

Arnei:
Hi, thanks for the review :)

I understand why you would like to have file type limitation. It can be all to easy to accidentally drag a file into the wrong field, and you'll only notice after event creation failed.

However, I think it would be better if we do not hard limit what can be added. Because Opencast is so configurable in how its workflows work, users may know better than we do which file types can and cannot go into a certain field. As such, we may suggest file types but should never limit them completely, at least in my opinion.

Getting a more descriptive error message would require changes in the endpoint, I would leave that for a different PR :')

ferishili:
Yes, that's true! However, please note that the regular upload button already has this limitation, so this creates a bit of inconsistency. It would probably be better to either apply the limit to both or remove it from both. What do you think?

Arnei:
The regular upload button that opens the systems file dialog? Yes, that dialog suggests supported file types. However, users can still choose to ignore that and upload whatever they want. So there is no "hard" limit there.

ferishili:
Yes, I mean the regular upload button. But there's still an inconsistency:

* **Regular button:** I can only select video files for the **presenter**, and I cannot select `.vtt` files there and vice versa, meaning I cannot select videos for the **subtitles**  but only `.vtt` files! (which is correct).
* **Drag & drop:** I *can* drag and drop `.vtt` files for the **presenter** and I also can upload video files for **subtitle** (which is not correct).

So the validation behavior is currently different between the two upload methods.

Arnei:
> I can only select video files for the presenter, and I cannot select .vtt files there and vice versa, meaning I cannot select videos for the subtitles but only .vtt files! (which is correct).

And I am telling you that you *can* upload whatever you want there ^^. It's just a bit more effort because you have to change "all supported file types" to "all files".

Maybe we should leave this discussion for a follow-up PR? With the old drag and drop (before it got broken), users could also upload "wrong" file types, so in that sense this PR just makes it work like it did before.

ferishili:
Oh, I see. If it was already like this before, then it should be handled in a separate issue and PR.

So from my side, this is now good to go!

mtneug:
There is [`event.upload.asset.options.properties`](https://github.com/opencast/opencast/blob/develop/etc/listproviders/event.upload.asset.options.properties) allowing to limit files by extension. Not sure if that's relevant here.

PR #1581, opened 2026-04-16
Fixes start date not showing on create event

Arnei:
When creating an event, the start date would not show on the source upload page. This patch fixes that.

### How to test this

Can be tested as is.


ferishili (approved):
Thanks @Arnei
This works fine!

PR #1584, opened 2026-04-23
Fix warning wrongly appearing on template change

Arnei:
When changing the template in the acl tab, if the current acl settings were invalid, a warning would appear *after* the template change about how the acl was invalid, when due to the template change it actually no longer wasn't.
This fixes that.

### How to test this

Get your ACL to be invalid (by removing all roles, or by adding a new role and not selecting a role in the dropdown). Then change template.

### DIFF
--- a/src/utils/aclUtils.ts
+++ b/src/utils/aclUtils.ts
@@ -102,6 +102,5 @@ export const handleTemplateChange = async <T extends { policies: TransformedAcl[
 
 	formik.setFieldValue("policies", template.acl);
 	formik.setFieldValue("aclTemplate", templateId);
-	// Is this necessary?
-	dispatch(checkAcls(formik.values.policies));
+	dispatch(checkAcls(template.acl));
 };


gregorydlogan:
Maybe I'm doing something wrong here, but the default `private` ACL has no roles, so I *immediately* get the modal's error message.  I can't imagine `private` actually works correctly the more I think about it, but I'm not sure if we should remove it, or fix it somehow.

Arnei:
At the very least we should not claim that an ACL needs at least one read/write role, and then violate that claim by having a template that is literally an empty ACL. Maybe add `ROLE_ADMIN` with read/write to `private`? Should effectively be the same, right?

gregorydlogan:
That would be functionally identical, but likely confusing for the user since they aren't listed in their own ACL...

Could we have the UI default to showing the user as an at-least-read role configuration if an empty template is selected?  If the user manually removes all of the roles the behaviour is correct, it's just loading an empty template that causes weirdness - I would expect that both

a) all of the templates work, even if they don't do what I want (ie, I can lock myself out of my own video)
b) I can't deliberately set ACLs such that no one aside from the global admin can help me.

I guess there's a discussion to be had about point a.  Should we even allow that?

Arnei:
The user itself is not listed in any of the templates. But the ACL tab in the event/series details makes sure to add the role of the user on template change, so that they are listed in their own ACL. If that does not happen for `private`, that is a bug actually. Which is not something this PR has to fix actually :P

I am not an admin, but from what I hear, you basically never want to lock yourself out of your own video. It probably should remain technically possible for some edge cases I can't even imagine. But from a usability standpoint we should ensure that a user never accidentally loses rights on their own video, only very deliberately.

PR #1589, opened 2026-04-28
Improve acl tab performance for systems with many roles

Arnei:
Memoization should prevent the dropdown options from getting recreated on every rerender. This should speed up non-dropdown related actions, such as adding new roles or changing access rights, on systems with many roles.

### How to test this

Can be tested against develop, as there are many roles there.

### DIFF
--- a/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
+++ b/src/components/shared/modals/ResourceDetailsAccessPolicyTab.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useMemo } from "react";
 import RenderMultiField from "../wizard/RenderMultiField";
 import {
 	Acl,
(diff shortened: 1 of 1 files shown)


ferishili (changes requested):
Thanks @Arnei,
I can confirm that this PR significantly improves the performance of the dropdown in the ACL tab.

One small observation (see attached video): when hovering over or changing dropdown options, it looks like the entire dropdown is re-rendered. Is this expected behavior, or could it be optimized to render once and then update without re-rendering on every interaction?

https://github.com/user-attachments/assets/42f0b98d-adcb-45f2-b3db-18c6165deea6

Arnei:
At the very least the visually changing elements must be re-rendered I suppose. If you have more ideas on how to optimize our dropdowns, you are welcome to contribut, arguably having dropdowns with this many options is bad design anyhow https://github.com/opencast/admin-interface/issues/1442. But that is not the goal of this PR.

ferishili (approved):

PR #1600, opened 2026-05-04
Show registration modal only when unregistered

### LINKED ISSUE #1516: Adopter Registration Pop-Up Keeps Popping Up
The Adopter Registration pop-up keeps appearing at the start of a session even after having been submitted in the past.
What should happen is that it stops popping up after a successful submission.

It is assumed that [the current code doesn't check wether the form has been submitted](https://github.com/opencast/opencast-admin-interface/blob/develop/src/components/Header.tsx#L137-L148).

See [initial discussion](https://matrix.to/#/!HBsLKoOIEEOkxAeQvC:matrix.org/$6er0UWw0o1ujjQoDS34eb85jNorESohaLJM6rPuzzAQ?via=matrix.org&via=elan-ev.de&via=uni-osnabrueck.de).

gregorydlogan:
This PR only shows the registraiton modal when the adopter has not already registered.

~This PR contains a few extra bits that haven't been turned on yet - notably the warning bell stuff for when the core can't reach the registration server.~

This PR now uses the bell if the user is not registered.  Notably: 
If you haven't registered at all:
Pasted image
If you have registered, but haven't agreed to the latest ToU:
Pasted image (2)

Clicking on either of these notices opens the registration modal.

Requires https://github.com/opencast/opencast/pull/7786 on the backend.

TODOs:
- ~Make the new strings into translatable strings rather than hardcoding them.~
- ~In the case of an out of date ToU, *uncheck* the agreed-to-ToU checkbox when presenting the registration modal, and highlight what's wrong.~
- In the case of an out of date ToU, launch the registration modal to the input pane, not the general info pane.

Fixed #1516


src/components/Header.tsx
```diff
@@ -51,13 +57,29 @@ const Header = () => {
...
+	}
+	// dispatch(fetchLatestToU());
+	// dispatch(fetchIsUpToDate());
+
+	const _getLatestToU = async () => {
+		await dispatch(fetchLatestToU());
+	};
+	const _getIsUpToDate = async () => {
+		await dispatch(fetchIsUpToDate());
+	};
```
Arnei: This is going unused and should be removed.

src/components/Header.tsx
```diff
@@ -51,13 +57,29 @@ const Header = () => {
...
 	const orgProperties = useAppSelector(state => getOrgProperties(state));
 	const displayTerms = (orgProperties["org.opencastproject.admin.display_terms"] || "false").toLowerCase() === "true";
 
 	const loadHealthStatus = async () => {
 		await dispatch(fetchHealthStatus());
 	};
 
+	if (registration == null) {
+		dispatch(fetchRegistration());
+	}
```
Arnei: This will dispatch a redux action (and thus a fetch request) every time the component rerenders and registration is still null. This could potentially cause multiple fetch requests. To avoid this, we usually use a useEffect that only runs once on component mount for fetch requests. (see the useEffect in e.g. SeriesDetails.tsx).

If you need to monitor the state of the action, you would usually add a variable to the redux state that tracks the action state (LOADING, SUCCEEDED, FAILED what have you). Then track the action state in your component via a selector and use a useEffect to respond properly.

And if you don't actually need all that redux boilerplate, you could also do the fetch requesting and state tracking here with useEffect and useState.

src/slices/registrationSlice.ts
```diff
@@ -0,0 +1,165 @@
+import { PayloadAction, createSlice } from "@reduxjs/toolkit";
+import axios from "axios";
+import { WritableDraft } from "immer";
+import { createAppAsyncThunk } from "../createAsyncThunkWithTypes";
+
+export type Registration = {
```
Arnei, line 5: Since all this PR really does is checking if ´/admin-ng/adopter/registration` returns null or not, creating a whole redux state is complete overkill. If you'll need this for some other PR, would it not be better to add it then?

Arnei (changes requested):

src/slices/registrationSlice.ts
gregorydlogan: Fair.  The goal here is to be able to set warnings in the notification bell in a few cases.  I'll rework this and come back when it's ready.

gregorydlogan: Is this more what you're looking for @Arnei?  Or are you wanting me to completely remove the slice and selectors?  There's an `adopterRegistrationUtils` file which powers the modal and works, but I just can't seem to get the relevant function(s) to *work* from the header...

Arnei: I was thinking of completely removing the slice and selectors and stuff, and do something like in About.tsx, aka just do the state thing locally in the component.

If you plan to make use of redux state in a later PR, it might make sense to use redux right now already, so you don't have to do work twice. But as a reviewer, I can't know that :D

gregorydlogan: You have inspired me to actually finish the work rather than continue stripping stuff out .  State and stuff is now *used*, yay.

gregorydlogan:
Now that the GHA issues are done, I made a final change and I think this is ready for re-review @Arnei

src/styles/components/_header.scss
```diff
@@ -161,6 +161,9 @@
         .dropdown-ul{
             width: auto;
             right: 8px;
+          .wide-text{
+            padding-right: 5px;
+          }
```
Arnei:
```suggestion
            .wide-text{
                padding-right: 5px;
            }
```

Arnei: You're using 2 spaces, but the file uses 4.

src/components/Header.tsx
```diff
@@ -138,18 +155,22 @@ const Header = () => {
...
+		const isAdmin = user.isAdmin || user.isOrgAdmin;
+		const isLocalhost = window.location.hostname === "localhost";
+		const lastDismissed = localStorage.getItem("adopterModalDismissed");
+		const THIRTY_DAYS = 30 * 24 * 60 * 60 * 1000;
+		const dismissedLongEnough = !lastDismissed || Date.now() - parseInt(lastDismissed) > THIRTY_DAYS;
+
+		if (isAdmin && !isLocalhost && dismissedLongEnough && registration == null) {
+		  showRegistrationModal();
+		}
+	}, [user, registration, dispatch]);
```
Arnei: This can cause an infinite loop. `dispatch(fetchRegistration());` will eventually update `registration`, which triggers a rerun of the `useEffect` hook, which causes `dispatch(fetchRegistration());` to be run again...

I recommend putting the `dispatch(fetch....`'s into their own useEffect hook that runs only once on mount. If you need to run them again, you should probably do so conditionally?

src/components/Header.tsx
```diff
@@ -361,6 +401,26 @@ const MenuNotify = ({
 					)}
 				</li>
 			))}
+      {!registering &&
```
Arnei: Not using tab indentation here

src/components/Header.tsx
```diff
@@ -361,6 +401,26 @@ const MenuNotify = ({
...
+        <li>
+          <ButtonLikeAnchor
+            onClick={() => showAdoptersRegistrationModal()}
+          >
+            <span className="wide-text">Registration</span>
+            <span className="multi-value multi-value-yellow">Unregistered</span>
```
Arnei: multi-value-yellow does not seem to exist (maybe I removed it in my cleaning frenzy?).

src/components/Header.tsx
```diff
@@ -361,6 +401,26 @@ const MenuNotify = ({
 					)}
 				</li>
 			))}
+      {!registering &&
+        <li>
+          <ButtonLikeAnchor
+            onClick={() => showAdoptersRegistrationModal()}
+          >
+            <span className="wide-text">Registration</span>
```
Arnei: Untranslated strings.

src/components/Header.tsx
```diff
@@ -361,6 +401,26 @@ const MenuNotify = ({
...
+            <span className="multi-value multi-value-yellow">Unregistered</span>
+				  </ButtonLikeAnchor>
+        </li>
+      }
+      {registering && !updatedToU &&
```
Arnei: So if I am unregistered and the ToU have been updated, there would be two notifications. Is that intended?

src/slices/registrationSlice.ts
```diff
@@ -0,0 +1,110 @@
+import { PayloadAction, createSlice } from "@reduxjs/toolkit";
```
Arnei, line 1: Inconsistent use of spaces and tabs in this file.

src/slices/registrationSlice.ts
```diff
@@ -0,0 +1,110 @@
...
+		setError(state, action: PayloadAction<{
+			error: RegistrationState["error"],
+		}>) {
+			state.error = action.payload.error;
+		},
+	},
+	// These are used for thunks
+	extraReducers: builder => {
+		builder
+			/* .addCase(fetchRegistration.pending, state => {
```
Arnei: Leftover code?

src/slices/registrationSlice.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export type RegistrationState = {
+  registration: Registration | null,
+  latestToU: string,
+  isRegistering: boolean,
+  agreedToToU: boolean,
+  error: boolean
+};
+
+type Temp = {
```
Arnei: `Temp` is not a great type name ^^

src/slices/registrationSlice.ts
```diff
@@ -0,0 +1,110 @@
...
+export type Registration = {
+  agreedToPolicy: boolean,
+  registered: boolean,
+  termsVersionAgreed: string,
+}
+
+export type RegistrationState = {
+  registration: Registration | null,
+  latestToU: string,
+  isRegistering: boolean,
```
Arnei: I feel like `isRegistering` is a bit confusingly named? This is not about being in the process of registering, is it? Maybe `canRegister` would be clearer?

src/slices/registrationSlice.ts
```diff
@@ -0,0 +1,110 @@
...
+export const fetchIsUpToDate = createAppAsyncThunk("registration/isUpToDate", async () => {
+  const res = await axios.get<boolean>("/admin-ng/adopter/isUpToDate");
+  return res.data;
+});
+
+const registrationSlice = createSlice({
+	name: "registration",
+	initialState,
+	reducers: {
+		setError(state, action: PayloadAction<{
```
Arnei: I don't see `setError` nor `error` being used anywhere. Leftover code?

Arnei (changes requested):

src/components/Header.tsx
gregorydlogan: `registering` is whether the backend is talking with the registration server.  `updatedToU` is a boolean of whether the ToU accepted by the user matches the latest provided by the backend.

You can firewall off your backend so that it can't reach the internet *and* also not have agreed to the latest ToU, so yes two separate notifications is expected in that case.

src/slices/registrationSlice.ts
gregorydlogan: I fought with naming here a lot, and I agree that it sucks.  This boolean is true if the backend can talk to the main registration server, false otherwise.

What about something like `ableToRegister`?

src/components/Header.tsx
gregorydlogan: Ok, did not know that.  I had that previously and moved them in there since it seemed simpler.  Should not need to re-run these, at least not until next time we boot.

src/components/Header.tsx
Arnei: The text in the little colored bubbles is now translated. Shouldn't "Registration" be translated too?

src/slices/registrationSlice.ts
Arnei: `ableToRegister` sounds good too!

gregorydlogan:
Ok, so I think this is more or less done...

I noticed that the registration modal itself was using a script in utils, so I removed it and moved its functionality into the selector.  Hopefully I didn't miss anything obvious.  This now requires https://github.com/opencast/opencast/pull/7786 as well.

src/components/shared/RegistrationModal.tsx
```diff
@@ -90,23 +111,13 @@ const RegistrationModalContent = () => {
...
-		setStatisticsSummary(info);
-	};
 
 	const handleSubmit = (values: Registration) => {
 		// post request for adopter information
-		postRegistration(values)
+		postAdopterRegistration(values)
 			.then(() => {
+				// Refetch the registration data since otherwise what we just submitted does not show up if the user immediately returns to the modal
+				dispatch(fetchRegistration());
```
Arnei: This seems like bad practice to me. We need to fetch registration data on modal open anyway, right? So if submitting closes the modal, and the user then reopens the modal, the data should be fetched anyway?

Unless we allow the user to submit without closing the modal, and the backend modifies the data we want to display?

src/components/shared/RegistrationModal.tsx
```diff
@@ -66,20 +80,27 @@ const RegistrationModalContent = () => {
 		street: "",
 		streetNo: "",
 		email: "",
+		termsVersionAgreed: "",
 		agreedToPolicy: false,
+		dateModified: "",
+		dateCreated: "",
 		registered: false,
+		statistics: null,
```
Arnei: Why are we adding statistics to formik? I don't think we want to let the user change them? And it makes handling formik values more annoying, because now we have to do stuff like filter the statistics variable away (e.g. below in the summary)

src/components/shared/RegistrationModal.tsx
```diff
@@ -66,20 +80,27 @@ const RegistrationModalContent = () => {
...
-		statistics: { [key: string]: unknown },
-	}>();
+	useEffect(() => {
+		if (!registrationLoaded) {
+			dispatch(fetchRegistration());
+		}
+		if (!statisticsLoaded) {
+			dispatch(fetchStatistics());
+		}
+	}, [registrationLoaded, statisticsLoaded, dispatch]);
```
Arnei: Not quite sure on the logic here. When registration has loaded but statistics has not yet loaded, we want to fetch statistics again?

If you just want to fetch things once on component mount, you'd do

```
useEffect(() => {
	dispatch(fetchRegistration());
	dispatch(fetchStatistics());
}, [dispatch]);
```

src/slices/registrationSlice.ts
```diff
@@ -51,6 +121,54 @@ export const fetchIsUpToDate = createAppAsyncThunk("registration/isUpToDate", as
...
 
+// This is whether the statistics used in the registration modal.
+export const fetchStatistics = createAppAsyncThunk("registration/statistics", async () => {
+  const res = await axios.get<Statistics>("/admin-ng/adopter/statistics");
+  return res.data;
+});
+
+
+  // post request for adopter information
+  export const postAdopterRegistration = async (
```
Arnei: This block is one indentation level off

Arnei (changes requested):

src/components/shared/RegistrationModal.tsx
gregorydlogan: I *think* this is an artifact of the `registrationLoaded` boolean, since it does not refetch without clearing that.

That fflag is there because the endpoint returns a bare `null` if there is no registration data, which makes differentiating between not-loaded and loaded-but-blank states hard. I hadn't started the backend pr changed when I added the loaded flags, so maybe I should change that in the backend pr...

src/components/shared/RegistrationModal.tsx
gregorydlogan: Will swap this back!

src/components/shared/RegistrationModal.tsx
gregorydlogan: This is a bad rebase, I'll fix it. I was having issues with Formik and thought this was related.

src/components/Header.tsx
Arnei: How has this been resolved?

gregorydlogan: Via https://github.com/opencast/admin-interface/pull/1600/changes/bf6b3afb66aa4493892c2356df4e675363a1fb91 :)

Arnei: Ah gotcha, did not realize. Please remove the other line this shows up then as well.

gregorydlogan: Done!

Arnei (approved):

PR #1604, opened 2026-05-11
Add change password option

Arnei:
Fixes https://github.com/opencast/admin-interface/issues/356.

Adds a new button in the user menu (the dropdown in the top right corner) that allows users to change their own password. The idea is to allow users to change their password without having to give them full access to the users table.

Requires a new endpoint: https://github.com/opencast/opencast/pull/7646

Bildschirmfoto vom 2026-05-11 13-48-27
Bildschirmfoto vom 2026-05-11 13-48-36

### How to test this

Requires an Opencast with the backend PR linked above.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -8,6 +8,7 @@ import { setSpecificServiceFilter } from "../slices/tableFilterSlice";
 import { getErrorCount, getHealthStatus } from "../selectors/healthSelectors";
 import {
 	getOrgProperties,
+	getUserBasicInfo,
 	getUserInformation,
 } from "../selectors/userInfoSelectors";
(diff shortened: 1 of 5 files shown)


mtneug:
For some users, changing the password is not possible. `admin` usually is an in-memory user with a password defined in a configuration file. Users logging in through an SSO system also shouldn't be able to change their password. Really, only JPA users should have the option. Is this already considered by the frontend?

Arnei:
This is considered by the frontend insofar as that a non-"manageable" user cannot change their password. Same as for the users table, where you cannot change a non-manageable user. I am not certain that "non-manageable" is the same as "only JPA", but I would expect it to be.

mtneug:
Nice. I think this is the correct check. At least `admin` and user references are non-"manageable".

PR #1605, opened 2026-05-11
Enable eslint no-misused-promises

Arnei:
Enables the rule no-misused-promises and makes the necessary changes.


gregorydlogan (changes requested):
This no longer passes checks when merged to `r/19.x` since #1604 was merged:

```
/home/greg/opencast/admin-interface/src/components/Header.tsx                                                                                 
  471:31  error  Promise-returning function provided to attribute where a void return was expected  @typescript-eslint/no-misused-promises    
                                                                                                                                              
✖ 1 problem (1 error, 0 warnings)
```

PR #1612, opened 2026-05-18
Fix workflow displayOrder not being respected

### LINKED ISSUE #1508: displayOrder in WF def ignored
Observed behaviour (OC v19.0):
The displayOrder property in workflow definitions (XML/YAML) has no effect.
Workflows in dropdowns (e.g. Start Task in Archive) are sorted alphabetically (by title).

Expected behaviour:
Workflows should sort by displayOrder value, as described in the [docs](https://docs.opencast.org/r/19.x/admin/#configuration/workflow/#step-2-describe-the-workflow).

Possible cause:
Feature may not be implemented in current code base.

Suggested action:
Either Add/Fix or Remove feature entirely.

Arnei:
Fixes #1508.

Workflows can have a "display order" which defines in which order they should be displayed in a dropdown.
Our workflow select dropdowns were not respecting this order, and this patch fixes that.

Also includes some code duplication reduction because why not and fixes an issue for scheduled events, where the workflow config was not properly changing when changing the workflow.

### How to test this

Can be tested as is, but for checking if the ordering is really respected you should change the "displayOrder" values in Opencast (in `etc/workflows`).


gregorydlogan:
Not that this is an issue with this PR, but I found it somewhat surprising that we've already handled cases where someone does something silly with the display order value:
```
2026-07-08 16:32:21,251 | WARN  | (WorkflowDefinitionScanner:251) - Unable to parse workflow from file 'schedule-and-upload.yaml', com.fasterxml.jackson.databind.JsonMappingException: Numeric value (1000000000000000000) out of range of int (-2147483648 - 2147483647)
 at [Source: (FileInputStream); line: 6, column: 34] (through reference chain: org.opencastproject.workflow.api.WorkflowDefinitionImpl["displayOrder"])
```

PR #1619, opened 2026-06-11
Enable writing to ghcr from pull requests

gregorydlogan:
This PR hopefully fixes test container image workflow's current failures.  Previously, this workflow used `pull_request_target`, which has additional, but unsafe, permissions in GHA.  We recently swapped to `pull_request`, which does *not* have permission to write to ghcr, despite adding `packages: write` to the workflow.  Those permissions would work correctly, but *only for PRs originating from the same repository*.

Instead we have defined a custom secret which does have said permissions, and this PR now makes use of that secret.

### DIFF
--- a/.github/workflows/pr-deploy-container-image.yml
+++ b/.github/workflows/pr-deploy-container-image.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          password: ${{ secrets.SUBMODULE_PR_DOCKER_WRITE }}
 
       - name: Extract metadata (tags, labels)
         id: meta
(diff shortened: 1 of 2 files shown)


mtneug:
I assume this is also about the question you ask in Matrix.

I don't know if I like this change. Does that mean any PR can upload images to ghcr? If we only use ghcr for test images I guess this is fine, but I can also see us use if for prod images. In fact, we already do that for octoka. Also, I don't think images are necessarily tied to repositories, i.e. PRs might be able to upload and overwrite images from other repos in the org. I would rather not have test images than that having this risk.

gregorydlogan:
The images have restrictions in terms of which repositories can write to them at a per-image level.  Even if someone were to jailbreak from here, they wouldn't be able to write to anything other than the admin interface repo.

Also, this approach doesn't work at all as is.  I've pushed a refactoring which *does* seem to work **and** doesn't need extra secrets.

.github/workflows/pr-deploy-container-image.yml
```diff
@@ -1,27 +1,19 @@
-name: PRs » Publish container image
+name: PRs » Push container image
 
 on:
-  push:
-    branches-ignore:
-      - 'dependabot/**'
-  pull_request:
+  workflow_run:
```
gregorydlogan, line 4: Workflows started with `workflow_run` have the requisite permissions (or at least can have, with the permissions block below) to write to the package repo.

PR #1620, opened 2026-06-12
Fix use of NODE_VERSION variable

gregorydlogan:
This PR resolves an issue where the GHA scripts were unable to find a value for `$NODE_VERSION` when installing Node because I was using the variable as a bash shell variable, rather than a GHA templating subtitution.  This PR swaps that, so now the builds run.

We haven't been noticing this because this only affects PRS based off the current tip of `develop`.


gregorydlogan:
This is *going* to conflict with #1619 and #1621, so marking this as draft.

PR #1653, opened 2026-08-10
Type acl default config keys

Arnei:
We know which keys we can receive from thebackend, so we might as well properly type them. Should make our code simpler to read.

### How to test this

This affects the ACL tab, so ensure that that tab still works as expected.


src/slices/aclSlice.ts
```diff
@@ -60,6 +60,17 @@ export type AclTemplate = {
...
 
+export type AclDefaults = {
+	"read_enabled": boolean,
+	"write_enabled": boolean,
+	"read_readonly": boolean,
+	"write_readonly": boolean,
+	"default_actions": string[],
+	"keep_on_template_switch_role_prefixes": string[],
+	"display_role_filter_blacklist_prefixes": string[],
+	// [key: string]: string
```
gregorydlogan: Leftovers?

gregorydlogan (changes requested):

